PR #56, opened 2013-02-19
Update download version to 0.8.0

gavinandresen:
Minor change for version 0.8.0 Bitcoin-Qt release


holaymolay:
nice work! I had noticed that as a suggestion, but I thought I should talk to you before installing stuff willy nilly

PR #550, opened 2014-09-12
add Quickwallet to bitcoin.org

Quickwallet:
This pull request adds Quickwallet to the list of wallets


crwatkins:
@Quickwallet thanks for your submission.

How does the user regain access to their funds if the Quickwallet server is down?

PR #556, opened 2014-09-16
Russian translation of the original PDF paper added

arvicco:
This is the correct pull request to master that only adds Russian PDF and 2 links to it. Please ignore the other ones, sorry for the confusion.


arvicco:
@saivann Not sure what happened, but it seems that Russian translation of original Satoshi paper was removed and replaced by English one. And the links from Russian localization pointing to an English version. I want to understand the reasons. Was there some discussion about this? I can't find anything in Issues, so I'd appreciate pointing to a relevant discussion.

harding:
@arvicco links should point to this page, which lists the Russian translation (and all the other accepted translations, as well as the original):

https://bitcoin.org/en/bitcoin-paper

This was discussed here:

https://github.com/bitcoin-dot-org/bitcoin.org/pull/839

If there's something broken, please let me know.

arvicco:
@harding Thanks! I was just wondering why Russian localization points to English version of the paper everywhere, rather than Russian one... In other words, may I change these references to the accepted Russian translation of the pdf, or it will contradict some policy?

harding:
The way I'd like it to work (which we can discuss if you think it's sub-optimal) is for all references to the whitepaper in bitcoin.org/ru/ to point to the Russian localized version of bitcoin.org/en/bitcoin-paper (which I don't think exists yet).  That way readers can select to read either your Russian version or (if their English is good enough) Nakamoto's original version.

Note that we did the same thing on the English version of the site so that anyone clicking a regular link to the paper can choose to read either the original or one of the translated versions.  (More importantly, I think, it makes everyone aware that there are translated versions; previously I think people on the English site assumed we only had the English original, and I was concerned that there was duplication of effort as multiple people kept translating it into the some other languages.)

arvicco:
@harding This actually makes sense. I'll change the links in Ru localization to point to https://bitcoin.org/en/bitcoin-paper then.

harding:
@arvicco If the links do something like this,

    <a href=\"#bitcoin-paper#\">original paper</a>

The link will automatically point to `/ru/<bitcoin paper translated into Russian>` when it becomes available or `/en/bitcoin-paper` if the Russian localized version is not available.

PR #577, opened 2014-09-25
list the Coinapult wallet

g-p-g:
Happy to answer any questions about the wallet and service.


saivann:
A quick update on the last point (account recovery by email); Circle asks users to answer security questions when recovering their account and Coinkite only sends a password hint by email. Coinbase seems to be the only service to allow resetting the password by email with no additional authentication. I am currently asking them if 2FA becomes mandatory as soon as the account reaches a certain threshold.

luke-jr:
@saivann I would find it annoying as a user if Coinbase made 2FA mandatory for mere logins.

saivann:
@luke-jr In case my last comment could be misleading, I am not saying they should use it in a certain way (e.g. I think some services ask 2FA mostly only when spending). I am only asking details on how they actually work, so I can know if compromising a high-value coinbase account remains possible by only having access to the user email account.

g-p-g:
The PR no longer has conflicts, again.

saivann:
@g-p-g It seems most other web wallets have measures in place to prevent users' to lose funds if their email account is compromised. Coinbase confirmed to me that account recovery by email will only work when the request comes from a recognized device for accounts with a balance over 1000 USD.

This being said, Xapo apparently is the only exception and since this is not a current requirement, I am commenting in favor of including Coinapult. However, I'd like to encourage you to put in place more demanding authentication for account recovery than only email verification and set spending limits on "bitcoins by email/SMS" between unregistered users.

I plan to submit something similar to "Spending bitcoins only by having access to the user email account is not possible" as a requirement soon. Unless there is opposition to that new requirement, Coinapult's listing would be short lived, although it could be resumed as soon as the wallet meets the requirement.

saivann:
In the absence of critical feedback, this pull request will be merged on January 22th.

g-p-g:
@saivann thanks for the update.

I'm guessing you meant the opposite so only accounts with less than 1k USD can get a password reset through email from this other service you're mentioning, right?

We're always improving the various different services we offer, and several others that are not directly visible to end users. So it's quite likely that we'll soon deploy improvements in this specific area you're commenting on; you can also send questions to support@coinapult.com regarding the status on specific features.

saivann:
@g-p-g Thanks!

>I'm guessing you meant the opposite so only accounts with less than 1k USD can get a password reset through email from this other service you're mentioning, right?

No actually regarding the approach used by Coinbase, they're allowing account recovery by email for all accounts, but as soon as the account reaches 1k USD, account recovery will only work when done from a device that was previously verified (successful login and email verification), I guess they're either keeping a whitelist of IP addresses or some cookie in the browser cache.

g-p-g:
I can't tell if there's any actual security on that approach or if it's trying to be secure by being obscure, I certainly don't know what's done there based on that description. In any case that's not entirely relevant, we're not trying to mimic it and instead we'll actually increase the security for password resets. Thanks again for going through all this and for providing a couple of useful suggestions.

PR #601, opened 2014-10-06
The GreenAddress has two-factor authentification

Perlover:
How with two-factor of third party it can be decentralized?
It's centralized wallet for Android!

### DIFF
--- a/_templates/choose-your-wallet.html
+++ b/_templates/choose-your-wallet.html
@@ -576,7 +576,7 @@ wallets:
         - ios
         check:
           control: "checkpasscontrolmulti"
-          decentralization: "checkpassdecentralizeservers"
+          decentralization: "checkfaildecentralizecentralized"
           transparency: "checkpasstransparencyopensource"
           environment: "checkpassenvironmenttwofactor"
           privacy: "checkpassprivacybasic"


harding:
Closing for multiple reasons:

1. The original intent of this PR was to remove the decentralized score from GreenAddress because the PR author believed they required two-factor authentication for spending.  But 2FA isn't required (you can use your backup instead, once the time lock expires).

2. The discussion expanded to other aspects of decentralization.  Some of those aspects have been addressed (e.g. in PR #608).

3. There has been no discussion on this issue for more than six months.

If, after all this time, anyone still has a policy recommendation based on this discussion, please open a new issue or pull request.

gurnec:
I think there's still some question over whether the GreenAddress mobile app should keep its score of
checkPassDecentralizeServers or should be changed to checkFailDecentralizeCentralized. I'm unclear if anything has changed on the @greenaddress side since [Saivann's comment above](https://github.com/bitcoin/bitcoin.org/pull/601#issuecomment-58077765).

Should a separate issue be opened for this?

saivann:
@gurnec Right. I think that @greenaddress response has been the longer but better option of going fully SPV with their latest app GreenBits.

PR #630, opened 2014-11-01
Drop good tor privacy score for mycelium

ghost1542:
Seems like Mycelium has a bug that makes it ignore socks settings (and tor).
https://github.com/mycelium-com/wallet/issues/109

@apetersson Feel free to comment here if you have relevant information, or to report when the app is fixed.

In the absence of critical feedback, this pull request will be merged on November 2th.


apetersson:
update: finally, in version 2.2. we have explicit hidden service support.

PR #663, opened 2014-11-30
Revisit blockchain.info listing

ghost1542:
Blockchain.info received a relaxed review by being in the first wallets added on bitcoin.org. For basic fairness with other wallets, I think it should be revisited with reasonable criterias at least as demanding as other wallets.

After some review and research, I have found several important issues summarized below.

**Bugs and losses**: BC.i has suffered from concerning issues during the past year (e.g. iOS denomination bug, weak key stretching, sizeable number of users losing funds, and more issues I haven't verified yet)

**Backup/Password Security**: BC.i hasn't adopted security features which are slowly becoming standard in other wallets (e.g. BIP32, random passphrases, backup on setup, rotating addresses, 2FA by default).

**Transparency**: Source code of the app has been reset or not updated repeatedly, making bitcoin.org often relay the false claim that the app is open-source.

To be fair, each of these issues would have blocked or delayed listing BC.i if the wallet was submitted today. Accordingly, I think the logical next step to incentivize security and reduce risks for the user is to raise the bar for BC.i like other wallets. I suggest 7 days for BC.i to review/comment the list of issues below (@zootreeves), then suspend BC.i listing for at least 60 days. After this period, the wallet could be re-submitted for review.

Note: This pull req is **not** for bashing BC.i, I know they are having a bad image lately and not interested to feed that. If anything, I hope this can give them an opportunity to show they can address these issues.

In the absence of critical feedback, this pull request will be merged on December 6th.
## 

Some issues for which I think fixes would be reasonable in the future:

BC.i's security relies on strong passwords (way more than a custodial wallet due to bruteforce attacks), yet BC.i's wallets are stored online (BC.i, email, Google drive, dropbox) and:
- Don't use truly random mnemonic passphrase, instead rely on users' passwords known to be weak. Edit: (Although this suggestion might not be _the_ good one, see @vbuterin comments below)
- Allow weak passwords such as 0000000000 - Edit: fixed
- Password strength warnings are optional (and blue) - Edit: fixed
- Password strength information makes false claims (126 years to crack 123456789123456789) - Edit: fixed
- It is unclear to me yet if the new key stretching is secure enough.

BC.i does not remind or require the user to enable 2FA.
Edit: (Although it now appears that 2FA was almost no effect with the current security model - similarly to wallets on bitcoin.org using no third parties)

BC.i still has unpublished source code for their Android app (4.0.17 on Google play, 4.0.12 on GitHub).
Edit: (Same issue with the new repository, 4.0.17 on Google play, 4.0.15 on GitHub)

BC.i does not enable HSTS (seems to me this could be a simple change to reduce MITM attacks).
Edit: Blockchain.info implemented HSTS.
## 

Some issues that may not meet future requirements:

The wallet does not rotate addresses.

The wallet does not uses BIP32 wallets.

Edit: The wallet does not uses deterministic ECDSA nonce (RFC6979).
## 

Some issues that may require to be fixed to avoid downgrading the control score:

BC.i does not let the user backup and access their wallet offline.

BC.i does not encourage or tell the user to download or print a copy of their wallet during setup.

BC.i requires recurrent manual backup, and automated backups is not set by default.

(BIP32 again could fix those issues altogether)


Dorson:
Disagree on the social level.

They give users the private keys. They also communicate well on Github.

It's not so bleak as you paint it from the social perspective.

The only real issue is that they have an HD wallet, and the keys are only given to the main address, you have to generate the keys on an external tool, so that you know the keys to the random addresses that the wallet has created.

But then again, name me another web wallet that gives you all the keys to all the HD addresses, in case you want to export to another wallet.


Proposal:

Talk to the devs on Github. Ask them to implement some features until some given date in 6 month or so. If not, then drop the wallet from the recommended list around that date ! Or put a clear warning in red around the wallet name.

Bashing this major wallet is not good, if the people behind it are well intended. At least some warning or grace period is needed, if we want to eject the good people left and right.

SomeoneWeird:
Way to revive a 3 year old thread :)

Dorson:
Maybe yes to a revival, seen that thread on twitter recently. Did not see the dates, before reading. It's a reoccurring topic, then the situation changes.

I think it's time to talk to the blockchain.info devs on Github maybe.

The other recommended wallets are all aging. The Blockchain.info  did upgrade to dome degree, so it might be a good time to revise the recommendation as: "use on your own risk or so", but at leas they dive you the keys for export/back-up, unlike other wallets.

The multisig wallets controlled are recommended on bitcoin.org. What will happen to the multisig, when the company goes out of business in a bust ?

Blockchain.info now has the 2F logins. The best ones in the space, where the one tab is doing the auth all automatically.

The only issue is the HD keys to export. 

Example: In case you have the 1000 addresses in there, with 1 USD for each, then it's a problem, because they only provide the export key to the main address, when you export into the alternative or new wallet.

Then again it might be an issue with all HD wallets.

The first and main factor for a wallet recommendation should be the keys. If you can export / back-up the keys, then the web wallet is better than the one that does not allow such thing.

Security is the next step to consider, or warn about.

Maybe there is a need for the list of the wallets to warn about. Like a list of the major wallet provides and what kind of down-sides they have, at the current date of the recommendation.

Dorson:
Furthermore just looked at the Github of https://github.com/blockchain

And the https://github.com/greenaddress

Greenaddress.it is recommended on bitcoin.org, while the blockchain.com / or .info is not, while it has more active development.

They seem to be polishing their shapeshift integration and the money making part of the wallet at the moment, which is OK and sustainable for a company that wants to make a profit.

I would not recommend the blockchain.info V3 wallet blindly, but at least they offer the access to the keys and the 2F auth + blocking of people who want to brute-force the login.

dabura667:
> What will happen to the multisig, when the company goes out of business in a bust?

You recover using your backup keys... Every multisig service and wallet offers backup keys (2 of 3, user holds 2 so can always recover, or 2 of 2 with timelocked recovery transactions sent periodically via mail)

kristovatlas:
> Example: In case you have the 1000 addresses in there, with 1 USD for each, then it's a problem, because they only provide the export key to the main address, when you export into the alternative or new wallet.

Blockchain.info uses BIP-44 for key derivation, which is compatible with other wallets including Trezor. There are, of course, many tools online that can derive individual addresses from the master xpub.

crwatkins:
@Dorson  You may or may not have noticed that this issue was closed over two years ago.  
The best method to add a wallet to the listings is to submit a PR with screenshots and proposed scoring.  The developer (or any interested third party) that is willing to support the listing and believes that the wallet meets our listing criteria is welcome to submit the PR.

PR #673, opened 2014-12-08
remove references to knc wallet

pstratem:
knc wallet appears to be abandonware

there hasn't been a commit since august

https://github.com/kncgroup/bitcoin-wallet/commits/master


saivann:
FWIW, I plan to submit "Avoid address reuse by using a new change address for each transaction" as a requirement when MultiBit is released. KnC and Hive for Mac would be delisted. The Hive team are interested to catch up eventually on this feature but their web wallet already meets the requirement.

harding:
I've received an additional complaint by email from that kncwallet.com no longer responds (verified) and that KnC wallet for iOS (which we don't list) does not work without a response from the website server.  (Although the email I received did imply that it's possible to recover funds using the seed, so thank goodness for that.)

Based on this information, later today I plan to rebase this commit removing KnC and merge it.  If the KnC team reappears, we can revert the change once they address our abandonware concerns.

saivann:
Sounds good to me

LinusU:
Just wanted to chime in and say that the outage of kncwallet.com is not the cause of the iOS problems. **Correlation does not imply causation.**

Regarding the issue at hand, I no longer work on the wallet and it's not my place to comment on it.

PR #685, opened 2014-12-18
Dev Docs: Document Blocks-First IBD & Orphan Blocks

harding:
Dev Docs: Document Blocks-First IBD & Orphan Blocks

Preview IBD section: http://dg0.dtrt.org/en/developer-guide#initial-block-download
Preview Orphan Blocks section: http://dg0.dtrt.org/en/developer-guide#orphan-blocks

This pull adds a new section about the Initial Block Download (IBD) method used in Bitcoin Core 0.9, which I've retroactively named blocks-first. It also adds a short section explaining the difference between orphan blocks and stale blocks.  See the commit message for minor details.

This is a partial resubmission of closed pull #660 focusing only on the current IBD method. This keeps the text shorter and, hopefully, easier to review. I also think this version is much, much better---it includes nearly 100% rewritten text plus five new illustrations that didn't appear in the original pull. After we get this merged, I'll rewrite and resubmit the headers-first part of pull #660.


harding:
Rebased to fix merge conflicts.

_includes/guide_p2p_network.md
```diff
@@ -140,6 +140,180 @@ In order to maintain a connection with a peer, nodes by default will send a mess
...
+is 4860...0000 as seen in the illustration above.)
+
+The IBD node uses the received inventories to request 128 blocks from
+the sync node in the `getdata` message illustrated below.
+
+![First GetData Message Sent During IBD](/img/dev/en-ibd-getdata.svg)
+
+It's important to headers-first nodes that the blocks be requested and
+sent in order because each block header references the header hash of
+the preceeding block. That means the IBD node can't fully validate a
```
saivann: @harding s/preceeding/preceding/

_includes/guide_p2p_network.md
```diff
@@ -140,6 +140,180 @@ In order to maintain a connection with a peer, nodes by default will send a mess
...
+But if it doesn't find a block with that hash, it takes the next header
+hash from the `getblocks` message and searches its block chain for that
+hash. If that hash matches, it will reply with 500 block inventories
+starting with the following hash from that point. But, again, if it
+doesn't find that hash, it will proceed to check the next hash in the
+message (and so on until it runs out of hashes in the message). If the
+last hash in the message (besides the stopping hash) doesn't match, it
+assumes the only block the two nodes have in common is block 0 and so it
+sends an `inv` starting with block 1 (the same `inv` message seen
+several illustrations above).
```
saivann: @harding FWIW, I was temporarily confused here, probably due to the repeating and conditional sentences, and felt this could perhaps be explained in a more concise way. What would you think of something similar to the following?

----

Upon receipt of the second getblocks message, the sync node searches its local best block chain for a block that matches one of the header hashes in the message, trying each hash in the order they were received. As soon as it finds a matching hash, it will reply with 500 block inventories starting with the following hash from that point. However if there is no matching hash (besides the stopping hash), it assumes the only block the two nodes have in common is block 0 and so it sends an inv starting with block 1 (the same inv message seen several illustrations above).

saivann:
@harding Thanks!! LGTM

harding:
Rebased to fix merge conflicts and added commit 3109903 making the changes @saivann suggested.  (Thanks!)  Preview updated.

In the absence of critical feedback, this will be merged around 22:00 UTC Sunday.

PR #692, opened 2014-12-24
Bitcoin paper - First version of brazilian portuguese translation

rodrigosilvap:


rodrigosilvap:
@saivann 
I also added the .odt file with the 'Translated by...' on 
https://github.com/saivann/bitcoinwhitepaper

harding:
@rodrigosilvap thanks!  We'll look at it the first chance we get, but it may take some time (@saivann and I are both very busy at the moment).

PR #697, opened 2015-01-01
Add Circle.com to the wallet page

ywecur:
Circle.com is a creditable and secure web wallet that enables instant purchase of bitcoin through both credit card transactions and U.S. bank transfers. They have existed for quite a while now and enable people to purchase bitcoin frictionlessly. I belive this warrants including them in this list.


saivann:
@ywecur Thanks for submitting a pull request.

Are you comfortable with git enough to fix the following issues?

- The screenshot is 205px large, it should be 250px.
- The environment score should be "checkpassenvironmenttwofactor".
- _translations/en.yml~ and _templates/choose-your-wallet.html~ need to be deleted (probably generated by gedit auto-backups?).
- Can you fix the typo in /abillity/ability/?
- Can you fix the typo in /sell bitcoin/sell bitcoins/?
- Can you add the missing dot at the end of the description?
- The icon is a bit pixelized, isn't there a better quality icon somewhere maybe?

As for listing Circle, I've been also thinking of adding them too. I've just checked all requirements and everything looks good as far as I'm concerned. Once the pull request is good enough, I would leave a 2 week period for other contributors to comment before merging the pull request.

One issue that could make Circle's listing a bit short-lived is that they enable HSTS only for a 24 hours period. Xapo is the only remaining wallet not supporting HSTS and I was supposed to open a pull request to submit the requirement. HSTS is a very easy change so it should probably not be hard for Circle to enable it, I'll send them an email about this in the next minutes, as well as to inform them about this pull request.

ywecur:
@saivann Thank you for reviewing my pull request. Sorry for the errors, I will fix them in a couple of hours.

ywecur:
As for the "sell bitcoin/sell bitcoins" issue, are we not at a stage to promote also buying fractions of bitcoins? 

I believe stating it as "bitcoin" is better as it does not fuel the missconception that bitcoin can only be bought in wholes.

saivann:
@ywecur Thanks! Unless I'm mistaken, "sell bitcoin" is a typo and you'd have to say "sell a bitcoin"? If you'd ask me, I think I have a preference for "sell bitcoins" because this seems more accurate to me (using bits or millibitcoins would probably be better but most people have no idea what this means and not all wallets use the same subunits). All pages linking to the wallet page, including the wallet page itself, prominently link to the "You need to know" page which explains the risks and responsibilities to the user. So in the end, I doubt this small line will really affect users spending. However if we keep it, we should probably update other wallets' texts to be fair (e.g. Mycelium: "secure your bitcoins", Coinbase: "buy and sell bitcoins").

ywecur:
@saivann If we were to make the change, "sell bitcoin" wouldn't be a typo as we are treating bitcoin as an uncountable noun e.g. "sell water", "sell bread". 

I would suggest that we do change the others to use the uncountable form as it more accurately represents bitcoins divisibility and may lead to less doubt among new users.

This is of course up to you and I will gladly change it to "bitcoins" if you chose.

saivann:
@ywecur Maybe this could do for now and I guess it's probably not worth making it a blocking issue. This said, to be honest, I doubt it's possible to treat "bitcoin" as an uncountable form in the long run given that it's used as a unit of measurement. As a matter of fact, you can always count exactly how much bitcoins you have. You're welcome if you see a better way to handle this. I really wish there was a simple subunit widely adopted...

ywecur:
I did not think about the fact that it is still widely used as a unit of measurement. The fact that Bitcoin is both the technology and a unit of measurement probably means that using the uncountable form would lead to more confusion.

As for getting a subunit adopted, maybe we should start promoting "bits" as many wallets and most of the community have already done?

saivann:
@ywecur I'm also fan of bits and I have submitted two short mentions/definitions for it previously:
https://bitcoin.org/en/vocabulary#bit
https://bitcoin.org/en/faq#wont-the-finite-amount-of-bitcoins-be-a-limitation

I guess the "bits" discussion could become a separate issue/pull req on it's own, but I guess it would make sense to use "bits" for Circle or any other wallet once they use the denomination. FWIW, it's unclear if there is a wide consensus for using bits. It's half-adopted at this point and there is still some resistance to it, so although I also think using the unit would fix many issues, using it where it likely doesn't belong to may cause unproductive discussions IMO.

ywecur:
@saivann I compleatly agree.

I have now addressed the issues you mentioned. What is the next step of the process?

saivann:
LGTM, in the absence of critical feedback, this pull request will be merged on January 15th.

@ywecur Thanks for your help. At this point other contributors can comment if they see any issue in the next two weeks. I've also sent an email to a Circle employee in case they prefer to use a different screenshot / description. Note: if you can provide a non-stretched screenshot, please feel free to update it.

JRogiest:
Using the term "wallet" loosely blurs the distinction between true "wallets" and IOU Bitcoin services.

ywecur:
@JRogiest That is why we are using the term "web wallet"

greenaddress:
@ywecur web wallet can be confusing I think.

Not all wallets that can be accessed via web are IOU Bitcoin services and vice versa not all apps give the user exclusive control.

ywecur:
@greenaddress Alright, I can see what you mean. Changing just this branch would not help though, we would have to update the entire wallet page.

saivann:
The "control over your private key" part is stated inside the score of each wallet. FWIW, hybrid wallets are also not like "typical" physical wallets. Currently the menu only lists wallets by platform compatibility, "Web" is one of these.

I understand (and mostly agree with) the idea that such services are very different than wallets... However this page is named "Choose your wallet", these services can be used like wallets, they're the first choice for many users and the "wallet" lingo is now used everywhere anyway... so I guess changing this on bitcoin.org alone would likely not make much sense if nobody else speaks the same language.

PR #706, opened 2015-01-09
Specify block halving time in regtest mode

shaulkf:


harding:
@shaulkf Thanks for finding this mistake and thanks for submitting the pull request!  (I actually didn't know the reward halved every 150 blocks, but I confirmed it's part of the code.)  I made a small edit to your change which was added as commit https://github.com/bitcoin/bitcoin.org/commit/e68d8f2f832cb943117957471d074bc569aec5bf -- it'll go live on the site in a few minutes.

Thanks again!

PR #710, opened 2015-01-12
[Bitcoin Core Downloads] Add 0.9.4 Release

harding:
Preview download page: http://dg1.dtrt.org/en/download
Preview release notes: http://dg1.dtrt.org/en/release/v0.9.4

@laanwj this adds the 0.9.4 release to Bitcoin.org. 

It also changes the OSX file name to the one now specified in upstream's [release-process.md](https://github.com/bitcoin/bitcoin/blob/v0.9.4/doc/release-process.md). (I'm assuming that's what will be used, but as soon as the binaries are added to bitcoin.org/bin, I'll check all the links.)

### DIFF
--- a/_releases/2015-01-12-v0.9.4.md
+++ b/_releases/2015-01-12-v0.9.4.md
@@ -1,0 +1,104 @@
+---
+title: Bitcoin Core version 0.9.4 released
+---
+Bitcoin Core version 0.9.4 is now available from:
+
+  <https://bitcoin.org/bin/0.9.4/>
+
+This is a new minor version release, bringing only bug fixes and updated
+translations. Upgrading to this release is recommended.
+
+Please report bugs using the issue tracker at github:
+
+  <https://github.com/bitcoin/bitcoin/issues>
+
+How to Upgrade
+===============
+
+If you are running an older version, shut it down. Wait until it has completely
+shut down (which might take a few minutes for older versions), then run the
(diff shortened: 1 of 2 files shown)


luke-jr:
I don't know if we're planning to do an actual release for 0.9.4.

harding:
@luke-jr I agree it's a little confused now based on the messages in #bitcoin-dev, but @laanwj did put some time into writing up official-looking release notes so I think he's planning a release.

In any case, I asked @saivann if I could prepare this pull so I could get some practice on the Bitcoin.org side of the release process---so even if there's no release, I learned stuff, which makes me happy. :-)

saivann:
@harding Thanks! LGTM, this pull request can simply be discarded if 0.9.4 is not released. Don't forget to update the date of the release when merging.

harding:
@saivann will do.  Thanks!

harding:
Closing this, as I don't think it's going to happen.  However, I'll leave the branch around for awhile just in case anyone wants to resurrect it.

laanwj:
Right sorry for the confusion - 0.9.4 works around a problem that isn't in the bitcoin.org-distributed executables. It was necessary to make the PPA and distro packages work again after OpenSSL 1.0.1.k.

We'll likely release 0.10 soon, as well as a 0.9.5 that will start the strict-DER softfork, so we don't want to steal their thunder by doing an intermediate 'dud' release.

harding:
@laanwj makes sense, and thanks for replying!

Issue #410, opened 2014-05-19
New Page: Describe How To Run A Full Node


harding:
As requested by Raúl Martínez and supported by @wladimir, add a page for advanced users describing how to install bitcoind as a system service and configure it (and your network) so it can act as a full peer.

I tried writing an outline for this page, but it quickly became evident that starting bitcoind as a service on Linux requires a fair amount of extra work and a lot of "if you have X, then do Y"-type descriptions covering the different ways different distributions handle init scripts.

I think it may be advantageous to delay writing this page until there's a package for Ubuntu, Windows, or OSX which lets bitcoind run as a background service.  See: https://github.com/bitcoin/bitcoin/issues/4124

Issue #698, opened 2015-01-02
Note for running nodes on OSX - disable display sleep & screensaver


bitofalefty:
Something I've noticed is that, when running a node on OSX (10.9.3) the ping time goes through the roof whenever the user account is 'locked' or the screensaver comes on, going up to about 10 seconds. I guess the machine is going into some sort of low power mode. The problem is that it isn't obvious at all to the casual node runner. Setting the display sleep to 'never' solved this problem for me and the latency is now below 100ms.

It may be worth adding a small note to the download page for OSX to disable display sleep (and computer sleep) to avoid this, if the user is interested in helping the network as much as possible.

saivann:
@harding This could go into the upcoming "Run your full node" page maybe? I guess this could be mentioned shortly for any platform, for example "Make sure your computer does not automatically go into sleep mode".

harding:
@saivann yes, I'll be sure to mention it on that page.  Thanks!

PR #711, opened 2015-01-15
Add Basic "How To Run A Full Node" Page

harding:
Preview: http://dg4.dtrt.org/en/full-node

* Add basic page with some general information plus instructions for Ubuntu 14.10.  I hope to write more instructions later, and hopefully we can find other volunteers to contribute instructions too.

* Change a link on the participate page to point to the full node page.

* This page is being added as English-only for now.  I figure we can convert it into a translation template after we have instructions for Windows and Mac OS X, and after the instrutions have actually been tested with the released version of Bitcoin Core 0.10.0.  (These Ubuntu instructions were tested with RC3 and the old Ubuntu packaging.)

Note: this page is less grandiose than I had hoped.  I realized I was procrastinating on it, so I decided to just see what I could do in an afternoon.  Suggestions for improvements welcome.

Related: issue #410


schildbach:
Awesome, thanks for writing this down!

en/full-node.md
```diff
@@ -0,0 +1,362 @@
...
+have an easy-to-use node.
+
+* Desktop or laptop hardware running recent versions of Windows, Mac OS
+  X, or Linux.
+
+* 50 gigabytes of free disk space
+
+* 2 gigabytes of memory (RAM)
+
+* Broadband Internet connection
```
saivann: Maybe "A good broadband Internet connection"?

To my understanding, many broadband connections aren't this fast and may in fact slow down the network.

en/full-node.md
```diff
@@ -0,0 +1,362 @@
...
+
+* 6 hours a day that your full node can be left running. (You can do
+  other things with your computer while running a full node)
+
+### Possible Problems
+
+* **Legal:** Bitcoin use is [prohibited or restricted in some
+  areas](http://bitlegal.io/)
```
saivann: A dot is missing at the end of the sentence.

en/full-node.md
```diff
@@ -0,0 +1,362 @@
...
+
+* **Anti-virus:** Several people have placed parts of known computer
+  viruses in the Bitcoin block chain. This block chain data can't infect
+  your computer, but some anti-virus programs quarantine the data any
+  way, making it more difficult to run a full node. This problem mostly
+  affects computers running Windows
```
saivann: A dot is missing at the end of the sentence.

en/full-node.md
```diff
@@ -0,0 +1,362 @@
...
+
+* **Attack target:** People who want to disrupt the Bitcoin network may
+  attack full nodes in ways that will affect other things you do with
+  your computer, such as an attack that limits your available download
+  bandwidth or an attack that prevents you from using your full node's
+  wallet for sending transactions
```
saivann: A dot is missing at the end of the sentence.

en/full-node.md
```diff
@@ -0,0 +1,362 @@
...
+
```
saivann: May I suggest adding the following (or similar)

\*\*Bandwidth limit\*\*: Some Internet plans will charge an additional amount for any excess upload bandwidth used that isn't included in the plan. You may want to check if your Internet connection is subjected to such limitations or monitor your bandwidth consumption.

en/full-node.md
```diff
@@ -0,0 +1,362 @@
...
+<div class="box" markdown="1">
+*Optional: Start Your Node At Login*
+
+Starting your node automatically each time you login to your computer
+makes it easy for you to contribute to the network. The easiest way
+to do this is to add Bitcoin Core GUI to the list of startup
+applications.
+
+Click the Ubuntu swirl icon to start the Dash, type `startup app`,
+and click the Startup Applications icon.
```
saivann: Bitcoin Core has a "Start Bitcoin Core on startup" option in it's own preferences box on Linux. This could maybe cut a few steps and be more compatible with other Linux distributions?

en/full-node.md
```diff
@@ -0,0 +1,362 @@
...
+
+#### Bicoin Core Daemon {#ubuntu-daemon}
+
+Before using the Bitcoin Core daemon, `bitcoind`, you need to create its
+configuration file with a user name and password. First create the
+`.bitcoin` directory, create (touch) the file, and set the file's
+permissions so that only your user account can read it.  From the
+terminal, type:
+
+    mkdir .bitcoin
```
saivann: Maybe these commands should be prefixed with "~/" just in case?

en/full-node.md
```diff
@@ -0,0 +1,362 @@
...
+    rpcuser=bitcoinrpc
+    rpcpassword=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+    (you do not need to remember this password)
+    The username and password MUST NOT be the same.
+    If the file does not exist, create it with owner-readable-only file permissions.
+    It is also recommended to set alertnotify so you are notified of problems;
+    for example: alertnotify=echo %s | mail -s "Bitcoin Alert" admin@foo.com
+
+The "rpcpassword" displayed will be unique for your system.  You can
+copy the the rpcuser and rpcpassword lines into your configuration file
```
saivann: s/the the/the/

en/full-node.md
```diff
@@ -0,0 +1,362 @@
...
+
+### Possible Problems
+
+* **Legal:** Bitcoin use is [prohibited or restricted in some
+  areas](http://bitlegal.io/)
+
+* **Anti-virus:** Several people have placed parts of known computer
+  viruses in the Bitcoin block chain. This block chain data can't infect
+  your computer, but some anti-virus programs quarantine the data any
+  way, making it more difficult to run a full node. This problem mostly
```
saivann: s/any way/anyway/?

en/full-node.md
```diff
@@ -0,0 +1,362 @@
...
+lang: en
+id: full-node
+title: "Running A Full Node - Bitcoin"
+---
+
+{% assign moreHelp="If you need more help, please ask in one of Bitcoin's many [communities](/en/community), such as [Bitcoin StackExchange](https://bitcoin.stackexchange.com/), [BitcoinTalk technical support](https://bitcointalk.org/index.php?board=4.0), or the [#bitcoin](https://webchat.freenode.net/?channels=bitcoin&uio=d4) IRC chatroom on Freenode." %}
+
+# Running A Full Node
+
+<p class="summary">Learn how run a full node</p>
```
saivann: Maybe making it a bit longer "Learn how run a full node on your computer or server." would be more readable since this string otherwise looks like another title between two titles?

saivann:
@harding Thanks! Very useful page to have, I just reported what I found, maybe I'll have time to review it more carefully a bit later. I've noticed there is no mention of opening port 8333? I wonder if the page should be translated, 25 languages is time-consuming and there is a lot of words there, but it's an important page too.

en/full-node.md
harding: Sure, I'll change that.  

For reference, I just checked the stats on my fastest and best connected node (with >99% uptime) and it says the average for the last 30 days is 592 Kbit outgoing and 63 Kbit incoming.  That outgoing is may be a bit high for a low-end DSL connection around here, but the average is skewed by the fact that my connection supports up to 100 Mbit/s and regularly sees peaks of about 14 MBit/s---much higher than supported by most home connections around here.

bitcoind3

Based on that, I think pretty much any broadband connection would be an asset to the network.  However, a connection with low outgoing speed will more commonly get saturated, giving a poor experience to the volunteer---so I think saying "a good connection" is warranted.

en/full-node.md
harding: It's a single-paragraph bullet point, so I don't end it with a period.  See the other bullet points in this list and other lists in the doc.  I open to suggestions to change the style, although I think we had a similar discussion before about sentences in tables where we decided to omit the trailing period.  (I don't care either way, as long as we agree.)

en/full-node.md
harding: Oh, wow, I didn't know that.  I just tried it and it did exactly the same thing my instructions did.  I'll revise to describe using that feature.  Thanks!

harding:
@saivann thanks for your comments!  I replied to a few; the rest I plan on implementing as you propose.

Yeah, about half a second after I clicked "Submit pull request", I realized I forgot about port 8333. :-)  I'll write something about it tomorrow.

AFAICT, translation has a lot of things going against it:

* **High word count:** currently about 2,000.  It'll likely be over 5,000 words by the time we add Ubuntu Server, generic Linux, Windows (7 and 8?), and Mac OS X instructions.

* **Screenshots:** we'd need to take new screenshots for each language.  That may not be hard for systems I can virtualize (like Ubuntu), but it does sound time consuming, and it may be impossible for systems I don't have any current access to (like OS X).

* **Frequent changes:** this guide will probably need to be partially rewritten for each Bitcoin Core release, with little rewrites needed for minor releases (every month or two) and larger rewrites needed for major releases (every six months or so).  It also has to be updated when the underlying operating systems change.  That means there will probably frequent translation updates.

It seems to me like it's probably better to put off translations for the short term until we see the page in action.   That way, if nobody ever visits the page or it ends up changing too often, we don't bother to even try to get it translated.

@schildbach My pleasure!

en/full-node.md
saivann: @harding Oops, if that's the case I don't remember. In only thought a dot was necessary as soon as we're including a verb. In either case, as long as it's not a typo, it's OK with me.

harding:
* Commit 702a8a9b7 replaces the original commit. The only change is that I deleted two of the images related to GUI autostart.  New images are part of the next commit.

* Commit 40bbde2e6 makes numerous corrections suggested by @saivann (thanks!). See the commit message for details.

    It also adds an extended section describing setting up port forwarding on home routers and opening firewall ports to allow inbound connections to port 8333.

The [preview](http://dg4.dtrt.org/en/full-node) has been updated.

I hope to get a Windows 7 VM setup tonight so I can write Windows instructions tomorrow.

jlopp:
Wow; great work. I'm wondering how you came to decide that 6 hours per day is the minimum requirement? The problem with nodes that don't run 24/7 is that they don't become well-connected since their IP address is constantly being purged from the address lists held be other nodes on the network. Gavin has also noted that we want nodes that are always on: https://www.reddit.com/r/Bitcoin/comments/1scd4z/im_running_a_full_node_and_so_should_you/cdw3lrh?context=3

harding:
@jlopp I used the highly scientific method called guessing.  :-)  

Based on monitoring my own full node stats---including the intermittent-on devel node on my laptop---I know that it typically takes 30 to 60 minutes of node uptime before the seeders start sending me BitcoinJ-based SPV clients.  Five hours of serving SPV clients seems like a reasonable contribution to the network to me.

In addition, based on my `getnettotals` results (I run it in an every-minute cronjob), I also know that my node tends to upload about 10 times as much as it downloads, and that it reaches that ratio in about 2 to 3 hours and then remains fairly constant.  Doubling that time to account for node variations also comes out to about six hours.

I actually think a lower value than 6 hours a day would still be a positive contribution to the network.  I specified higher requirements than strictly necessary in all categories, including uptime, because there was some concern about contributors with marginal resources becoming disappointed and making comments that would hurt volunteer morale.

I agree with Gavin's comment that we need more always-on full nodes.  However, those nodes haven't appeared in the year since he made that comment---so what are we to do?

harding:
Commit 83c419ebd1 adds instructions for Windows 7, Ubuntu Server, and Other Linux Distributions---see the complete list of changes below. This is my final planned commit for the full-node page, although I will re-review when 0.10.0 final is released. The [preview](http://dg4.dtrt.org/en/full-node) has been updated.

* Added instructions for Windows 7, but only for Bitcoin Core GUI. I added a stub for anyone who wants to write instructions for using the daemon on Windows.

* Added instructions for Ubuntu 14.04 LTS Server, but only for Bitcoin core daemon. I assume most server users run headless.

* Added instructions for Other Linux Distributions, for both GUI and daemon. Hopefully the instructions are general enough to apply to most distributions but specific enough that they actually help readers.

* Added a stub for Windows 8.1 as I don't have access to a copy, and all the pay-per-hour Windows VPSes I can find run some version of Windows Server. (I have the same problem with OS X.)

* Hid some subsections in the table of contents: I found having subsections named "Bitcoin Core GUI" and "...Daemon" within multiple sections distracting, so I hid them in the TOC.

* Added basic PGP verification instructions: I didn't try to explain PGP to newbies, but I did provide instructions useful to people who have used PGP before.  These instructions are currently displayed in the Windows 7 and Other Linux Distributions sections (where users download from Bitcoin.org).

* Made sure the end of each install section points to the Network Configuration section so users open port 8333.

saivann:
@harding Maybe you could still mention something like "Ideally, you should keep your PC running. However, 6 hours a day remains a good minimum"?

harding:
@saivann, @jlopp added commit 9940b07 mentioning that more hours are better, and that running continuously would be best of all.  [Preview](http://dg4.dtrt.org/en/full-node) updated.

jlopp:
ACK.

This will be a great resource and I'll do what I can to get it added to as many FAQs as possible. From the discussions I've had with devs, it seems that the best hope we have of getting more people to run nodes is to make it easier. Getting rid of the need for bootstrapping is a good start; hopefully pruning will be another step in the right direction. But the ideal scenario would be for it to be straightforward enough that a guide such as this isn't even necessary :)

harding:
@jlopp Thanks!  And I would love to see this guide become unnecessary.

jlopp:
*Bump* will this be getting merged and pushed soon?

harding:
@jlopp Since these instructions are specific to Bitcoin Core 0.10.0, which hasn't been released yet, I'm not planning on merging this until 0.10.0 is released on Bitcoin.org and the Ubuntu PPA repository has been updated.

(However, unless there are any changes to 0.10.0's install procedures or someone finds a problem on the page, I am finished writing.)

saivann:
@harding I have just read through the whole document once again and it's awesome! Thanks for your extensive research and step-by-step instructions!

The only remaining suggestion I have is about dropping some duplicate content. What would you think about having only one Ubuntu version (Ubuntu 14.04 or Ubuntu 14.10), with graphical and command line instructions? At a first glance, the only difference I have found between the duplicate command line instructions is how to access the terminal when installing Bitcoin Core (one over SSH, one with the swirl icon). Maybe this can be merged together?

harding:
@saivann good idea re: Ubuntu instructions---I'll do that.  Thanks for the complement!

harding:
Commit b29ab4a54 merges the two different Ubuntu instructions together as suggested by @saivann.  [Preview](dg4.dtrt.org/en/full-node#ubuntu-1410) updated.

saivann:
LGTM, Thanks!

harding:
Commit aaab53f adds a quick note about screensavers activating a low-power mode that prevents nodes from uploading data.  This was reported in issue #698 but I forgot about it.  (Sorry!)

harding:
Note: 0.10.0 packages have been uploaded for Ubuntu 14.10, so I'll be retesting these instructions tomorrow.  I'll also work on revising the block chain size text as discussed in #749 

I'm aiming for merging this Monday morning EDT (noonish UTC).

harding:
Rebased and added commit f61f66d which adds a link to this guide from the Download page.  Also updated the preview.

In the absence of critical feedback, this will be merged around 12:00 UTC Monday.

chepkasov:
Hi, Harding and All!

Unfortunately, websites are still not published. Why is that? Can you help me?

We're needed to add bitcoin news website: http://www.coinfox.info (ENG), http://www.coinfox.ru (RU) and http://www.coinfox.es (ES) for https://bitcoin.org/ru/resources ("News" maybe)


Sincerely,

Igor Chepkasov
Cryptocurrency Foundation Russia, Founding chairman

PR #713, opened 2015-01-18
Fix statement about "majority of miners"

kanzure:
Instead of talking about a majority of miners, it is better to talk about a majority of available hashrate. There is no way to determine the number of miners that exist, and it is confusing to talk about a "majority of miners" as a consequence of this and other reasons.

The original version of the statement was:

> colludetxt2: "However, a majority of miners could arbitrarily choose to block or reverse recent transactions. A majority of users can also put pressure for some changes to be adopted. Because Bitcoin only works correctly with a complete consensus between all users, changing the protocol can be very difficult and requires an overwhelming majority of users to adopt the changes in such a way that remaining users have nearly no choice but to follow. As a general rule, it is hard to imagine why any Bitcoin user would choose to adopt any change that could compromise their own money."

And this changes it to say:

> colludetxt2: "However, a majority of miner hashrate could arbitrarily choose to block or reverse recent transactions. A majority of users can also put pressure for some changes to be adopted. Because Bitcoin only works correctly with a complete consensus between all users, changing the protocol can be very difficult and requires an overwhelming majority of users to adopt the changes in such a way that remaining users have nearly no choice but to follow. As a general rule, it is hard to imagine why any Bitcoin user would choose to adopt any change that could compromise their own money."

There may be other errors lurking in this block.


harding:
LGTM, although as a nitpick, I'd say "hash rate" instead of "hashrate".

kanzure:
> LGTM, although as a nitpick, I'd say "hash rate" instead of "hashrate".

Agreed, I will make that change. Btw, how do you feel about "a majority of network-total miner hash rate"? I am trying to make this maximally unambiguous....

kanzure:
Hmm, talking about hash rate is probably ambiguous here anyway. Really this could be simplified by instead talking about proof-of-work. Unfortunately this just boils down to "valid proofs of work are valid and correct nodes will accept valid proofs of work that meet the other requirements for blockchain tip acceptance".

saivann:
Does this sound better: "Miners controlling the majority of the hash rate..."? Or if "hash rate is too "technical", it could be "computing power".

harding:
@saivann since even a powerful minority could make the attack described in that parapgrah, we could probably just simplify down to: "However, powerful miners could..."

saivann:
@kanzure Thanks for updating your branch, but I guess you accidentally used "powerful mining" instead of "powerful miners"?

kanzure:
> but I guess you accidentally used "powerful mining" instead of "powerful miners"?

You caught me! (fixed)

saivann:
@kanzure No problem :) Thanks for the pull request!

PR #714, opened 2015-01-19
Order release history and set latest version using versions number instead of releases date

saivann:
[This comment](https://github.com/bitcoin/bitcoin.org/pull/710#issuecomment-70415594) outlined that releases numbers might not always be in chronological order (e.g. 0.9.5 to be released after 0.10.0) which would affect the version served on the download page as well as the order on the version history page.

This pull request should prevent this issue as long as version numbers do not use more decimal places or digits than this example "000.000.000.000.000" .

In the absence of critical feedback, this pull request will be merged on January 21th.

### DIFF
--- a/_plugins/releases.rb
+++ b/_plugins/releases.rb
@@ -20,6 +20,7 @@ require 'yaml'
 module Jekyll
   
   class ReleasePage < Page
+
     def initialize(site, base, lang, srcdir, src, dstdir, dst, year, month, day)
       @site = site
       @base = base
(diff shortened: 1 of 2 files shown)


harding:
@saivann glad you caught that for the downloads page!

I don't think it's a big deal either way, but it seems to me like the [release history page](https://bitcoin.org/en/version-history) should be in chronological order.

saivann:
@harding This would actually create a "chronological ordering per major release". All 0.9.* releases would be together in chronological orders, followed by all 0.10.* releases, etc.

harding:
@saivann makes sense.  Thanks!

Issue #715, opened 2015-01-20
Update Developer Reference to include bitcoind REST API


jlopp:
See https://github.com/bitcoin/bitcoin/blob/master/doc/REST-interface.md

harding:
@jlopp thanks for opening the issue!  I should be finishing up pull #711 tonight, and the REST docs are next on my todo list.

PR #720, opened 2015-01-26
Dev Docs: Add REST Documentation

harding:
Preview: http://dg1.dtrt.org/en/developer-reference#http-rest

This adds documentation for the new REST API in Bitcoin Core 0.10.0.  It uses the same formating we use for the [JSON-RPC documentation](https://bitcoin.org/en/developer-reference#remote-procedure-calls-rpcs) with the section headings formatting based on [Twitter's REST docs](https://dev.twitter.com/rest/public).  (As this is the first REST docs I've written, I looked for examples and liked theirs the most.)

Feedback on content or formatting appreciated.

CC: @jlopp (requested docs in issue #715), @jonasschnelli (did a lot of work on the REST interface)


jonasschnelli:
Nice work!
- Would it make sense to mention that `-txindex` needs a `-reindex=1`? Because from time to time there are question like this: https://github.com/bitcoin/bitcoin/issues/5667.
- Maybe it would be good to tell the users once again that they should NOT allow public access to the bitcoind http/rest server

ACK.

harding:
@jonasschnelli good ideas!  I'll make both those changes momentarily.  Thanks for your review!

jlopp:
ACK

harding:
Commit 73eb097 adds both notes suggested by @jonasschnelli, adding them to the relevant parts of both the REST and JSON-RPC docs.  [Preview](http://dg1.dtrt.org/en/developer-reference#http-rest) updated.

In the absence of critical feedback, this pull will be merged around 20:00 UTC Tuesday.

PR #721, opened 2015-01-26
Dev Docs: Mention Upcoming v3 Blocks/BIP66

harding:
Preview: http://dg2.dtrt.org/en/developer-reference#block-versions

Briefly mention upcoming v3 blocks and BIP66 (strict DER).  We don't currently have any docs about creating signatures, so I don't think there's anything else we need to change.

Note: the link to BIP66 is broken because Greg Maxwell just assigned the BIP number. I won't merge this request until the link works.

### DIFF
--- a/_autocrossref.yaml
+++ b/_autocrossref.yaml
@@ -72,6 +72,7 @@ data-pushing op code:
 data-pushing op codes: data-pushing op code
 denomination:
 denominations: denomination
+DER:
 DER format: der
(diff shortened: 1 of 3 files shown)


saivann:
LGTM. The link to BIP66 is now working.

harding:
@saivann thanks!

In the absence of critical feedback, this will be merged around 22:00 UTC today.

PR #727, opened 2015-02-02
Dev Docs: Deprecate `gethashespersec` rpc call in 0.10.0.

ruimarinho:
This was recently removed on bitcoin core. See https://github.com/bitcoin/bitcoin/pull/5599 for details.

### DIFF
--- a/_autocrossref.yaml
+++ b/_autocrossref.yaml
@@ -544,3 +544,4 @@ Bitcoin Core 0.9.0:
 Bitcoin Core 0.9.1:
 Bitcoin Core 0.9.3:
 Bitcoin Core 0.10.0:
+Bitcoin Core master:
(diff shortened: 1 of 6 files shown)


harding:
@ruimarinho thanks for the pull request!  It looks like `gethashespersec` is still part of 0.10---it's only been removed in master.  It's also not in the list of [pending pull requests for 0.10](https://github.com/bitcoin/bitcoin/pulls?q=is%3Aopen+is%3Apr+milestone%3A0.10.0).

I'm fine with moving it to the "Removed RPCs" section of the quick ref, but I think that we should say that it's still in 0.10.0---unless that changes.  First thing tomorrow, I'll send you a pull request against your branch with this change plus some other minor details.

Thanks again!

ruimarinho:
You're right, nice catch. Thank you for the fixes!

harding:
@ruimarinho thanks again for the pull!

In the absence of critical feedback, this branch will be merged around 18:00 UTC Tuesday.

PR #728, opened 2015-02-02
Remove "France Bitcoin" from resources page

louisjc:
"[France Bitcoin](http://francebitcoin.com/)" should be moved from Useful places to the News section, since it's only a news website.
Edit : Maybe it should even be deleted as the last post is almost 1 year old...


saivann:
@louisjc Actually this website seems unmaintained (last post on March 12th), how about just removing the link? Other contributors also previously made that suggestion.

louisjc:
@saivann Agree, I edited my comment with this just before your answer.

saivann:
In the absence of critical feedback, this pull request will be merged on February 4th.

PR #731, opened 2015-02-04
Add Coinprism to the wallet list

Flavien:
This pull request adds the [Coinprism Android mobile wallet](https://github.com/Coinprism/android-wallet) to the list of wallets.


Flavien:
Any update on the review?

harding:
@Flavien we're still heavily backlogged on reviews, but there is now a volunteer working on it.  We'll let you know if there's anything we need from you.  Thanks for your patience.

harding:
@Flavien thank you for your patience; we've begun reviewing Coinprism.

If I understand correctly, Coinprism reuses the same address over and over, even sending change back to it.  This seems to be not only the default behavior, but also something that the user can't disable.  (Please let me know if I misunderstand.)

We don't want to promote any wallets that encourage address use.  This is currently a soft requirement for the site, but we're almost certainly going to make it a hard requirement when Multibit HD is released in a few weeks.

How do you feel about us putting our review on hold until Coinprism upgrades to not reuse addresses by default?

Thank you again, and sorry about the wait.

PR #733, opened 2015-02-04
Small error in the French translation (Bitcoin instead of Bit)

louisjc:
There is a small error in the French translation of the [Vocabulary](https://bitcoin.org/fr/vocabulaire#bit) page:
"**Bit** is a common unit" is translated into French as "**Bitcoin** est une unité courante"


saivann:
@louisjc Indeed, thanks for fixing this issue!

Issue #734, opened 2015-02-05
Addition to resources page


FrancisPouliot:
If possible I'd like to add two resources that a team of 15 volunteers led by myself have been working on since last summer.

In directories: the Bitcoin Donation Directory, the most complete list of non-profit projects that accept Bitcoin Donation http://givebtc.org/index.php/bitcoin-donation-directory/

Learning resources: the Bitcoin Handbook for Non-Profits, self-explanatory title:http://www.givebtc.org/GiveBTC_-_Handbook_for_Non_Profits.pdf

Both are located at givebtc.org

Thanks !

harding:
@FrancisPouliot I just browsed the directory and skimmed the PDF---both look very nice.  (And I'm particularly impressed by the depth of content and attractive formatting of the PDF.)  Thanks for all of your work!

@saivann I think we can add the donation directory to the directory part of the [resources page](https://bitcoin.org/en/resources), but I don't know where we could put a link to the PDF.  Do you have any ideas?  (I'd be happy to prepare the pull request once we decide what to do.)

saivann:
@FrancisPouliot Thank you very much for your hard work on this! Looks professional and well written, at least for the parts I have reviewed. And the directory also seems extensive.

@harding Agreed, thanks! I actually just opened the pull request #738 (I used what @FrancisPouliot suggested in his first comment, which I guess is the most appropriate for now).

PR #738, opened 2015-02-06
Add links to givebtc.org

saivann:
As suggested in #734

In the absence of critical feedback, this pull request will be merged on February 8th.

### DIFF
--- a/_templates/resources.html
+++ b/_templates/resources.html
@@ -22,6 +22,7 @@ id: resources
     <p><a href="http://usebitcoins.info/">{% translate linkmerchants %}</a> - usebitcoins.info</p>
     <p><a href="http://howtobuybitcoins.info/">{% translate linkexchanges %}</a> - howtobuybitcoins.info</p>
     <p><a href="https://en.bitcoin.it/wiki/Category:Shopping_Cart_Interfaces">{% translate linkmerchantstools %}</a> - en.bitcoin.it</p>
+    <p><a href="http://givebtc.org/index.php/bitcoin-donation-directory/">{% translate linknonprofits %}</a> - givebtc.org</p>
   </div>
 </div>
 <div>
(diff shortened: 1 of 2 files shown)


harding:
LGTM.  Thanks!

luke-jr:
Page 23 misexplains Bitcoin in an attempt to go into unreasonably technical details and encourages address reuse. It's not clear how to get this resolved. While it may arguably not be a direct security risk (worst case scenario, losing all the funds, it's the same as if nobody donated in the first place), not even explaining the risks being taken on may lead it to be used outside of the charity context.

saivann:
@luke-jr By "unreasonably technical details" I guess you are referring to mentioning "private keys"? Maybe you could make some suggestions in case they plan on updating the document? I guess it'll be quite hard to discourage address reuse for the specific case of printed materials and websites, but indeed, maybe they can consider the idea of adding more details on the lack of privacy, and why reusing addresses in other scenarios may not be very suitable.

@FrancisPouliot Please see @luke-jr comment above in case you want to comment.

luke-jr:
Yeah, end users should never be exposed to private keys, and should see the wallet as an opaque entity/file itself. It makes false equivalences between bitcoin addresses and email addresses, as another example. Probably those paragraphs should be rewritten from the perspective of "you can generate an address - this is used to receive a single transaction generally, but when you don't care who is sending it, you can often get away with just publishing it for everyone to use (although this is still a bad idea for security and privacy concerns)"

harding:
@luke-jr do you consider this a blocking issue, or is it ok to merge this commit?

FrancisPouliot:
@luke-jr can't really have it both ways, i.e. say it's too technical and then say that I don't explain address re-use. Later in the document there is a specific section about wallets written by Michael Perklin on the topic. 

I don't make false equivalences between an email address and a Bitcoin address - it's just a very basic security tip, i.e. keep your private key safe, with an everyday example: "You can think of them as your email address and your email password: you share your email address with others to receive mail and you use your password to view or send email."

This document will never be perfect to satisfy everyone - I had to make editorial choices as to how I would explain some things considering the target audience.The topic of addres re-use is covered later in the doc (see p. 29) and the nature of a Bitcoin address is covered before (p.10 and 11)

harding:
Thanks @FrancisPouliot for your clarification!

I'm guessing that @luke-jr doesn't think the page 23 problem is a blocking issue, so---unless I hear otherwise from him---I plan to merge this commit in 24 hours (around 15:00 UTC Wednesday).

PR #750, opened 2015-02-17
fixed grammar error: have -> has

merland:


harding:
@merland thanks!  I merged this fix, and it should appear on the site within about 15 minutes.

PR #755, opened 2015-02-19
Reduce future changes to the javascript code of the download page

saivann:
I have been working with the help of @harding on allowing the binaries and download page to be closely monitored by scripts. One remaining issue I had was the number of alerts triggered by unimportant changes in the javascript code. This pull request aims to reduce this useless noise by two means.

By only loading the few necessary javascript functions required for this page to work.
By not changing the javascript code on the page on each release.

I'll be happy to wait and rebase this pull request if needed given that the download page is receiving a lot of changes recently.

Edit:

Live preview: http://bitcointest1.us.to/en/download
Result diff: http://bitcointest1.us.to/diff.html


harding:
@saivann I like it!  I haven't tested, but it LGTM.  I'll rebase my own work on top of this since what you have is cleaner and since I haven't started my cherry picking yet.  Thanks!

saivann:
@harding Thanks! To my testing here it works just fine, I've made javascript alert the links just to make sure they were correct and tested the resulting site on two platforms. Other pages work correctly too.

saivann:
In the absence of critical feedback, this pull request will be merged on February 21th.

PR #758, opened 2015-02-20
Dev Docs: Add Inline Template Plugin And Use It For RPC/REST

harding:
This pull adds a new plugin that allows converting inline YAML to output formatted by a standard Liquid template.  This pull also reformats the RPC/REST API tables to YAML and then uses a template to create the *exact* same tables as before.  This is verified: diffing the site built on master HEAD against the site built on this final commit shows zero changes.

(No preview provided for the above reason.)

This backend code should make it easier to implement alternative table layouts as discussed in pull #757.

The YAML input for a table looks like this, with n, t, p, and d standing for name, type, presence, and description in the output table.

    {% itemplate ntpd1 %}
    - n: "Keys Or Addresses"
      t: "array"
      p: "Required<br>(exactly 1)"
      d: "An array of strings with each string being a public key or address"

    - n: "→<br>Key Or Address"
      t: "string"
      p: "Required<br>(1 or more)"
      d: "A public key against which signatures will be checked.  Alternatively, this may be a P2PKH address belonging to the wallet---the corresponding public key will be substituted.  There must be at least as many keys as specified by the Required parameter, and there may be more keys"
    {% enditemplate %}

The first commit in this pull adds the plugin and the template, and also adds some necessary newlines to the vars.md file.  The second commit just runs the sed script provided in its commit message.

### DIFF
--- a/_includes/helpers/vars.md
+++ b/_includes/helpers/vars.md
@@ -3,16 +3,19 @@ This file is licensed under the MIT License (MIT) available on
 http://opensource.org/licenses/MIT.
 {% endcomment %}
 
-{% capture INCLUDE_F_LIST_TRANSACTIONS %}| {{DEPTH}}<br>`involvesWatchonly` | bool           | Optional<br>(0 or 1)        | *Added in Bitcoin Core 0.10.0*<br><br>Set to `true` if the payment involves a watch-only address.  Otherwise not returned
+{% capture INCLUDE_F_LIST_TRANSACTIONS %}
+| {{DEPTH}}<br>`involvesWatchonly` | bool           | Optional<br>(0 or 1)        | *Added in Bitcoin Core 0.10.0*<br><br>Set to `true` if the payment involves a watch-only address.  Otherwise not returned
 | {{DEPTH}}<br>`account`        | string            | Required<br>(exactly 1)     | The account which the payment was credited to or debited from.  May be an empty string ("") for the default account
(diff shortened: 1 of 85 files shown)


saivann:
@harding Untested LGTM, thanks! Did you test using plain HTML to see if it worked fine with a colspan? I wonder if autocrossreflinks would still apply fine within tables?

harding:
@saivann thanks!  (Reply under the announcement below.)

In the absence of critical feedback, this will be merged in about 48 hours. (00:00 UTC Monday)

> Did you test using plain HTML to see if it works fine with a colspan?

Here's what [this quick commit](https://github.com/harding/bitcoin.org/commit/acbad58227e39763713aa63e2a44bbf0aec18757) renders as (note the autoxref links):

2015-02-20-185427_629x404_scrot

Note that I haven't added any CSS, so that's the default table style.  (I just realized the Kramdown-created tables use thead and tbody, but my table doesn't, which is why it looks different from the other tables in the docs.  I'll fix that and make other changes.)

I do have to say that editing a Liquid template is a heck of a lot easier than messing around with advanced CSS features, so I think this plugin really is the way to go.  Thanks for pushing me to do better on #757!

saivann:
@harding That all sounds excellent and clean to me!! Thanks for taking time to find the ideal solution - this can become quite hard over time.

I've also played with flexbox a little in the past and found it pretty great, the persisting lack of browser support has always been the main blocking issue. I am glad to see it's almost an issue of the past.

PR #759, opened 2015-02-20
Updated bitcoin market cap (date and value)

merland:


harding:
Hi, @merland.  Thanks for all the text updates lately!  I have two comments here:

1. If you have the time, could rephrase that entire sentence to make it less specific (and remove the date entirely) so we won't need to update it in the future?  That would help save the translators a lot of time (we have about 25 translations that need to be updated each time a string changes).

2. It looks like your repository's master branch has fallen out of sync with our master branch, which is what is creating the extra commits shown above.  This isn't a major problem for now---I can always cherry-pick your important commits---but you may want to read [this StackOverflow page](http://stackoverflow.com/questions/7244321/how-to-update-github-forked-repository) about updating your master branch.

Thanks again!

merland:
Thanks for the feedback. I am well aware of the the problem with the translations, good idea about rephrasing to something more persistant. I'll give it a try.
Yes, I did manage to mess up something about my repo.... I'll do my homework till next time!
Thanks for your work.

PR #761, opened 2015-02-21
Update some images

louisjc:
I've updated some images:
* The chart and merchants map with more recent data
* The resiliency image from weusecoin with the new version


harding:
@louisjc thanks for the update!  Also, thanks for taking the time to change the years embedded in all the translations.

For the price chart, I think we should keep your commit 5732a62 changing all the translations, but also add an extra commit changing the English translation to "Bitcoin price over time".  (Maybe you also want to change the translation for any language you speak well to say the same thing.)  Once that gets translated, we won't have to update the text again the next time we update the images.  Thanks again!

(I'll check over the images more carefully when I get home to make sure they're compressed.)

saivann:
I'd like to suggest the resiliency image is not updated; it is not explicit what license applies to the new video, and it is likely CC BY-ND 4.0 (No deritative).

louisjc:
@saivann Yep you're right, I had not thought that the license could have changed.
@harding The images where not compressed, I fixed it. If we change the text to be independent of the date, should we hotlink the image from bitcoincharts.com so it would be updated in real time ?

harding:
@louisjc thanks for compressing the images!

I'd prefer to manually update the image every 6 or 12 months rather than hotlinking it.  That seems like the polite thing to do, as well as ensuring the FAQ page displays correctly even if BitStamp is temporarily unavailable.

Thanks!

louisjc:
@saivann I've removed the resiliency image commit.
@harding I've added the extra commit you suggested (in English and French).

harding:
@louisjc merged.  Thanks again!

PR #767, opened 2015-02-24
Fix typos "Bicoin" -> "Bitcoin"

anduck:


harding:
@anduck oh, wow.  I can't believe we missed that.  Thanks so much for fixing it!  The site should be updated within about 15 minutes.

PR #770, opened 2015-02-24
Bitcoin for Developers: Change links

louisjc:
I find misleading that the [Bitcoin for Developers](https://bitcoin.org/en/bitcoin-for-developers) page has the same "Get started with Bitcoin" button as the other pages but with a different link (english only, even if the button is translated).
I propose:
   1. Restore the same link for the button as the other pages ([Getting started](https://bitcoin.org/en/getting-started)).
   2. Add a Developer Documentation link without translating: I think this is not a problem / more indicative because the documentation is English only.

### DIFF
--- a/_templates/bitcoin-for-developers.html
+++ b/_templates/bitcoin-for-developers.html
@@ -26,4 +26,5 @@ id: bitcoin-for-developers
 <h2><img class="titleicon" src="/img/icons/ico_micro.svg" alt="Icon" />{% translate micro %}</h2>
 <p>{% translate microtext %}</p>
 
-<div class="mainbutton"><a href="/en/developer-documentation"><img src="/img/icons/but_bitcoin.svg" alt="icon">{% translate getstarted layout %}</a></div>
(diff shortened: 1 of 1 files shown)


harding:
@louisjc good idea!  Thanks!  For everyone else, here's what it looks like:

2015-02-24-075622_630x241_scrot

I think having a lonely untranslated string might be a bit confusing, so I think the docs link should say, "Developer Documentation (English)".  Of course, that looks weird when you're on the English translation, so we should probably put in a conditional like this:

    {% case page.lang %}
      {% when 'en' %}
        (link text "Developer Documentation")
      {% else %}
        (link text "Developer Documentation (English)")
    {% endcase %}

What do you think?

louisjc:
Good idea, is there a reason not to do directly:
```
<a href="...">Developer Documentation {% if page.lang != "en" %}(English){% endif %}</a>
```

harding:
@louisjc indeed, that direct way is better!

louisjc:
@harding I've rebase added it!

harding:
Tested both English and other languages, and it LGTM (screenshot from German below).  If nobody else has any comments, I'll merge this tomorrow.  Thanks!

2015-02-24-092056_548x242_scrot

PR #774, opened 2015-02-25
SVG optimisation; the return

louisjc:
As suggested by @saivann in https://github.com/bitcoin/bitcoin.org/pull/726, I've optimized every icon SVG file. I've used [this batch tool](https://github.com/petercollingridge/SVG-Optimiser) with non destructive options plus some manual fixing. The images should be visually identical and according to my tests on firefox, they are (double check may be required). They are now 3 times smaller, it should save some bandwidth.

*NB: For futures images, the [Javascript version](http://petercollingridge.appspot.com/svg-editor) should be enough.*


harding:
According to my tests, the optimized images save 251,227 bytes is disk usage.  Here's is a single image of all the current SVGs from my browser:

2015-02-25-121417_688x652_scrot

And here is an image of all the optimized images:

2015-02-25-121347_695x654_scrot

The only differences I see are a slight difference in color tone in the RSS icon and the !-inside-triangle warning images.  I don't think that matters, so this pull LGTM.  Thanks for another great PR, @louisjc!

saivann:
@louisjc Thanks, great work!

There is some compatibility issues with Safari 4, the screenshot below shows how many icons on the https://bitcoin.org/en/innovation page are rendering (some parts of the icons do now show). I do not have higher Safari versions to test how widespread this issue would be yet, any idea of a way of fixing this? Or does anyone have time to test higher versions of Safari?

capture du 2015-02-25 13 45 22

louisjc:
@saivann It works fine with Safari 8 on OS X 10.10. 
I would like to debug it but I think there is no way to get an older Safari...

saivann:
@louisjc Thanks for your additional testing! I've found a Safari 5.0 version and couldn't reproduce this issue with it either, so I'd say it probably affects an almost inexistent number of browsers. LGTM!

harding:
Thanks both @louisjc and @saivann for your additional testing.

In the absence of critical feedback, this will be merged around 22:00 UTC Thursday.

louisjc:
How about add Safari 4 in the fallbackSVG ?
Safari 4 don't seems to support window.SVGSVGElement, so :
```
function supportsSVG() {
...
+ if (!(window.SVGSVGElement)) return false;
...
}
```
Should do it. Again I can't test it on Safari 4, all i can say is that it only trigger on IE version already under the fallback, and do not trigger on recent FF, Safari and Chrome.

saivann:
If Safari 4 is the only affected browser, I suggest we postpone the javascript improvement for when we'll need to make other changes to base.js, since editing base.js will trigger an alert for the people monitoring the download page. Safari 4 is virtually extinct with 0.0% of visitors according to each stats I've checked.

harding:
@saivann it's probably worth creating a branch like bitcoin/minor-breaking-changes where we can put stuff like this until we have a major change, so that we don't forget the minor stuff and can merge it in together with the major stuff.

Anyway, based on @saivann's comment, I continue to plan to merge this PR as-is around 22:00 UTC today.

Issue #773, opened 2015-02-25
QA: Site Build Problem Follow-Up


harding:
This morning, #bitcoin IRC user waxwing reported that all of the tables in the RPC and REST section were broken:

![build-problem](https://cloud.githubusercontent.com/assets/61096/6373653/1292aa38-bcda-11e4-8841-f05c9b0e2da0.png)

Recalling a similar situation from about a year ago, I was able to quickly push an apparent fix in commit 84f5b6f160. This issue isn't about the bug itself but about preventing future problems.

I find this issue particularly concerning because I can't replicate it locally, and the same commit that produced the broken page also compiles cleanly on Travis.  I suspect that the build server is compiling the site differently from the way I compile it, and that's the source of the immediate problem, so my first recommendation is:

> Select one canonical way to compile the site and make sure that both Travis and the build server use it.  Also describe it in the README.

The Makefile has a test that would've caught this problem if the site was built through make rather than by calling `bundle exec jekyll build` directly.  One concern we've had with the Makefile is that it may implement tests that require programs/libraries not installed on the build server.  To deal with that, my next recomendation comes in two parts:

> 1. I push a commit adding a `site-build` target to the Makefile that builds the site and only runs tests that we know work on the build server. Travis and users can still use `make all` to run all tests.
>
> 2. We build the site using `make site-build`

Finally, I'd like to know if the live site differs from my local site, so I suggest we add to the Makefile a post-build step that runs a command like this:

    find _site -type f | xargs sha256sum > _site/sha256sums.txt

And then another Makefile target that can be manually run to diff the local _site/sha256sums.txt against the bitcoin.org/sha256sums.txt, showing whether any files built locally are different from the current site files.  (This is not a security step, just a QA step.)

Some files will likely always differ, such as the events page that includes data from Meetup or the two pages that include commit counts from GitHub---but I think most of the rest of the pages are reasonably deterministically built and should be identical between local and remote builds.  In short form, I recommend:

> I create a pull request adding the above-described commands to the Makefile.

Also, for discussion, I think the site building script should be part of the repository. Since anyone who has commit access can push a commit adding an rsh server to the _plugins directory, it seems like executing the build from a file in the repository doesn't introduce any new attack surface. However, I'm happy to discuss this further and also review the current build script for problems before making it public.

[...]

PR #776, opened 2015-02-26
QA: Simplify Comparison Between Local And Remote Site Files

harding:
This commit adds an extra step to the build process to SHA256 checksum all the files that are built purely based on repository contents.  The checksums file is uploaded to the webserver with the rest of the site content.

A separate target is added to the Makefile to compare the remote sha256sums file to a locally-built file to see if they differ.  This allows us to detect when a remote build may have gone astray.  (This is for QA only, not security.)

I will be adding this step to my usual merge workflow:

1. Merge pull request(s) at the command line.

2. `git diff HEAD^` to review the patchset

3. `git commit -a -S --amend` to sign the merge

4. `make all` to build the site locally and run the automated tests

5. (Optional) manually preview significantly changed pages using my localhost webserver and browser

6. `git push upstream master` to push the merged commits

7. Wait a few minutes for the revised site to build and be uploaded.  Then run `make manual-checks` to run this new check to see if the actual Bitcoin.org site matches what I built locally.

Some background info on my desire for this change can be found in issue #773


saivann:
@harding Half untested LGTM, you made it more simple than I thought!

harding:
@saivann thanks!

I'll merge this we the next other thing that I merge, and then confirm that it works once the site gets updated.

harding:
Just a quick follow-up that this commit does not work as expected because the site build does not use the same version of the less (CSS) compiler/compressor we have in our Gemfile, so the main CSS file is slightly different than what I compile locally.  For example:

    -h1{font-size:2em;margin:0.67em 0}
    +h1{font-size:2em;margin:.67em 0}

That wouldn't be a problem, except that we use the MD5 hash of the CSS file as its filename to ensure that browsers download the most recent CSS file---so a slightly different CSS filename means every HTML file on the site now has a different hash.

Once we get the site build script added to the repository, I'll work on making it use the same build process we describe in the README.md and that we use on Travis.

saivann:
@harding Ah.. Well actually I kept using the ubuntu packages for the build machine (are the gems used by bundle secure, properly signed and verified?). If you feel comfortable with it, please feel free to update and test the build machine configuration, and updated the update_site.sh script once it's merged.

PR #777, opened 2015-02-26
Links: Add two websites

louisjc:
I propose to add two websites
* [CryptoFR](http://www.cryptofr.com/):  A growing French forum, created by [some disappointed](https://bitcointalk.org/index.php?topic=682126.0) of bitcointalk.
* [GoBitcoin.io](http://gobitcoin.io/en/): A multilingual Bitcoin Converter. I hesitated few times to propose this one because I administer it, I'm therefore not the best to judge.


harding:
The cryptoFR forum does look active, and your price converter seems a reasonable enough addition.  I'd like to leave this pull open for a couple days in case anyone else wants to comment, but otherwise it LGTM.  Thanks!

PR #781, opened 2015-02-27
Add script used to build and update website

saivann:
(As suggested in #773)

These scripts are already in use in the build machines, which are now shared between two administrator (David and me), so it is becoming more important to make sure everyone keeps the same version.

Edit: These scripts wouldn't be auto-updated via the repository.


harding:
LGTM.

In the absence of critical feedback, this will be merged around 14:00 UTC Monday.

PR #783, opened 2015-03-01
Small change: make Getting started more dense

louisjc:
Just a very small change, I've reduced the margin on the Getting started page. This make the page a bit denser, useful for mobile and small screens.

### DIFF
--- a/_less/screen.less
+++ b/_less/screen.less
@@ -351,7 +351,7 @@ table td,table th{
 .menusimple li ul li.active a{
 	color:#fff;
 }
-	
+
 .content{
 	position:relative;
 	padding:30px 40px 40px 40px;
@@ -602,6 +602,9 @@ table td,table th{
 	padding-bottom:0;
 	position:absolute;
 	bottom:35px;
+}
+.starttitle a{
+	margin:0 20px auto;
 }
 
 .introlink{
@@ -1890,7 +1893,7 @@ h2 .rssicon{
 .press h2{
 	font-size:150%;
 }
-	
+
 .press-volunteer{
 	text-align:left;
 	margin-bottom:40px;


saivann:
@louisjc Can you apply this to the mobile layout only perhaps, to keep the desktop layout aerated? At the end of the CSS file you can file the rules applied to the mobile layout.

louisjc:
Don't you think this is an improvement for both mobile and desktops? My window size did make a perfect argument for it :

margin

saivann:
@louisjc Mmh, to be honest I think the existing margins are better on desktop, insofar as keeping things areated seems to be the usual watchword for modern designs. However my main issue is with the lack of margin between the "How to use Bitcoin" and "How to accept Bitcoin" subsections of the page.

louisjc:
Didn't saw you want to keep it aerated (I read unaltered), i think this is a good point you're right! I think I should close it then.

While I'm in the small layout changes, is there a reason why none of the images are center (I think about [this page](https://bitcoin.org/en/how-it-works) and the [faq](https://bitcoin.org/en/faq)), I was going to open a pull about it but I would like to hear your opinion before.

harding:
@louisjc I like the idea of centering images.  I've wanted that before for the dev docs.

I don't want to impose, but if you're looking for a project, another thing I'd like to see done is have all of the clickable links on the site use the same blue color (including the small boxes like Read More on the getting started page), and then make sure we don't use a similar blue color for anything else (like subheads).

I'm no UI expert, but I think having a single color for all links is recommended.

saivann:
@louisjc Personally I think your changes are an improvement for the mobile layout! In case you want to apply it only to mobiles :) . Thanks for your work on this. Centering pictures is generally fine with me!

louisjc:
@saivann Sorry for the delay, I've moved the change to the mobile section :)

harding:
@louisjc looks good to me, thanks!  I'll merge this in the next time I do a merge.

PR #786, opened 2015-03-05
Add btcwallet and update url for btcd.

dajohi:


harding:
Looks legit.  I'll merge in a few minutes.  Thanks, @dajohi!

PR #789, opened 2015-03-14
QA: Use HTML Proofer To Check Internal Links

harding:
Uses Ruby [html-proofer](https://github.com/gjtorikian/html-proofer) to check the links and some other attributes.  This commit also fixes the various problems it found, as well as dealing with some of its non-problem complaints (it doesn't like anchor (a) tags without either an href, name, or id).

Running HTML proofer takes about 12 minutes on my system (with up to two threads), during which it prints no text. Travis CI times out after 10 minutes of nothing being written to stdout, so this commit also adds a background process the Makefile to print a line every minute while make runs.

The production build server (VPS) will have to be updated before this can be merged because of the new dependencies, so I'll do that shortly before merging.


_templates/about-us.html
```diff
@@ -41,22 +41,22 @@
...
-  <p><a>Jacob Burenstam</a><span>Swedish</span></p>
-  <p><a>Péter Kemenczés</a><span>Hungarian</span></p>
-  <p><a>Matija Mazi</a><span>Slovenian</span></p>
-  <p><a>Mihai Onosie</a><span>Romanian</span></p>
-  <p><a>Boštjan Pirnar</a><span>Slovenian</span></p>
-  <p><a>Luigigiuseppe Prosperi</a><span>Italian</span></p>
-  <p><a>Thomas Pryds</a><span>Danish</span></p>
+  <p>Garland William Binns III<span>Maintenance</span></p>
+  <p>@arvicco<span>Russian</span></p>
+  <p>Simon Alexander Hinterreiter<span">German</span></p>
```
saivann: @harding: I just noticed an orphan double quote here.

saivann:
Untested LGTM, thanks! Maybe it would be good to preview the website just to be sure there isn't any broken HTML?

harding:
@saivann thanks!  I did preview the website locally; that orphan quote slipped by during a last-minute fix---thanks for catching it!

HTML proofer has a check option that validates HTML; I don't have it enabled in this pull but I'll probably be trying it out next week.

In the absence of critical feedback, I'll merge this pull Monday.

harding:
@saivann FYI: when I installed and ran the update to the build script from this pull, it didn't work.  I'm not sure why.  I wrestled with it for a while but decided it would be better off waiting until I have some time to better integrate the Makefile, bundle, and the build script.

I changed the script back on the server and pushed c2bc8edff3a5 reverting the change here in the repository and verified that the site is now building correctly again.

PR #800, opened 2015-04-04
Wallets: Revert "Drop Coinapult wallet"

harding:
This reverts commit 6fa5684924587fa010b0d2bfacd95c291df7a3f6.

Some relevant quotes, good and bad, from Coinapult:

> Despite further in depth analysis, we have not yet been able to determine the specific entry point of the attack. (Source: [Blog][])

> we replaced our hot wallet with a 2 of 3 multisig system which currently requires manual cosigning by a Coinapult executive for each withdrawal. (Source: [blog][])

> Our highest priority now is to shift from the interim multisig schema to one with client-side signing. (Source: [blog][])

> Coinapult will manually approve (co-sign) all withdrawals, so expect some delays.(Source: [maintenance notice][])

It isn't clear to me whether or not they've re-enabled instant bitcoin withdrawals, nor whether or not they're now fractional reserve.  However, we don't actually have requirements for either of those things, so if nobody objects, I will merge this pull request on Monday the 13th.

[blog]: https://medium.com/@Coinapult/full-service-restored-85409ddaa5ea
[maintenance notice]: https://coinapult.com/march

CC: @g-p-g


g-p-g:
Thanks @harding. Answering your two implicit questions:

1) You can initiate a withdrawal at any point, and if that happens successfully it means the current multisig wallet has enough funds to cover for it. We don't keep all the funds in a single multisig wallet, it's refilled as necessary. After the withdrawal is initiated, it has to be cosigned manually -- this takes time, specially during weekends and outside business hours, we're checking what we can do to reduce this time.

2) Right to the point: Coinapult can cover all current obligations, there's no fractional reserve.

saivann:
LGTM

g-p-g:
Thanks all, sorry for the inconvenience.

PR #807, opened 2015-04-07
Missing address for events causes build script to fail

buren:
[_events.yml:L76-79](https://github.com/bitcoin/bitcoin.org/blob/master/_events.yml#L73-79) 
```YAML
- date: 2015-05-02
  title: "WeBTCExpo"
  venue:
  address:
  city: "Amsterdam"
  country: "Netherlands"
  link: "http://webtcexpo.com/"
```
does not define any address and raises an error on this line: [events.rb#L106](https://github.com/bitcoin/bitcoin.org/blob/master/_plugins/events.rb#L106)

```ruby
# In the middle of line 106:
data['address']  # => nil
# And therefore
data['address'] + ',' # => undefined method `+' for nil:NilClass
```

I think the simplest solution would to be to make the address optional and let google place a generic pin based of `data['city']`, `data['country']` therefore this PR.

What do you think? :eyes:

### DIFF
--- a/_plugins/events.rb
+++ b/_plugins/events.rb
@@ -103,7 +103,8 @@ module Jekyll
         # Get geolocalisation data from Google Maps
         if data.has_key?('address')
           begin
-            geoloc = JSON.parse(open("https://maps.googleapis.com/maps/api/geocode/json?address=" + CGI::escape(data['address'] + ', ' + data['city'] + ', ' + data['country']) + "&sensor=false","User-Agent"=>"Ruby/#{RUBY_VERSION}").read)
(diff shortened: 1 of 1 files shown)


saivann:
@buren A side effect of publishing events without address is that the pin on the map becomes inaccurate and, although this might need verification, events without clear venue seem to be cancelled more often. The whole idea of publishing an event with incomplete data seemed a bit odd to me.

Maybe I'm wrong though and perhaps there is more good sides of being more permissive. I do think however that it might be more appropriate to not add a pin on the map when it's not going to point to the right location.

buren:
I can definitely agree :) Closing.

PR #809, opened 2015-04-07
Fixes the position of the nav icon when resizeing the window on Chrome

### LINKED ISSUE #808: Mobile menu icon displays below the navbar when resizing the window on chrome
I'm seeing a rather funny bug on Chrome, OSX 10.10 & Windows 7, which causes the mobile menubar to be displayed under the navbar, see the screenshot, it only happens when resizing the window and you can "fix" it be unsetting and setting `float:right` to `ul#menumobile`..

It renders fine if the page is small enough for the mobile nav to be shown from the getgo.

bitcoin-mobile-navbar-chrome

buren:
This fixes #808 for some reason, must be a bug in chrome.. Tested the change on safari and firefox and then still render it fine. I'm a bit frustrated that I have no clue why this fixes the issue in chrome :)

@saivann

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -40,8 +40,8 @@
       </ul>
       </li>
     </ul>
-    <a class="logo" href="/{{ page.lang }}/"><img src="/img/icons/logotop.svg" alt="Bitcoin"></a>
     <a id="menumobile" class="menumobile" onclick="mobileMenuShow(event);" href="#"></a>
+    <a class="logo" href="/{{ page.lang }}/"><img src="/img/icons/logotop.svg" alt="Bitcoin"></a>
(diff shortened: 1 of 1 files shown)


saivann:
Strange indeed. LGTM, thanks!

harding:
@saivann that change to the base template will trigger the download-page-changed alert, right?  So I should combine it with the 0.10.1 update?

saivann:
@harding The check is not this strict, so AFAIK you can merge it whenever you want. The alerts will be triggered by changes in more sensitive elements only, such as javascript code, links, specific filetypes, etc.

harding:
@saivann oh, right.  I'll merge it in a couple hours with the rest of my updates.  Thanks!

PR #810, opened 2015-04-07
Stylefix on dev docs page

buren:
https://bitcoin.org/en/developer-documentation the links with icons looks broken on tablets/large phones.

### Before

brokenstyle

### After

displayinline

### DIFF
--- a/_less/screen.less
+++ b/_less/screen.less
@@ -1098,6 +1098,7 @@ table td,table th{
 .resources .resourcesext{
 	margin-top:5px;
 	min-height:0;
+	display: inline-block;
 }
 .resources .resourcesext p{
 	padding-left:24px;


buren:
I'm sorry if I come across as spammy. Do you prefer each separate change, however small, to be in its own PR or should I group of these kind of smaller changes together?

harding:
@buren I quickly tested this change and it looked fine to me.  I'll test it a bit more throughly and, if there are no problems, merge it.

Regarding grouping changes, what you've been doing is perfect so far.  The official policy is to try to make it as easy as possible to review the changes, so related changes should be grouped together and unrelated changes should be separated apart irrespective of the actual size of the change.

harding:
@buren oh, and thanks for all of your work!

PR #813, opened 2015-04-09
Layout: Add anchor links to all subheads

harding:
This pull adds anchors to all subheadings (h2 through h6) on the site, except for the subheadings on the Choose Your Wallet page (which are almost all hidden by default).

Also added is a test to the Makefile that errors out if any subhead doesn't have the `id` attribute necessary to create an anchor.

This has no effect on the appearance of the site, but when combined with a pull request #814, it add an iconified link to the a subhead's anchor when a Javascript-enabled user hovers over the subhead.

Live preview of this PR plus PR #814 merged together: http://dg1.dtrt.org/en/bitcoin-for-individuals

Screenshots of the anchor link icon in different layouts:

2015-04-09-112414_1280x800_scrot

2015-04-09-112425_1280x800_scrot


_templates/download.html
```diff
@@ -21,7 +21,9 @@
 <link rel="alternate" type="application/rss+xml" href="/en/rss/releases.rss" title="Bitcoin Core releases">
 <div class="download">
   <h1>{% translate pagetitle %}</h1>
-  <h2>{% translate latestversion %} {{ site.DOWNLOAD_VERSION }}<a type="application/rss+xml" href="/en/rss/releases.rss"><img src="/img/icons/icon_rss.svg" alt="rss" class="rssicon"></a></h2>
+  <div class="inline-block">
+    <h2 id="latestversion">{% translate latestversion %} {{ site.DOWNLOAD_VERSION }}<a type="application/rss+xml" href="/en/rss/releases.rss"><img src="/img/icons/icon_rss.svg" alt="rss" class="rssicon"></a></h2>
+  </div>
```
saivann: @harding This workaround kind of breaks down the layout for mobile users (see screenshot below), probably because the download button isn't shown. May I suggest you convert this title into a stylized DIV?

capture du 2015-04-09 15 30 50

harding: @saivann thanks for catching that!  I really do have to remember to do more mobile testing.  I fixed it by removing the commit that added the workaround, removing the header IDs so no anchor links are displayed, and exempting the Download page from the anchor link checking (it's not like anyone is likely to link to the anchors on this page anyway).  I checked that in Firefox's responsive test mode and it's no longer broken.

harding:
In the absence of critical feedback, I will merge this pull around 23:00 UTC Saturday.

PR #814, opened 2015-04-09
Layout: Add anchor link affordance to all pages

harding:
This pull calls a Javascript function to display a hyperlinked icon whenever a Javascript-enabled user hovers on any subhead on the site that has an anchor set.  Combined with PR #813, it makes it easy for users to link to any section within the site.

This pull request is being added to the bitcoin-core-0.10.1 branch because updating the Javascript will trigger an alert to people monitoring the page.  Since that alert will be triggered when 0.10.1 is released, this will avoid an unnecessary extra alert.

Live preview of this PR plus PR #813 merged together: http://dg1.dtrt.org/en/bitcoin-for-individuals

Screenshots of the anchor link icon in different layouts:

2015-04-09-112414_1280x800_scrot

2015-04-09-112425_1280x800_scrot

Note: this branch will fail Travis CI because its based on the 0.10.1 branch that itself fails until 0.10.1 final is uploaded to the site.

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -106,7 +106,10 @@
     <div class="footerlicense">© Bitcoin Project 2009-{{ site.time | date: '%Y' }} {% translate footer layout %}</div>
     </div>
   </div>
-  <script type="text/javascript">fallbackSVG();</script>
+  <script type="text/javascript">
+    fallbackSVG();
+    addAnchorLinks();
+  </script>
 </body>
 
 </html>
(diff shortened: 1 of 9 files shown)


saivann:
@harding Looks good! One bug: the "addAnchorLinks" function should be moved to base.js otherwise it will be called on the download page, and fail since the function won't exist on that page. I don't think this function calls other functions that are missing in base.js, but it's probably worth checking in your javascript console to confirm there isn't any error.

harding:
@saivann good catch.  Thanks!  In this branch (pull #814), I moved the JS code as you suggested.  In the other branch (pull #813) I fixed a stylesheet issue that made the link icon appear all the way on the side of the page even though the header text is centered here.  For that, I just copied @buren's method from a couple days ago.

I've updated the combined preview and verified there are no JS errors: http://dg1.dtrt.org/en/download

PR #815, opened 2015-04-09
Expend the "Participate" menu entry

saivann:
Live preview: http://bitcointest1.us.to/

I remember reading comments saying the "Development" page was buried and hard to find, and the current guide to run a full node isn't visible in the menus either. I feel like these two would find their right audience and get more visibility by becoming a submenu of the "Participate" menu entry.

capture du 2015-04-09 17 38 46


harding:
@saivann LGTM. Thanks!

saivann:
In the absence of critical feedback, this pull request will be merged on April 11th.

(I will take care to also apply the changes upstream on Transifex)

PR #816, opened 2015-04-10
Reword description of transaction execution

gwillen:
I believe that the current description of transaction execution is not accurately describing how the stack machine works. (It's close but uses slightly misleading wording.) While data values are pushed when executed in the script, other opcodes are never pushed; they are simply executed.

### DIFF
--- a/_includes/guide_transactions.md
+++ b/_includes/guide_transactions.md
@@ -224,12 +224,12 @@ of the process.
   pubkey hash at the top of the stack.
 
 * Now it gets interesting: Alice's pubkey script executes `OP_EQUALVERIFY`.
-  `OP_EQUALVERIFY` is like executing `OP_EQUAL` followed by `OP_VERIFY` (not shown).
+  `OP_EQUALVERIFY` is equivalent to executing `OP_EQUAL` followed by `OP_VERIFY` (not shown).
 
-    `OP_EQUAL` (not shown) checks the two values below it; in this
+    `OP_EQUAL` (not shown) checks the two values at the top of the stack; in this
     case, it checks whether the pubkey hash generated from the full
     public key Bob provided equals the pubkey hash Alice provided when
-    she created transaction #1. `OP_EQUAL` pops (removes from the stack)
+    she created transaction #1. `OP_EQUAL` pops (removes from the top of the stack)
     the two values it compared, and replaces them with the result of that comparison:
     zero (*false*) or one (*true*).
 
--- a/_includes/guide_transactions.md
+++ b/_includes/guide_transactions.md
@@ -199,8 +199,8 @@ sections about stacks. These are programming terms. Also "above",
 "below", "top", and "bottom" are commonly used relative directions or
 locations in stack descriptions. -harding -->
 
-To test whether the transaction is valid, signature script and pubkey script arguments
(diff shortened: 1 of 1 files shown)


_includes/guide_transactions.md
```diff
@@ -211,38 +211,37 @@ of the process.
...
   This creates a hash of Bob's public key.
 
 * Alice's pubkey script then pushes the pubkey hash that Bob gave her for the
   first transaction.  At this point, there should be two copies of Bob's
   pubkey hash at the top of the stack.
 
-* Now it gets interesting: Alice's pubkey script adds `OP_EQUALVERIFY` to the
-  stack. `OP_EQUALVERIFY` expands to `OP_EQUAL` and `OP_VERIFY` (not shown).
+* Now it gets interesting: Alice's pubkey script executes `OP_EQUALVERIFY`.
+  `OP_EQUALVERIFY` is like executing `OP_EQUAL` followed by `OP_VERIFY` (not shown).
```
apoelstra: Can you change "is like" to "is equivalent to"?

_includes/guide_transactions.md
```diff
@@ -211,38 +211,37 @@ of the process.
...
 
     `OP_EQUAL` (not shown) checks the two values below it; in this
```
apoelstra: "two values below it" doesn't make sense now that we're not talk as though opcodes are on the stack; this should be changed to "two values at the top of the stack".

_includes/guide_transactions.md
```diff
@@ -211,38 +211,37 @@ of the process.
...
     case, it checks whether the pubkey hash generated from the full
     public key Bob provided equals the pubkey hash Alice provided when
-    she created transaction #1. `OP_EQUAL` then replaces itself and
-    the two values it compared with the result of that comparison:
+    she created transaction #1. `OP_EQUAL` pops (removes from the stack)
+    the two values it compared, and replaces them with the result of that comparison:
```
apoelstra: "removes from the stack" → "removes from the top of the stack"

apoelstra:
Other than the minor comments I added inline, I strongly encourage accepting this change; it gives a much clearer picture of the stack machine operation.

gwillen:
Thanks @apoelstra, made the suggested changes.

This probably wants to be squashed with rebase into a single commit (I made the edits through the github web editor which can't automatically do that.) I can pull to a real git client and do that myself if requested.

apoelstra:
Thanks @gwillen, with these changes I support this PR.

maaku:
ACK. This is much clearer (and correct). Thanks for catching this @gwillen .

harding:
This LGTM.  I'll squash it down to one commit and merge it as soon as I fix the unrelated problem that's causing the CI to fail.  Thanks for doing this @gwillen, and @apetersson and @maaku for comments!

@gwillen out of curiosity, did you use the Edit links from on the documentation pages?  I added those a few months ago and have been surprised at how rarely they're used.

gwillen:
@harding Yes, I did use the Edit link! Other than that I wouldn't have realized it was editable. And the git web-edit workflow is surprisingly good (I've never used it before.) It really reduces the friction for stuff like this.

Thanks!

saivann:
@gwillen Nice to hear, thanks for your help!

PR #818, opened 2015-04-10
Skip htmlproofer when partially building the website

saivann:
This little change allows me to benefit from other Makefile tests by getting meaningful build failures when doing partial builds. I guess HTMLProofer cannot do much in this specific case, as having missing pages and broken links is an expected outcome of doing a partial build.

(Note: If we later choose to go as far as to validate HTML, we could easily tell HTMLProofer to only do HTML validation and skip link checks)

### DIFF
--- a/_contrib/bco-htmlproof
+++ b/_contrib/bco-htmlproof
@@ -2,6 +2,12 @@
 
 require 'html/proofer'
 
+## Don't check links when partially building the website.
+if !ENV['ENABLED_PLUGINS'].nil? or !ENV['ENABLED_LANGS'].nil?
+  print 'HTMLProofer skipped (partial build)' + "\n"
+  exit
+end
+
 ## Will throw an exception (exiting false) if any internal links don't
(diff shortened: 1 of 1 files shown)


harding:
@saivann I don't have any problem with this approach, but I've been doing my quick builds using `make valid` which runs all of the fast tests so you get instant feedback, but doesn't run any of the slow tests like HTML proofer and the test to download the HTTP headers for the Bitcoin Core binaries.  This is documented [here](https://github.com/bitcoin/bitcoin.org#building-with-make).

I do think `make valid` might need some updating after the Glossary PR gets merged, as it reports some missing links if the glossary plugin is disabled, but I'll fix that when I merge `_autocrossref.yml` and `_includes/references.md` into `_config.yml`.  (In the meantime, we can just use `make -i valid` to prevent make from stopping after an error.)

saivann:
@harding Ah, right. I should have looked at the Makefile more carefully. Thanks!

PR #820, opened 2015-04-11
Fix endless loop bug in build scripts

saivann:
The subshell used in the script to prevent it from building if another concurrent build starts had the unintended consequence of creating an endless loop when the build script fails.

This pull request allows the script to terminate correctly.

Additionally this pull request enables the quick Makefile tests to catch issues in order to prevent the server to push any broken update (as we've already planned to do, it currently works fine on the build machine according to my testing).


harding:
LGTM.  Thanks, @saivann!

PR #821, opened 2015-04-11
Standardize translation links

sandakersmann:
Changed all translation links so they start with capital letters


saivann:
That indeed seems more consistent. I just found Facebook seems to be doing the same, so I guess it's acceptable. For some reason, Google uses lowercase letters for some languages.

@sandakersmann Thanks for your help!

LGTM (commit d0ab1a2) once Travis CI passes.

PR #823, opened 2015-04-12
Various Backend & Site Docs Updates

harding:
Separated into multiple commits for easy review, this pull includes:

1. Require Jekyll 1.3.0 through Bundle. No more alternative versions allowed if using the Makefile. Require Ruby 2.0.0. Remove a plugin that's no longer necessary in Jekyll >=1.0.0. Remove plugin workarounds that are no longer required on Jekyll 1.3.0.  Verified that layout is correct after workaround removal.

2. Add Makefile targets for `make deployment` (for build server) and `make travis` (for Travis CI).  Deployment uses only libs from Bundle (i.e. `bundle --deployment` and also updates libs automatically when possible); otherwise it's identical to `make valid`.  Travis is `make all` with an extra keep alive previously part of `make all`.

3. Add a new plugin for `jekyll serve` to make `/foo` serve the file at `/foo.html` (reducing the amount of instructions we need). Add Makefile target for `make preview` to launch `bundle exec jekyll serve`. Major update of site docs to describe installing; I spent several hours repeatedly starting from a bare install of Ubuntu Server and then testing these instructions until a successful build/preview of the site, so hopefully they're bulletproof.

Before merging this commit, I will install RVM and Ruby 2.0.0 on the site build server and translation build server. (I'll also monitor the site build server and actual site uploads for problems immediately after commit.  In order to ensure I have time to fix any problems, I'll only consider merging before noon my localtime.)

After this is merged and the site successfully updated, I'll re-merge the developer glossary and also rebase the Bitcoin Core 0.10.1 branch (which should be fine, but why take chances?)


harding:
Oh, I just remembered to diff the `_site` directory between before and after this change and everything is identical except for `/README.md` (as expected) and `/sitemap.xml` (non-deterministic, I guess; everything LGTM in both versions of the file).

README.md
```diff
@@ -210,13 +251,45 @@ commands:
...
+| Plugin       | Seconds | Remote APIs    | Used For
+|--------------|---------|----------------|------------------------
+| alerts       | 5       | --             | Network alert pages
+| autocrossref | 90      | --             | Developer documentation
+| contributors | 5       | GitHub.com     | Contributor listings
+| events       | 5       | Meetup.com; Google Maps | Events page
+| glossary     | 30      | --             | Developer glossary
+| redirects    | 20      | --             | Redirects from old URLs
+| releases     | 10      | --             | Bitcoin Core release notes; Download page
+| sitemap      | 0       | --             | /sitemap.xml
```
saivann: @harding FWIW, the run time for this plugin is influenced by the number of languages enabled. With no language disabled, I can see a 10 seconds difference, in case you'd like to update the number.

README.md
```diff
@@ -134,73 +134,114 @@ You simply need to push additional commits on the appropriate branch of your Git
...
+And tell your system to use it:
 
-    bundle exec jekyll build
+    rvm use default
 
-You can then copy the output files from _site/ to the root of your web server.
-If you have no web server, run `bundle exec jekyll serve` and visit
-http://127.0.0.1:4000/. This server requires you to add a trailing ".html"
-by hand in your browser address bar.
+(Note: you can use a different default Ruby, but you if you ever change
```
saivann: s/but you if you ever/but if you ever/

Makefile
```diff
@@ -28,16 +25,35 @@ test: pre-build-tests post-build-tests
...
 
-## Pre-build tests which, aggregated together, take less than 5 seconds to run on a typical PC
+## Install dependencies (development version)
+install-deps-development:
+	bundle install
+
+## Install dependencies (deployment version)
+install-deps-deployment:
+	bundle install --deployment --without :slow_test
+
```
saivann: @harding Is it expected that we have two "install-deps-deployment" Makefile rules?

Gemfile
```diff
@@ -1,15 +1,32 @@
...
+## README.md. Then push your branch and make sure Travis supports that
+## version. Then remind one of the site maintainers that they need to
+## run `rvm install <VERSION>` on the build server(s) before they commit
+## to master
+ruby '2.0.0'
 
+## Used on the build server. If you add a package here (like nokogiri)
+## that has non-Gem dependencies (like zlib), please remind the site
+## maintainers that they need to manually update the build server(s)
+## before they commit to master. If `bundle install` can satisify all
```
saivann: s/satisify/satisfy/

saivann:
This is fantastic, having "jekyll serve" working with clean urls has been a repeated wish with no solution. Kudos for finding a workaround! This will make the build process much more approcheable for new contributors. Everything else also LGTM (except the few small comments above)!

Makefile
harding: @saivann one is "-**deploy**ment" (used by the build server) and the other is "-**develop**ment" (used, somewhat unnecessarily, by Travis CI).  Travis actually checks out the deps on its own, but I wanted a rule that replicated what it does for local testing.

harding:
@saivann thanks!  Addressed what I think is all the applicable feedback.

In the absence of critical feedback, I'll begin updating the build at translation servers around 11:00 UTC tomorrow; then I'll merge this and re-enable the glossary.  When I've done that and confirmed it all works, I'll rebase the 0.10.1 branch.

Makefile
saivann: Oh x_X, right

PR #829, opened 2015-04-20
Link "offline wallet" in FAQ to Armory wallet website

sunnankar:
No text changed to prevent translation issues. Added URL to Armory since they innovated offline wallets.


harding:
@sunnankar thanks again for another pull request. :-)  I don't think the Armory website has much to do with offline wallets, and we have a description of offline wallets in this site's developer documentation.

harding:
Closing: I disagree with implementing this and there has been no response from the PR author in more than a week.  We can re-open at any time if further discussion is desired.

PR #831, opened 2015-04-20
Link "cryptographic algorithms" in FAQ to Block Hashing Algorithm Wiki page

sunnankar:
No text changed. Link to article on Block hashing algorithm for reference.


harding:
@sunnankar thanks for this pull request.  I don't think we need to add a link here, as the site already has more than 200 pages of developer documentation (including an extended section about how the block chain works).

saivann:
Agreed with @harding, changing links in translations is time-consuming only for convenience links.

harding:
Closing: two votes against implementing this and no response from PR author in more than a week.  We can re-open at any time if further discussion is desired.

PR #843, opened 2015-04-27
Add date and magnet link for 0.10.1

saivann:


harding:
@saivann I actually did this on my local computer and am running the tests locally, and I'll push as soon as they pass.  Sorry, I should've said something!

saivann:
@harding Thanks, np

PR #849, opened 2015-05-01
Separate event script from event data

saivann:
When monitoring sensitive changes on the website, regular script changes on all event pages are generating unhelpful non-descriptive noise, which might grab my attention away from real suspicious script changes.

This change renders events links and data directly on the page (inside an hidden DIV) and lets the script grab required data from there instead.

### DIFF
--- a/_less/screen.less
+++ b/_less/screen.less
@@ -1942,6 +1942,9 @@ h2 .rssicon{
 	background-image:none;
 	color:#fff;
 }
+.eventdata{
+	display:none;
+}
 
 .detectmobile{
 	width:0;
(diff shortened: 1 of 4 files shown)


harding:
@saivann do you need me to do something similar with https://github.com/bitcoin/bitcoin.org/blob/master/js/devsearch.js ?

saivann:
@harding Thanks for asking. I think it's always a good practice to keep data outside of scripts, but the devsearch.js script is way less of an issue compared to the event page. The later receives automatic unimportant modifications many times a week.

harding:
Tested LGTM.  @saivann extra thanks for adding the "this file is built non-deterministically" back in! :-)

saivann:
@harding Thanks for the careful testing and deterministic monitoring on the website :)

In the absence of critical feedback, this pull request will be merged on May 2th.

PR #861, opened 2015-05-23
content (en/fr): unconfirmed txs aren't secure

arthurbouquet:
Hi, 
On  https://bitcoin.org/en/you-need-to-know - "Instant transactions are less secure"
>  ...asking for a small transaction fee or using a detection system for unsafe transactions can increase security...  

As for as I know, there isn't any such thing as "instant transactions": only unconfirmed txs which aren't secure! I suggest some changes in this pull request:
* en.yml: replace "instant" & "instanttxt" labels with "unconfirmed" & "unconfirmedtxt", update content to reflect my point.
* fr.yml: same than en.yml (I'm a native French speaker, btw sorry for my english :-).
* _translation/*.yml: labels updated, content not updated.
* _templates/you-need-to-know.html: update labels in layout.

PS:  I don't have ruby/all the stuffs to build and preview the changes localy, I can't be sure if there won't be any problem with the modification in the html template (h2's id) event if it's "seems ok" (CI will tell :-).


harding:
I think this phrasing change is good, and this pull LGTM as long as Travis CI doesn't catch any problems with the h2 id changing.  Thanks, @arthurbouquet!

@saivann could you please review the French change?

saivann:
@arthurbouquet Thanks, I'm fine with the English and French rephrasing. "Unconfirmed" in indeed more common lingo. However can we avoid changing the label? It doesn't really matter what ID we use here, changing it I think only causes unnecessary work when managing translations and translators (for instance this has the side effect of deleting translation history and comments on Transifex).

arthurbouquet:
@saivann It's up to you guys, I changed the label because it was "wrong" from me / I don't know what is a instant tx in bitcoin land :) 
What shoud I do? Should I close this one and create another one with with suggested changes ? (sorry it's my first pull request on github...)

harding:
@arthurbouquet you can just reverse the `sed -i` command I gave you yesterday, run `git diff` to review the changes, `git commit --amend -a` to amend your previous commit, and then `git push -f` (the -f is for "force") to push the changes to your GitHub branch.

Travis CI will re-check the changes, and if everything looks good, I'll merge.

By the way, sorry for giving you bad advice.  I thought changing the translation anchor made sense too, although I see @saivann's point about it creating needless work on the translation side.

arthurbouquet:
No worries, commit has been updated to revert labels changes and build is in progrss! (thx for the instructions).

harding:
Merged.  @arthurbouquet the change should appear on the live site within about 15 minutes.  Thanks again for your work!

saivann:
@harding Uh sorry, I didn't know that you suggested updating the label. It's not a big deal either way for a single translation. I'd be more discouraged if we changed every label name that is more or less relevant due to text changes or just lack of attention :) .

@arthurbouquet Thanks for making the change!

PR #873, opened 2015-06-03
Dev Docs: Remove Outdated Info From Payment Processing Section

harding:
A small change that removes some information that has become outdated or unnecessary since I first wrote this section 14 months ago.


saivann:
@harding LGTM

harding:
@saivann thanks!

In the absence of critical feedback, this will be merged Tuesday.

Issue #895, opened 2015-06-15
Bitcoin Conference Kiev 2015


voffkas:
- date: 2015-09-15
      title: "Bitcoin Conference Kiev 2015"
      venue: "NSC Olimpiyskiy"
      address: "Velyka Vasylkivska str., 55"
      city: "Kyiv"
      country: "Ukraine"
      link: "http://bitcoinconf.com.ua/en"

voffkas:
Can you fix date plesase?

- date: 2015-09-24

justin130:
24 June 2015

Sent from my iPhone

On 18 Jun 2015, at 13:00, Voffkas <notifications@github.com<mailto:notifications@github.com>> wrote:


Can you fix date plesase?

  *   date: 2015-09-24

-

harding:
Sure, but note that your [website](http://bitcoinconf.com.ua/en) still says 24 Sept 2015.

voffkas:
24 Sept 2015 - correct date

harding:
@voffkas oh, ok.  I'll fix it again, but it'll take a few minutes.

harding:
@voffkas the correct date, 24 Sept 2015, should be displayed on the site in about 15 minutes.  Thanks for letting us know.

voffkas:
Thank you!

Issue #885, opened 2015-06-12
Money20/20 Europe


VeronicM:
The World's Largest Fintech Event Is Coming To Europe

Money20/20 Europe is the Eurocentric catalyst for the growth and development of the Payments & Financial Services Ecosystem.

With 2,000+ attendees, including 250 CEOs, from over 1,000 companies and 50+ countries, Money20/20’s Copenhagen launch is critical to realizing the vision of disruptive ways in which consumers and businesses manage, spend and borrow money in Europe. This essential, must-attend industry gathering will take place from April 4-7 2016 at the Bella Centre, Copenhagen.

Money20/20 organises the world’s largest events focused on payments and financial services innovation for connected commerce at the intersection of mobile, retail, marketing services, data and technology. In April 2016, Copenhagen will take centre stage as the companies and individuals building, disrupting and challenging the ways consumers and business manage, spend and borrow money come together at scale.

PR #898, opened 2015-06-16
Add events from repo and coindesk

coderwill:
Reviewed and resolved a few open issues (#885, #886, #895) where events were pending to be added. #885 was a duplicate, fwiw.

Added latest events that had complete info that were listed on CoinDesk after reviewing them in detail.


harding:
@coderwill thanks!  Merged as b611432

PR #901, opened 2015-06-16
Update choose-your-wallet.html

schildbach:
A few weeks ago, I updated the BlackBerry version of the app to the same codebase as the Android version. So it now supports HD wallets and tries its best to not reuse addresses.


harding:
I don't think we have any reviewers with a BlackBerry to test this, but hearing that it's the same codebase as BW4A is good enough for me.

If anyone does have a BlackBerry and wants to test this, please leave a comment.  Otherwise I'll merge on Friday.

Thanks, @schildbach!

saivann:
LGTM, thanks!

crwatkins:
I have no BlackBerry, but I'm good with @harding's proposal.

Issue #896, opened 2015-06-16
Mycelium is not an open source wallet.


GandalfBitcoin:
The following description on bitcoin.org is not true: "Mycelium Bitcoin Wallet is an open source wallet".

Actually, the Microsoft Reference Source License ( [MS-RSL](https://en.wikipedia.org/wiki/Shared_source) ) used by Mycelium is the most restrictive of the Microsoft Shared Source licenses. We should not consider MS-RSL as an [open souce](http://www.quora.com/Why-is-MS-RSL-Microsoft-reference-source-license-not-considered-an-open-source-license) license.

So the description open source is totally wrong, and we should correct this. Otherwise it may cause misunderstanding.

saivann:
Agreed that "open source", while technically true, commonly also refer to a free license and is misleading (that's why one associated score was previously rewritten to "developers of this wallet publish the source code"). No one apparently catched this one. How about we just drop this part and say instead "Mycelium is a wallet for Android designed for security, speed, and ease of use."

ping @apetersson

saivann:
@GandalfBitcoin I'd appreciate you avoid using bold or uppercase in pull request (equivalent to screaming) and avoid an aggressive tone as much as possible. There is no Wikipedia-like code of conduct on this repository, but if there would be one, I guess this would be a fairly common guideline :) . We all try to be collaborative, thanks in advance.

GandalfBitcoin:
@saivann ok. sorry for the bold and uppercase.

theymos:
Maybe replace "open source" with "source-available" where appropriate.

saivann:
@theymos That would become "Mycelium Bitcoin Wallet is a source-available
wallet".

Are you sure it's understandable? If so, I'm also fine with it.

@GandalfBitcoin Thanks!

theymos:
@saivann That sentence is entirely understandable, though it's kind of clunky as an opening sentence. (I didn't previously look into where that sentence appeared.) Your suggestion is probably better.

schildbach:
You could also write "…is open source, but not free software" to make it clearer.

jandreske:
I saw "The source code is available for inspection" somewhere, which also includes hints at why this question is important to many.

GandalfBitcoin:
MS-PL and MS-RL are considered as open source by the Open Source Initiative and free by the Free Software Foundation.

[...]

PR #902, opened 2015-06-16
Update Mycelium description

saivann:
As suggested in #896, this pull request updates Mycelium's description, and drops any mention of it being open-source.

In the absence of critical feedback, this pull request will be merged on June 17th.


harding:
@saivann thanks!  Except for the comment to Craig below, this LGTM.

@crwatkins I know you had some concerns with another wallet that listed compatibility with hardware wallets that weren't really supported.  Do you think this needs investigation before being updated?

crwatkins:
@harding I can personally confirm (from previous testing) that Mycelium on Android is compatible with Trezor.  I can further confirm that MultiBit HD with the same Trezor can access the same account as Mycelium with that Trezor.  That's compatible in my book.  However, I don't believe that I would consider Mycelium compatible with MultiBit HD (without a Trezor).   I believe that (without a Trezor) Mycelium uses a BIP44 path of m/44'/0'/0'/c/i while MultiBit HD uses a BIP32 path of m/0'/c/i.  

My suggested fix with the least number of words changed would be "... offers compatibilty with Trezor, Multibit HD with Trezor, and others."  However, that's still not as clear as it could be.  Perhaps @apetersson can confirm this and suggest something that is clearer.

harding:
@crwatkins wow, thanks!  Let's see what @apetersson says.

jim618:
@crwatkins Primarily to cater for the use case of someone losing their Trezor and needing to recover their bitcoins in MultiBit HD with no replacement Trezor we programmed up that the user can create what we call a 'soft' wallet that uses a path of m/44'/0'/0'/c/i.

This is a wallet stored on the desktop computer with all the private keys stored locally and the keys use the Trezor path derivation of m/44'/0'/0'/c/i 

It is available in the restore as an option - see screenshot:

screen shot 2015-06-17 at 09 07 22

Thus more technically minded users _could_ restore their Mycelium seed phrase and create the same wallet on MultiBit HD.

We've not created a Wizard for this as we find users (and ourselves !) get confused by the different sorts of HD wallets. We don't want people thinking that a 'Trezor soft wallet' gives the same level of protection as an actual Trezor.

To make it as clear as we can to users we have a section in the Manage Wallet | Wallet dashboard titled 'Wallet capabilities'. This gives a description of the wallet e.g. for standard MultiBit HD wallets it says:

```
Standard MultiBit HD wallet
Private keys are encrypted on this machine
BIP32 compliant. m/0h (Account 1 only)
```

tl;dr A Mycelium wallet can be recreated on MultiBit HD if the user really wants to.

crwatkins:
Thanks Jim.  Just to be certain, however:  A wallet created on MultiBit HD could not be recovered with Mycelium, right?

gurnec:
@crwatkins Apologies for interjecting, but to answer your question: MulitBit HD always *creates* BIP-32 path wallets, and Mycelium only supports BIP-44 path wallets, so "yes, right".

I should also point out that Mycelium supports multiple BIP-44 accounts (I haven't tested this with a Trezor though), and a BIP-44 wallet imported into MultiBit HD will only see the first account with account index 0, called "Account 1" by Mycelium. Any other accounts are ignored.

If it's of any future use, I try to maintain a [spreadsheet here](https://onedrive.live.com/redir?resid=584F122BA17116EE%21313) listing deterministic wallet compatibility details, along with compatibility charts between different wallets (and a discussion thread is [here on BitcoinTalk](https://bitcointalk.org/index.php?topic=1000544.0)).

saivann:
If I understand correctly, MultiBit HD is compatible with Mycelium, but not the other way around. Accordingly, I have dropped MultiBit HD from the description.

@crwatkins @jim618 @gurnec Thanks a lot for the comments.

@gurnec Cool research! Would be great I think if it was published somewhere like the wiki.

harding:
Updated text LGTM.  Thanks everyone!

saivann:
In the absence of critical feedback, this pull request will be merged on June 18th.

crwatkins:
Thanks @gurnec 

@saivann LGTM

PR #908, opened 2015-06-22
Mailing list move

laanwj:
Move the mailing list from sourceforge to Linux Foundation.

Note: this does not update the text of all translations, which in many cases still refers to sourceforge, just the link and link text.


btcdrak:
ACK

saivann:
I have added a commit to stop mentioning sourceforge in the English texts (so translations can later be updated).

@laanwj Do you know what we should do with bitcoin-security@lists.sourceforge.net?

harding:
@saivann bitcoin-security@SF is staying there for the time being.

harding:
Reviewed diff and the previews in English and Spanish, and also looked for any other relevant strings.  LGTM.  I'm going to merge now, thanks!

syaifudin857:
Join

PR #914, opened 2015-06-23
Revise Mission Statement

coderwill:
Hello, as a result of the recent [Hard Fork Policy](https://bitcoin.org/en/posts/hard-fork-policy), the ***Mission*** should be revised on the [About Page](https://bitcoin.org/en/about-us) to remove the following statement because it's no longer true:

>Remain a neutral informative resource about Bitcoin.

Here are a few definitions of the word *neutral*, from [*The Free Dictionary*](http://www.thefreedictionary.com/Neutral):
> adj. Not aligned with, supporting, or favoring either side in a war, dispute, or contest.
> adj. Belonging to neither side in a controversy: on neutral ground.
> n. One who takes no side in a controversy.

In lieu of the stance and policy Bitcoin.org has taken regarding the subject of [Hard Forks](https://bitcoin.org/en/posts/hard-fork-policy) it is no longer possible for it to remain neutral. Leaving this statement of neutrality here on the site is misleading to site visitors as well as other contributors who have or are considering participating in the project for this very reason.

**Note:** An alternative commit for this PR could be to remove the word neutral and adjust the rest of the sentence, accordingly.


ABISprotocol:
No, as this seems to be a rather biased issue.  One could just as easily state that the hard fork policy is an attempt to make a stand for neutrality and stability; the wallets shown on the site most people would say are good examples (though many people would dispute which is best) but are always under development.  As I mentioned before in another issue, I hope that issue(s) are left open for long(er) for discussion than they have been previously in this repository, but I can see where this issue is already headed, simply attacking bitcoin.org for having a statement of neutrality.  I find that a fruitless discussion, and frankly I think that energies would be best spent elsewhere.

luke-jr:
bitcoin.org can remain neutral without promoting altcoins. The neutrality is simply scoped to *Bitcoin*.

harding:
@coderwill I'm concerned you're making this pull request not because you think we've failed to apply the mission statement correctly, but because you have other problems with the new policy and you think this is the most effective way of attacking it.

In other words, if we merged this PR, would it make you happy with the policy?

coderwill:
@harding No, this PR is because as a result of the [policy](https://bitcoin.org/en/posts/hard-fork-policy), the following mission statement is no longer true:

Mission Statement:
>Remain a neutral informative resource about Bitcoin.

Excerpt of New Policy:
>Contentious hard forks are bad for Bitcoin. At the very best, a contentious hard fork will leave people who chose the losing side of the fork feeling disenfranchised. At the very worst, it will make bitcoins permanently lose their value. In between are many possible outcomes, but none of them are good.

That said, if people think the Mission Statement compared to the Excerpt is neutral, I guess I'll just have to exist in the minority. :)

Also, just wanted to point out a couple logical fallacies in your feedback:
* If I submit a PR with a critique of a page it does not mean I am attacking.
* Assuming your question wasn't just rhetorical, I shouldn't have to be happy with the Hard Fork Policy in order to want to improve the accuracy of the Mission Statement.

luke-jr:
@coderwill Again, the new policy is not in any way conflicting with the mission statement. Bitcoin.org remains a neutral information resources *about Bitcoin*, despite not promoting altcoins (which the new policy is addressing).

harding:
@coderwill my question wasn't rhetorical.  If removing the neutrality line made you happy about the policy, a policy which clearly has made you unhappy, then I would consider that to be a strong point in favor of this PR.

Since it won't make you happy, I say NACK to this PR.

My reason is this: I find the policy on being neutral to be inspiring.  It heavily influenced my writing of the new policy, and my first outline for what became this statement (sent by GPG-signed email to several people) quoted and linked to the policy.

I understand that you don't think the new policy is neutral, but it is my best attempt at being neutral while accomplishing the goal of helping to keep Bitcoin users safe.  In that sense, the neutrality bullet point worked, and to remove it now would be to disregard a useful tool in achieving an objective we all want for Bitcoin.org.

I closed your last PR earlier than you expected, so how about I leave it up to you to decide when discussion on this PR is done?  If you think it's ready for merging, say so to me in an @ message; if you think the discussion has run itself out and it's not going to get merged, close it yourself.

coderwill:
Let's leave this open for one week until June 30, in case anyone else wants to share new or additional feedback on the proposed change.

Like I said, maybe I'm just part of a minority, here. Fair enough, if so.

saivann:
Let's get this straight; because the neutrality of one decision is questioned, it is proposed in this pull request to stop any other effort toward neutrality (effectively making things worse).

Neutrality is hard, it's never perfect. I don't see this as a valid reason for knowingly making the website a biased resource and dismissing efforts spent on neutrality (countless hours spent here). 

tl;dr; I agree these discussions are time consuming and counterproductive.

schildbach:
I agree with @coderwill that the policy change is in conflict with the mission statement. I don't agree the mission statement should be removed. Rather, the policy should be changed.

@luke-jr The policy does not only touch altcoins, but also Bitcoin forks.

luke-jr:
@schildbach No, it is intentionally worded to only apply to altcoins.

coderwill:
@luke-jr If the intent was to apply only to altcoins, perhaps the word would be referenced in the policy. Unfortunately, it is not, the word "altcoin" doesn't even appear once.

gurnec:
The current policy is "if contentious thing X happens, we'll choose side A." There's just *no way* to interpret this as neutral; neutral would be "if contentious thing X happens, we'll describe both sides A and B, and promote neither over the other."

@saivann says:
> Neutrality is hard, it's never perfect.

Yes, it's impossible to make every bit of text on bitcoin.org follow every mission statement goal; sometimes they're self-contradictory.

@harding says:
> it is my best attempt at being neutral while accomplishing the goal of helping to keep Bitcoin users safe.

As it's currently written, the mission statement doesn't mention "keeping Bitcoin users safe." I'd suggest:
 1. Modifying the mission statement's first goal to encompass user safety more generally;
 2. Clarifying that user safety *trumps* neutrality.

This would more closely fit the intent of the new policy without throwing out neutrality altogether (of course we can still argue about whether or not the new policy *actually achieves* this intent, but that's a discussion for the existing PR of the policy).

luke-jr:
@gurnec It's phrased "contentious" to avoid confusing people, but a "contentious hard fork" is in fact what an altcoin is. Bitcoin is a consensus system, and hardfork changes by design need consensus.

harding:
Everyone: I started to write replies, but I got cranky, so I'm going to simply repeat my NACK to this PR.  I think the new policy is neutral and is not in conflict with our mission statement.

I do not plan to post on this PR again.  Thank you all for your time and comments; I'm sorry I'm not able to satisfy you.

PR #920, opened 2015-06-26
Simplify javascript code for expanding lists and blocks

saivann:
There is currently 4 different javascript functions duplicating code for doing mostly the same thing; animating lists or boxes as they expand on user demand.

This pull request makes some optimizations to simplify and shorten the code, by adopting better practices such as moving styles to CSS stylesheets and using a single function to do the animation.

At the same time, some big blocks of related CSS code used only for IE8's compatibility are moved into a separate file to reduce the load for most browsers and simplify compatibility maintenance.

I have tested the changes with IE8-9-10 and various OP, CH, SA, FF versions.

Live preview:
http://bitcointest1.us.to/en/press
http://bitcointest1.us.to/en/secure-your-wallet
http://bitcointest1.us.to/en/choose-your-wallet
http://bitcointest1.us.to/en/development

In the absence of critical feedback, this pull request will be merged on June 29th.


harding:
Reviewed code and lightly tested. LGTM b27bcaf  Thanks!

PR #921, opened 2015-06-26
Use standard proper noun capitalization for Trezor to maintain consistency throughout site.

coderwill:
pryds, who is coordinating a good bit of the Dutch translation, was asking why TREZOR wasn't in all-caps in the Mycelium Android Wallet description. He/She wasn't sure how it should be cased in the Dutch translation - this PR puts in all caps to consistently match the other mentions on the site.


gurnec:
Just for (brief) discussion—should it be "Trezor" instead? Wikipedia uses the rationale: excluding all-caps trademarks

> avoids drawing undue attention to some subjects rather than others [...] even if the trademark owner considers nonstandard [all-caps] formatting "official"

([reference][1])

I think either way is fine, as long as it's consistent.

[1]: https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Trademarks#General_rules

saivann:
FWIW, I had the same sentiment when looking at TREZOR vs Ledger Nano. I think doing the same as Wikipedia would make a lot of sense.

coderwill:
@gurnec @saivann - thanks for the feedback - I just swapped it so that it's the other way around, no instances of all caps.

saivann:
@coderwill There is a remaining all-caps TREZOR in the choose-your-wallet template. A useful trick; you can use "git grep TREZOR" or "grep -r TREZOR --color" to find all occurences of something to make sure you can fix them all.

coderwill:
@saivann Ah thanks, I see it - I was only doing a search inside the strings file.

coderwill:
Updated the two additional references on the Choose Your Wallet page.

saivann:
Commit c9855a8 LGTM, thanks!

I think not allowing uppercase trademarks fits pretty closely with previous pull requests for not allowing superlatives in wallet descriptions.

harding:
c9855a8 LGTM too, and I strongly approve of the policy of avoiding all-caps when reasonable.

In the absence of critical feedback, this will be merged Saturday.

crwatkins:
LGTM.  Kudos to @gurnec for the precedence reference at Wikipedia.  I will personally adopt those styles.

harding:
Yeah, the Wikipedia style guide is pretty good for online documentation.  It's our [fallback for the dev docs](https://github.com/bitcoin-dot-org/bitcoin.org/wiki/Documentation-Style-Guide):

> For all style matters not explicitly covered in this guide, the authority is the [Wikipedia Manual Of Style.](http://en.wikipedia.org/wiki/Wikipedia%3aManual_of_Style)

PR #922, opened 2015-06-26
DevDocs: update a mis-directed link and clarify SIGHASH_SINGLE definitions

mruddy:
First, there was a mis-directed link that was going to the SIGHASH_NONE definition page.

Second, the definitions for SIGHASH_SINGLE didn't seem to be consistent with the current code. The definitions at https://en.bitcoin.it/wiki/OP_CHECKSIG seem much closer. The code I'm looking at is https://github.com/bitcoin/bitcoin/blob/bc60b2b4b401f0adff5b8b9678903ff8feb5867b/src/script/interpreter.cpp#L-1024-1101

Hopefully, these definitions are closer to how it actually works.

### DIFF
--- a/_data/glossary/en/sighash-single.yaml
+++ b/_data/glossary/en/sighash-single.yaml
@@ -7,10 +7,10 @@ required:
     title_max_40_characters_no_formatting: SIGHASH_SINGLE
 
     summary_max_255_characters_no_formatting: >
-        Signature hash type which signs the output corresponding to this input
+        Signature hash type that signs the output corresponding to this input
         (the one with the same index value), this input, and any other inputs
-        partially. Allows modification of other outputs. The partial signing of
-        other inputs allows modification of their sequence numbers.
+        partially. Allows modification of other outputs and the sequence number
+        of other inputs.
 
     synonyms_shown_in_glossary_capitalize_first_letter:
         - SIGHASH_SINGLE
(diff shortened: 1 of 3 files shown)


_data/glossary/en/sighash-single.yaml
```diff
@@ -7,9 +7,10 @@ required:
     title_max_40_characters_no_formatting: SIGHASH_SINGLE
 
     summary_max_255_characters_no_formatting: >
-        Signature hash type which only signs its input and the output
-        with the same index value, allowing modification of other inputs
-        and outputs.
+        Signature hash type which signs the output corresponding to this input
+        (the one with the same index value), this input, and any other inputs
+        partially. Allows modification of other outputs. The partial signing of
+        other inputs allows modification of their sequence numbers.
```
harding: @mruddy this text is good, but is there any chance you can edit it down to 255 characters or less?  I think dropping the final sentence would be fine since it links to the full description.

harding:
Except for the one character limitation comment above, commits up to b01af43 LGTM.  Thanks, @mruddy!

mruddy:
@harding Thanks for the feedback. I just refined that definition to make it shorter. Glad I could help!

harding:
LGTM.  Thanks!

In the absence of critical feedback, this will be merged Monday.

PR #923, opened 2015-06-26
Load Youtube iframe on click only

saivann:
Live preview: http://bitcointest2.us.to/en/

This change improves performances for the landing page by not loading a remote iframe everytime a visitor opens the website, and instead mimick YouTube until the user clicks to load the video. The iframe is then appended on user demand with autoplay=1. The result is seemless, but faster.

This also improves privacy by cutting off Youtube from receiving data from bitcoin.org's visitors. This issue was outlined by me in a previous discussion with @luke-jr .

It's worth to report that this introduces a small bug for mobile users, they now need to tap twice in order to start the video. This sort of issue does not seem uncommon and given the strong improvement in performances for mobiles, I think the benefits outpace this drawback (which might later be fixable with some Youtube API quirks I haven't much explored at the moment).

Edit: Last commit also fixes a bug in the check-for-subheading-anchors rule of the Makefile which was preventing this pull request to pass Travis build.


harding:
Tested LGTM Firefox 34.4.0.  Also used the network profiler to verify no off-site resources are loaded until the video is clicked.  Thanks, @saivann, I like both the privacy enhancement and the visitor bandwidth savings!

saivann:
@harding Thanks!

Note: I've made some last-minute minor changes, making the javascript function reusable elsewhere without hardcoded variables and moving it to main.js for that purpose (and to avoid duplicated code). Live preview updated and plenty of old and recent browsers tested.

In the absence of critical feedback, this pull request will be merged on June 29th.

louisjc:
Just for fun:
[June 26](http://tools.pingdom.com/fpt/#!/dgm19O/https://bitcoin.org/en/) 622.4kB - 1.29s
[June 29](http://tools.pingdom.com/fpt/#!/A06Xl/https://bitcoin.org/en/) 165.1kB - 734ms
Bitcoin.org became so much faster in just 3 days!

PR #924, opened 2015-06-27
Add Wladimir's Releases Key

harding:
From [this message](https://www.mail-archive.com/bitcoin-dev@lists.linuxfoundation.org/msg00126.html):

    -----BEGIN PGP SIGNED MESSAGE-----
    Hash: SHA512

    Hello,

    Starting with 0.11.0rc3, SHA256SUMS.asc will be signed with the following key:

        pub   4096R/36C2E964 2015-06-24 Wladimir J. van der Laan (Bitcoin Core binary release signing key) <laanwj@gmail.com>
        Primary key fingerprint: 01EA 5486 DE18 A882 D4C2  6845 90C8 019E 36C2 E964

    For gitian and commit signing I will keep using this key.

    Wladimir
    -----BEGIN PGP SIGNATURE-----
    Version: GnuPG v1

    iQEcBAEBCgAGBQJViphCAAoJEHSBCwEjRsmmtRoIALBzJMGXzoj5t9OQSedxjnjP
    sxfHuBwQxeuPYXbRlMjY5UZhmabbt0/mLRfVSdscnCzp0YxbMRwD7I6MdHqXyBtd
    oS+TUfMNir5lk7Ti2hRStgvxqsAbHUJ08LlqpJXV5dq3QgeJyJwZM76a6yyaGwxP
    SwqvKklQZ/qdrKOgjjn6d5HywgsmybJSDzEDR3k+ogkLsfM1jcpqZhwFeRVpk94m
    SgZGLLx5zAIKcLHn4I1FaZ+OAmmS0ukYcmotMOUk6NBEjHTDfjEFBrbrlwvL4G7r
    kjd1mRxkaJMxX3nJicXiEQClVoeUrMVyJrrsTGyPixSicdQbItuyLWXm37fAfE0=
    =4v49
    -----END PGP SIGNATURE-----


gmaxwell:
ACK

gmaxwell:
ACK

saivann:
LGTM, although wouldn't it be more useful to include both keys into laanwj.asc?

There is likely plenty of links pointing to that key, and there is less chances that people might end up confused because they didn't get the right key. We also wouldn't need two PGP keys linked on the [development](https://bitcoin.org/en/development) page.

laanwj:
Makes sense. ACK.

@saivann one argument to keep the keys separate is that they'll be used for separate purposes: the PGP link with the mail is there mainly to mail privately about security issues, the release key will only sign `SHA256.asc`. Though I also agree downloading them all at once is more convenient.
(but taking that argument further, so would an `.asc` with all developer keys...)

saivann:
@laanwj I'm not sure if you're for or against a single key by reading your comment? Either way, my main concern is about people who already refer to the existing link, and new users who will read "binaries are signed by Wladimir" and get your public key for that purpose. If we keep two keys, I think it'd be nice to add a commit to link to it from the Development page too (maybe the layout may need some adaptation).

harding:
My main reason for keeping them separate was that I assumed they had different security properties, maybe with the code signing key being kept on an offline computer.[1]  I thought keeping them separate would help prevent people who only wanted the code signing key from accidentally importing both and, if the regular key becomes compromised, not noticing that a malicious binary was signed by the wrong key.

I'm happy to combine them if you want.

For listing the code signing keys, we should probably have a list of what versions are signed by which keys.  Maybe just a single sentence added to the Development and Download pages?  I'll look into that.

[1] I do this: my master signing key has never been on an online computer, but the encryption and signing subkeys I use every day are on my regular laptop.  In case my laptop is compromised, I can revoke the daily keys while keeping the master key useful.  I also use the master signing key for signing things that require the extra strong non-repudiation property.

harding:
Added commit 9d351edbab13626f72f88b3dcfb97084b4ade99b which adds the following text to the Download page (highlighted here in green): 

screenshot-btcorg localhost 2015-06-27 16-50-45

On mobiles, it looks like:

screenshot-btcorg localhost 2015-06-27 16-51-15

I only checked as far back as 0.8.6 because that's all we have hosted in `/bin` and the old SourceForge files are unavailable.

harding:
Added commit 3a89870ad7f15cad7750d20cdc72c162a12e0438 by @saivann which much improves upon my crude attempt at listing keys.  Screenshot below:

screenshot

I tested it, and it LGTM.  On https://github.com/harding/bitcoin.org/pull/21#issuecomment-116183461 @saivann also said, "adding links to keys in the download page mostly makes me neutral now regarding my previous suggestion [to combine the keys]."

So, if there are no further problems identified and everyone is ok with separate keys, I'll merge on Monday.  Thanks everyone!

saivann:
@harding Thanks, sounds good!

PR #929, opened 2015-07-02
Airbitz update

i3inary:
Updates to choose your wallet 6 months later.


crwatkins:
@i3inary Thanks for updating your entry.

I see you have updated from centralized servers to random servers.  I was not involved in your initial review, but I believe Airbitz (in the past) connected to Obelisk servers run by Airbitz.  I assume that has changed.  Can you describe the new situation including the number of servers and the selection method?  Does it still involve getting this list of servers from a different Airbitz controlled server?  Thanks for any light you can shed on these changes.

paullinator:
Hi @crwatkins. The method of choosing a server is at random. The list of servers is pulled from an Airbitz server, although the app will only update once a day. There are 8 total servers up and running right now run by both Airbitz and 3rd parties.

crwatkins:
Hey @paullinator, thanks for clarifying that.  I'm all for updating Airbitz to ```checkpasstransparencyopensource```, but I can't recommend the change to ```checkpassvalidationservers```.  I believe ```checkfailvalidationcentralized``` is currently more appropriate given that using only 8 servers is pretty "centralized" in my book.  If I'm not mistaken, Airbitz runs a majority of them.  I really like where you are going with your server decentralization, and I applaud your efforts, but I don't think we are quite there yet.

paullinator:
@crwatkins Thanks for the feedback. We're constantly pushing for broader adoption and we will have more 3rd party nodes launched (including some by Open Bazaar). What would you consider a fair number to be considered checkpassvalidationservers?

crwatkins:
Good question, Paul.  Let's review the type of servers used by some of our listed wallets.  It seems that our "gold standard" (not a real score) number of servers is at about 6000, involving connecting randomly to full nodes.  The next level that we seem to have been listing is about 50 active, mostly independently run, servers.  For example, Electrum, GreenAddress, and Coinomi use Electrum servers.  Our next level seems to involve only a few servers, usually run by the same organization.  Examples here are Insight/BWS servers, Mycelium, and Obelisk/bs servers.

Currently the only wallet scored with ```checkpassvalidationservers``` is GreenAddress.  Personally, I wouldn't feel very private having only 50 servers to choose from, but I would have trouble arguing for a number much larger than 50, if only for consistency reasons (and I'm for as much consistency as we can stand).

Thanks for your support of Obelisk/bs. It certainly provides some interesting and unique capabilities.

i3inary:
Updated.

crwatkins:
Thanks for the update, @i3inary; however I think you might have meant to revert validation and not transparency.

i3inary:
Sorry slipped by me.  Should be updated now.

crwatkins:
Thanks @i3inary.  LGTM.

harding:
Preview of  17307d0 LGTM also.  Thanks!

In the absence of critical feedback, this will be merged Friday.

PR #931, opened 2015-07-03
Simplify javascript code for wallets

saivann:
Sorry in advance for these large changes. A lot of code isn't really modified, only reorganized.

Live preview: http://bitcointest1.us.to/en/choose-your-wallet

It appeared to me that the javascript code for wallets was a bit unreadable with confusing function names, so I bundled many functions dedicated for events on the page into two single functions (walletMenuListener and walletListener) with subfunctions (closures) and tried to shorten / improve syntax a bit elsewhere.

This pull request also fixes a bug where the bubble of a wallet would show if the mouse is hovering over the icon after the user clicked within the bubble, as well as fixing a few unintended use of global variables through the code.

I have tested these changes against IE 8-9, FF, SA, CH, OP, Android 2.3.6

In the absence of critical feedback, this pull request will be merged on July 6th.

### DIFF
--- a/_templates/choose-your-wallet.html
+++ b/_templates/choose-your-wallet.html
@@ -878,7 +878,7 @@ wallets:
 <h1>{% translate pagetitle %}</h1>
 <p class="summary">{% translate pagedesc %}</p>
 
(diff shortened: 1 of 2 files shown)


saivann:
Quick update; I am waiting for additional feedback from @crwatkins about iPad bugs with click events.

saivann:
This pull request has been replaced by #950

PR #932, opened 2015-07-03
Hide wallets on click, not on mouseout

saivann:
Live preview: http://bitcointest2.us.to/en/choose-your-wallet

This is a change I've noticed on bitcoin.com which I like. I often found that I have trouble navigating between wallets as the mouse easily slips outside of the bubble, which forces to me click on the wallet again to re-open the bubble. Opening and closing that bubble on click seems more consistent and convenient to me (and is actually simplier to implement).

This change keeps compatibility with javascript-disabled browsers, thanks to :hover CSS rules only overriden by javascript when it's enabled.

I have tested this pull request against IE 8-9, FF, SA, CH, OP, Android 2.3.6

Edit: I had to apply the same behavior to the menu following a bug report from @crwatkins

### DIFF
--- a/_less/screen.less
+++ b/_less/screen.less
@@ -1295,7 +1295,6 @@ div.post {
 	background-repeat:no-repeat;
 	background-position:left 6px;
 	color:#383838;
-	cursor:pointer;
 	margin:15px 10px;
 	font-size:130%;
 	font-weight:bold;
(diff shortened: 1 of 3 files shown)


saivann:
This pull request has been replaced by #950

PR #936, opened 2015-07-04
Alerts: SPV Mining more details

harding:
This adds more details to the current alert message.  With F2Pool now fully validating, I've also reduced the number of recommended extra confrimations to wait from 30 to 15.  According to @gmaxwell's online version of the [fork probability calculator](https://people.xiph.org/~greg/attack_success.html), 15 confirmations on the longest chain from 25% of hash rate has a probability of 0.01%.

Preview: http://dg0.dtrt.org/en/alert/2015-07-04-spv-mining

CC: @petertodd @luke-jr

### DIFF
--- a/_alerts/2015-07-04-spv-mining.html
+++ b/_alerts/2015-07-04-spv-mining.html
@@ -7,14 +7,16 @@ banner: "WARNING: many wallets currently vulnerable to double-spending of confir
 <p><em>This document is being updated as new information arrives.  Last
 update: 2015-07-04 09:00 UTC</em></p>
 
-{% assign confs="15" %}
+{% assign confs="30" %}
 
 <h2 id="summary">Summary</h2>
 
 <p>Your bitcoins are safe if you received them in transactions confirmed before 2015-07-04 08:00 UTC.</p>
 
-<p>After that time, confirmation scores are not as reliable as they
-usually are for users of certain software:</p>
+<p>However, there has been a problem with a planned upgrade. For
+   bitcoins received later than the time above, confirmation scores are
+   significantly less reliable then they usually are for users of
+   certain software:</p>
 
 <ul>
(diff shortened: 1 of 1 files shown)


luke-jr:
0.01% is too high IMO.

harding:
@luke-jr I'll leave it at 30 then until we're sure more hash rate is safe.

saivann:
@harding Thanks for writing this. Please find fixes for broken HTML here, last commit on this branch:
https://github.com/bitcoin-dot-org/bitcoin.org/commits/spv-smallfixes

I can't comment on the accuracy, but otherwise LGTM!

harding:
Commit https://github.com/bitcoin-dot-org/bitcoin.org/commit/c04bfa6832d4032afe271dffbec6f56aa6ad716a changes the extra confirmations back to 30 and removes the text about SPV wallets being more vulnerable than usual because of expected soft fork problems.  (Requested by @gmaxwell, who just said it looks good on IRC)

harding:
Pushed commit 9f30adba9fdcb59be50aa6a399938e6045409d4f to master with the following changes:

- Move list of good/bad pools to the wiki.  If we suspect people are lying about what validation their pools do, we can move the list back here; otherwise, I think letting people list their own pools will be a more productive use of everyone's time.

- Suggest upgrading to 0.10.2 because of DoS vulnerabilities in earlier versions unrelated to this alert.

- Update the times

saivann:
@harding Thanks! Just fixed a typo with an additional commit.

harding:
@saivann remember you need to sign tip commits (or update manually on the build server).  I have another small change to make as [recommended by reddit](https://www.reddit.com/r/Bitcoin/comments/3c305f/if_you_are_using_any_wallet_other_than_bitcoin/csrvunr), so I'll push that in a moment after I see if there are any other good suggestions.

PR #939, opened 2015-07-04
Avoid reinstalling dependencies with bundler on every build

saivann:
This change brings significant performance improvement for the build process, as the build script otherwise deletes ./vendor, which forces bundler to reinstall every dependency from scratch on every build, which then delays the build for approximately 3 minutes.

It also seems to me that downloading and installing dependencies on every build exposes us to some more third party risk for no reason.

Git clean man file for the -x option:
https://git-scm.com/docs/git-clean

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,11 @@ install-deps-development:
 
 ## Install dependencies (deployment version)
 install-deps-deployment:
+ifdef BUNDLE_DIR
+	bundle install --deployment --without :slow_test --path=$(BUNDLE_DIR)
+else
 	bundle install --deployment --without :slow_test
+endif
 
 ## Pre-build tests which, aggregated together, take less than 10 seconds to run on a typical PC
 pre-build-tests-fast: check-for-non-ascii-urls check-for-wrong-filename-assignments \
--- a/_build/update_site.sh
+++ b/_build/update_site.sh
@@ -9,9 +9,12 @@ source /etc/profile.d/rvm.sh
 
 AUTHORIZED_SIGNERS_DIR='/bitcoin.org/auto-build-committers.gnupg'
 REPO='https://github.com/bitcoin-dot-org/bitcoin.org.git'
+BUNDLE_DIR='/bitcoin.org/bundle'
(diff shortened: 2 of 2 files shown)


harding:
@saivann huh.  I think that's a regression, as I don't remember it doing that when I first set it up.  (But I just tested, and you are correct)  However, I haven't been able to find a plausible commit where anything changed.

I think dropping `-x` is the good option for now.  However, if we want it back later, we can add `--path=/some/directory` to the `install-deps-deployment` target in the Makefile.  (See `bundle help install` for details.)

LGTM.  Suggest merging whenever you're ready to manually update the live build script.  Thanks for catching this!

saivann:
@harding Ah... for a moment I forgot that the repository is copied to a temporary directory before "bundle install" is ever called, so my previous solution just wouldn't work. I've just implemented your second suggestion which I think should be working fine this time. Thanks!

Re Regression: At least that's now the intended behavior according to their [current documentation](http://bundler.io/v1.7/man/bundle-install.1.html#DEPLOYMENT-MODE)

    Gems are installed to vendor/bundle not your default system location

saivann:
Note; I haven't implemented this solution on the build server yet. If there is no opposition, I will apply these changes to the VM later today.

harding:
4e40909 untested LGTM.  Thanks!

saivann:
Thanks! I just finished testing and the script is working as intended.

harding:
I don't know if this is related, but the about-us and development pages aren't displaying the contributor listings from github.  I'm looking into it now.

saivann:
@harding Ah, I restarted the builds quite many times and the GitHub API has a limit on the number of calls you can make per minute. I would really not be surprised if that's the only reason. This API call use to fail once in a while, but works fine the vast majority of the time.

harding:
@saivann yeah, that was my first guess too (I used to have the issue for local builds all the time until you added selective plugin disablement), and it seems to have been the correct guess: I triggered a rebuild of the site on the server with no other changes, and it looks fine now.  thanks!

saivann:
@harding Thanks a lot for watching at those issues so carefully!

Issue #941, opened 2015-07-05
Build script does not terminate properly when detecting concurrent build


saivann:
I've worked on this issue for a few hours with incomplete conclusions.

The current build script on the build VM should (and previously did) stop to build as soon as it detects another build script starting on a more recent commit. It turns out that the build script still detects concurrent builds and terminate accordingly. However, persistent resource intensive ruby processes indicate that the child processes called by the script in fact do not terminate.

The following lines should in theory kill all sub-processes as soon as the script is exiting. But it turns out that they seem to be somehow ignored for some reason.

~~~
trap "rm -rf $WORKDIR; kill 0; exit 1;" EXIT
trap "rm -rf $WORKDIR; kill 0;" SIGINT
~~~

As a result, it suffices for a few commits within a short timeframe to start multiple simultaneous build scripts, all of which are uselessly slowing down each other for no reason by eating RAM and CPU. In the worst case scenario, this can lead to extremely long delay for website updates or crashing the VM, although this seems unlikely in our case as we don't push commits constantly.

/tmp also gets polluted to a point that it risks filling up all disk space, leading to impredictable behavior from the build script. For instance, 9522 folders (most being empty) did accumulate leading to using 100% of the disk space. It seems like periodic reboot (clearing /tmp) is what prevented that issue to be noticed.

Confusingly, this issue didn't happen before (and still don't happen locally on my PC). Upon further testing, I noticed a conclusive fact; when the following line is commented, the build script deletes his temporary directory as expected when the script exits (although it fails to build due to RVM not being configured properly), which suggests that a solution could be close.

~~~
source /etc/profile.d/rvm.sh
~~~

@harding You've completed a fair amount of work with configuring the VMs with RVM and bundler. I was wondering if you had interest for investigating that issue further and see if you could find a workaround?

harding:
Working on it now.  I've always wanted to improve that part of the script to use my typical locking anyway.  Thanks!

saivann:
@harding Thanks!

PR #944, opened 2015-07-05
Build: Push Function Include Down Into Sub-Shell

harding:
Fixes issue (bug) #941 

While installing rvm as root (which is a pain) for a test environment, I realized what the problem with the current script was:

- `kill 0` kills all process in the current process group.  That includes *children* of the current shell, but not the shell itself (which has the pgid of its parent shell).

- By sourcing `rvm.sh` we were including the `rvm` shell function into the shell.  Unlike commands such as git, rsync, etc..., `rvm` wasn't a child process.

- So, when `kill 0` was run, it had no effect on the `rvm` function.

The solution was easy: push the sourcing of the `rvm.sh` function down a level into the build process sub-shell.

This PR has been implemented and well tested on the build server (and implemented and lightly tested on the translation preview server).  The testing on the build server included:

- A regular build. Rsync was run and the tmp directory in the /tmp directory was removed as expected.

- A build with an artificially-induced error.  The build aborted and the tmp directory was removed.  Rsync was not run.

- Two successive builds: when the second build started, the first build aborted and its tmp directory was removed.  When the second build finished, rsync was run and the tmp directory was removed.

This PR also now prints a message on success, as the shell-generated "Terminated" message after a successful run confused me for a bit.

I think this is all I'm going to do for now.  I have an idea for queuing of builds using lock files, but now that the main problem is fixed, I find I'm not so enthusiastic about implementing it. :-)

As noted, the solution is already deployed, so I think this PR can be merged as soon as @saivann has reviewed it. (It can be merged using the GitHub merge button since it doesn't affect website content.)

### DIFF
--- a/_build/update_site.sh
+++ b/_build/update_site.sh
@@ -5,8 +5,6 @@
 
 PATH=/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/bin
 
-source /etc/profile.d/rvm.sh
-
 AUTHORIZED_SIGNERS_DIR='/bitcoin.org/auto-build-committers.gnupg'
 REPO='https://github.com/bitcoin-dot-org/bitcoin.org.git'
 BUNDLE_DIR='/bitcoin.org/bundle'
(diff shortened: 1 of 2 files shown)


saivann:
@harding Ah, that makes sense. I've tested the script locally and it works perfectly. Thanks a lot for completing my investigation (and for updating the translation script on GitHub and on the VM as well)!!

PR #946, opened 2015-07-06
Alerts: Update SPV Mining Alert

harding:
Preview: http://dg0.dtrt.org/en/alert/2015-07-04-spv-mining

- Add bolded notice to top telling people the event is on going.  Some people on Reddit think it was a one-time fork.

- Link Electrum users [here](https://en.bitcoin.it/wiki/July_2015_Forks#Electrum) for help finding a Bitcoin Core 0.10.x-powered server.

- Add a table of forks related to SPV mining.  Link to a list of invalid block hashes.

- Remove hopeful note that miners were enabling full validation.  This appears not to be the case. :-(

- Add an update mentioning the July 5th fork

Also, used a hack to enable Markdown support.  (Markdown allows inline HTML, so I didn't reformat anything I didn't have to.  Tomorrow I may convert the whole thing to Markdown.)

### DIFF
--- a/_alerts/2015-07-04-spv-mining.html
+++ b/_alerts/2015-07-04-spv-mining.html
@@ -126,7 +126,7 @@ The paragraphs that follow explain the cause more throughly.
 
 <p>The recommended fix, which was attempted, was to
    get all miners off of SPV mining and back to full validation (at
-   least temporarily). If this happens, Bitcoin.org will reduce our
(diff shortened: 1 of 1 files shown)


saivann:
@harding Agreed with converting it all to markdown, that'll be helpful as an example for future alerts. Please feel free to pick what you think is relevant from #947, otherwise this LGTM! Thanks!

PR #948, opened 2015-07-06
Allow Use Of Markdown For Alerts

harding:
This PR makes it possible to write alerts in Markdown without having to use any hacks, and changes the current SPV mining alert to the new format (minimal changes).  (HTML alerts still work and were tested.)

The changes to `_plugins/alerts.rb` consist entirely of getting it to use the right file name as the `.html` extension had been assumed.

This PR is not urgent, so there's no rush to review.


saivann:
I was skeptical that any change was needed in alerts.rb (or that the destination file should be a .md file), but after diff-ing the final pages, that's indeed the right way to do it! Thanks!

I opened a separate pull request to convert all HTML to markdown, however I don't think it should be blocking this PR if there is any issue with it or if it needs time to review.

harding:
@saivann yeah, it confused me for a while too.  I thought all that was broken was the alert banner, but the makefile tests caught the problem.  Thanks for reviewing!

In the absence of critical feedback, this pull will be merged (late) Tuesday.

PR #950, opened 2015-07-07
Rework menus for mobile compatibility

saivann:
(This pull request replaces #931 and #932)

Live preview: http://bitcointest4.us.to/en/choose-your-wallet

This is an important re-design of the underlying javascript functions and CSS stylesheets used for navigation menus (top menu and wallet menu).

This pull request should fix several navigation bugs and issues for iOS, as well as speeding up navigation and preventing accidental clicks on Android, by making good use of the touch events.

The dropdown menus on mobile devices also now avoid using CSS :hover, and can be expanded or shrinked by clicking on the parent element of any menu entry. This avoids confusion and accidental clicks by moving elements on the screen only on user request.

The wallets are now also populated, displayed and hidden on click and not on mouseover, which prevents accidentally closing a wallet bubble only because the pointer of the mouse slipped out of the bubble, and allows to shorten and simplify javascript code quite a lot. This also fixes a bug where a wallet would inconsistently display on mouseover after user clicked inside its bubble.

Remaining changes include improving the syntax and comments for the related CSS and javascript code, as well as fixing a few accidental use of global variables.

Tested on OP 9,12, FF3.5,38, SA 4, IE 6-10, CH 43, Android 2.3.6
Tested English and Arabic RTL layout
Tested website with javascript disabled

Since this is a massive pull request, I do not expect someone to fully review the code before merge. More testing of the live preview above is very appreciated, especially on iOS devices.

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -41,14 +41,14 @@
       </ul>
       </li>
     </ul>
-    <a id="menumobile" class="menumobile" onclick="mobileMenuShow(event);" href="#"></a>
+    <a id="menumobile" class="menumobile" onclick="mobileMenuShow(event);" ontouchstart="mobileMenuShow(event);" href="#"></a>
     <a class="logo" href="/{{ page.lang }}/"><img src="/img/icons/logotop.svg" alt="Bitcoin"></a>
     <div id="langselect" class="langselect"><select onchange="window.location=this.value;">
     {% for lang in site.langsorder %}{% assign active = ''%}{% if lang == page.lang %}{% assign active = ' selected="selected"'%}{% endif %}
           <option value="/{{ lang }}/{% translate {{page.id}} url {{lang}} %}"{{ active }}>{{ site.langs[lang] }}</option>
     {% endfor %}
     </select></div>
-    <ul id="menusimple" class="menusimple" onclick="mobileMenuHover(event);">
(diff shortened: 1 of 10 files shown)


saivann:
In the absence of critical feedback, this pull request will be merged on July 13th.

harding:
Testing this out on a phone, it looks like the only options are Mobile, Desktop, Hardware, and Web:

screen shot 2015-07-04 at 12 19 09

I wonder if it would be better to default to opening the Mobile menu on mobile devices:

screen shot 2015-07-04 at 12 17 25

That way visitors know they can filter down to their particular device.

I realize that this isn't really a comment on the code changes in this PR---it just happened to occur to me while reviewing since I don't usually visit this page---so let me know if I should open a separate issue or my own PR about it.

harding:
Note to any other reviewers: changing languages doesn't work on @saivann's preview, but it does work for me on a local build.  I think for the preview @saivann just did a quick build without any of the translations enabled.

saivann:
@harding Thanks, that makes sense! I've just added a commit and updated the live preview.

Re translations enabled: Indeed, I've enabled English and Arabic translations only for that live preview to save time. This should be enough to test both normal and RTL layout, but I can include more languages if needed.

harding:
LGTM 3ac6158.  Reviewed code and moderately tested using FF on desktop and an ancient Android phone.  Thanks!

PR #957, opened 2015-07-10
Layouts: Allow Alternative Base Layouts & Add Breadcrumbs

harding:
Preview: http://dg1.dtrt.org/en/developer-documentation

This pull contains three commits:

1. **Split the base layout into multiple files:** this makes it easy to create alternative layouts for collections of pages while reusing as much code as possible (so updates are easy).  This pull does not use this feature, but I hope to use it in a future pull to create a "sub site" with a dozen pages dedicated to Bitcoin Core.

    Note: this commit diff is quite ugly because it produces HTML output that is identical to current master HEAD.

2. **Add whitespace to base layouts:** this whitespace-only commit fixes the ugliness of the previous commit.

3. **Add breadcrumbs to dev docs:** this changes the base template to allow adding breadcrumbs to pages, and proceeds to add them to all dev docs.  They can be added to other pages too, but the dev docs seemed like a good place to test them out.  Here's what they look like:

    screenshot-dg1 dtrt org 2015-07-10 11-46-10

### DIFF
--- a/_includes/layout/base/breadcrumbs.html
+++ b/_includes/layout/base/breadcrumbs.html
@@ -1,0 +1,28 @@
+{% comment %}
+This file is licensed under the MIT License (MIT) available on
+http://opensource.org/licenses/MIT.
+{% endcomment %}
+
+<div class="breadcrumbs">
+  {% for crumb in page.breadcrumbs %}
+    {% if forloop.last %}
+      {% case crumb %}
+        {% when "GLOSSARY_ENTRY_TITLE" %}
+          {% comment %} ## Only show first synonym in crumb ## {% endcomment %}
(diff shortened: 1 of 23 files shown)


saivann:
@harding I haven't had time to review and test the changes yet but I think it's generally excellent ideas. If there is no hurry, I will likely be able to take a look very soon.

harding:
@saivann there is no hurry, I'm still at least two weeks away from needing this merged.  I'll try to rebase this tomorrow (Saturday) to fix the merge conflict.

saivann:
LGTM up to commit 7d74941, thanks! That makes the layout way more understandable and modular.

I verified that the final output is identical (beside whitespaces) and content of each file in _includes is the same as the current base.html layout.

harding:
@saivann thanks!

In the absence of critical feedback, this will be merged Monday.

PR #964, opened 2015-07-15
Alerts: Disable SPV Mining Alert

harding:
This PR removes the alert banner from every page of the site and changes the bolded text near the top of the alert to:

screenshot-btcorg localhost 2015-07-15 08-59-50

I'm not sure we should be disabling the alert banner: if I understand correctly, several large miners have decided to stay with SPV mining (or worse, previous-header-hash-only mining).

On the other hand, I'm not aware of any short-term proposals to restore SPV wallet confirmation reliability to what it was when every miner did full validation---so the alternative seems to be keeping this alert up indefinitely, and I don't want to do that.

Comments appreciated.

### DIFF
--- a/_alerts/2015-07-04-spv-mining.md
+++ b/_alerts/2015-07-04-spv-mining.md
@@ -1,21 +1,23 @@
 ---
 title: "Some Miners Generating Invalid Blocks"
 alias: "spv-mining"
-active: true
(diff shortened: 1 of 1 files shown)


saivann:
Untested LGTM e509a92

harding:
Since the Reddit and Bitcoin Talk alerts have been un-stickied, I'm going to merge this pull in a couple hours if there are no further comments.

PR #966, opened 2015-07-19
Copyrights: Add MIT License Header To More Files

harding:
This pull updates the existing makefile test for copyright headers to also cover all files with a YAML header except those in the `_alerts` directory.  (I don't want anything to stop us from quickly publishing an alert.)

The rest of this pull adds any missing copyright headers.  I believe all the files with newly-added headers either came from the Bitcoin Core project (release notes), which is MIT licensed, or were created by @saivann or me.

### DIFF
--- a/404.html
+++ b/404.html
@@ -1,4 +1,7 @@
 ---
+# This file is licensed under the MIT License (MIT) available on
+# http://opensource.org/licenses/MIT.
+
 layout: base
 lang: en
 id: index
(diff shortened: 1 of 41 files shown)


saivann:
Commit e038560 LGTM, thanks for improving the Makefile rule!

harding:
@saivann thanks!

In the absence of critical feedback, this will be merged Sunday.

PR #972, opened 2015-07-24
Remove Mentions Of Low Fees And Instant Transactions

harding:
Preview: http://dg4.dtrt.org/en/

The current site text makes repeated mention of Bitcoin providing low transaction fees and instant transactions, as well as strongly implying that any transaction paying a fee or having medium priority will be confirmed in the next block.

As instant transactions can be unsafe and the other things are less likely to be true in as available block space becomes fully utilized, this pull removes or clarifies those sections.

To keep this pull easy to review, preference was given to removing sections rather than replacing them with new sections.  (In a few cases, this required commenting out adjacent text to keep page layouts balanced.) Clarifications were kept as minimal as possible also to make review as easy as possible.

I suggest that replacement sections or more comprehensive revisions be submitted as a separate pull request.

## Removals

Screenshots below are from current Bitcoin.org with removals highlighted.  Links are to the pages on the preview site with this commit applied.

[Main page:](http://dg4.dtrt.org/en/)

screenshot-bitcoin org 2015-07-23 18-48-23

[Bitcoin for individuals:](http://dg4.dtrt.org/en/bitcoin-for-individuals)

screenshot-bitcoin org 2015-07-23 22-29-02

[Bitcoin for businesses:](http://dg4.dtrt.org/en/bitcoin-for-businesses)

screenshot-bitcoin org 2015-07-23 18-56-05

[Bitcoin for developers:](http://dg4.dtrt.org/en/bitcoin-for-developers)

screenshot-bitcoin org 2015-07-23 19-03-11

[FAQ:](http://dg4.dtrt.org/en/faq)

(Note: many other non-removal changes were made; see the clarifications section below.)

screenshot-bitcoin org 2015-07-23 19-16-09

[Innovation:](http://dg4.dtrt.org/en/innovation)

screenshot-bitcoin org 2015-07-23 19-42-10

[Press:](http://dg4.dtrt.org/en/press)

screenshot-bitcoin org 2015-07-23 19-45-19

## Clarifications

Please see the diff for text changes.


404.html
```diff
@@ -15,12 +15,12 @@
 <!--Fallback page, don't allow indexing-->
 <meta name="robots" content="noindex">
 <h1>An open source P2P digital currency</h1>
+{% comment %}
 <p style="font-size:125%;">Bitcoin is a currency, a protocol, and a software that enables
 <ul style="font-size:125%">
-  <li>Instant peer to peer transactions</li>
```
luke-jr: I don't think this needs to go. Instant payments may not be secure, but they're just fine for getting money from friends and family (what "peer to peer" suggests to me here).

404.html
```diff
@@ -15,12 +15,12 @@
...
   <li>Worldwide payments</li>
-  <li>Low or zero processing fees</li>
```
luke-jr: Can't we just remove "or zero"? Although even now, my zero-fee transactions continue to work fine...

_templates/press.html
```diff
@@ -19,11 +19,9 @@
   quotejeremyliew: "The potential for disruption is enormous."
   quotewencescasares: "Right now Bitcoin feels like the Internet before the browser."
   quotetylerwinklevoss: "We have elected to put our money and faith in a mathematical framework that is free of politics and human error."
-  quotemaxkeiser: "It's the cheapest way to move money around."
```
luke-jr: Is something cheaper now?

_translations/en.yml
```diff
@@ -31,12 +31,8 @@ en:
     title: "Bitcoin for Businesses - Bitcoin"
     pagetitle: "Bitcoin for Businesses"
     summary: "Bitcoin is a very secure and inexpensive way to handle payments."
-    lowfee: "The lowest fees out there"
-    lowfeetext: "Bitcoin's high cryptographic security allows it to process transactions in a very efficient and inexpensive way. You can make and receive payments using the Bitcoin network with almost no fees. In most cases, fees are not strictly required but they are recommended for faster confirmation of your transaction."
```
luke-jr: Removing the last sentence should be sufficient?

_translations/en.yml
```diff
@@ -54,27 +50,21 @@ en:
...
   bitcoin-for-individuals:
     title: "Bitcoin for Individuals - Bitcoin"
     pagetitle: "Bitcoin for Individuals"
     summary: "Bitcoin is the simplest way to exchange money at very low cost."
     mobile: "Mobile payments made easy"
     mobiletext: "Bitcoin on mobiles allows you to pay with a simple two step scan-and-pay. No need to sign up, swipe your card, type a PIN, or sign anything. All you need to receive Bitcoin payments is to display the QR code in your Bitcoin wallet app and let your friend scan your mobile, or touch the two phones together (using NFC radio technology)."
-    international: "Fast international payments"
-    internationaltext: "Bitcoins can be transferred from Africa to Canada in 10 minutes. There is no bank to slow down the process, level outrageous fees, or freeze the transfer. You can pay your neighbors the same way as you can pay a member of your family in another country."
```
luke-jr: Would be okay if we just take out "10 minutes". Maybe "hours" if some time span is desired.

404.html
harding: "peer to peer" implied "from a stranger" to me because when I use other P2P protocols like BitTorrent, most of my peers are strangers.

We could rephrase to "Instant transactions from friends", but that sounds like weak marketing to me.  It implies (correctly) that transactions from strangers aren't instant, so it comes off as a negative point.

I think it's better to remove it and replace it entirely with a completely different advantage of Bitcoin.

404.html
harding: The motivation for this PR is people on the -dev mailing list saying that we have been misleading people into thinking that the current state of zero-fee and low-fee transactions will continue in perpetuity.

I think that's what this statement does: it makes people think that low fees are a fundamental part of the system, not just the current state.  (This same feedback also applies to two of your other diff comments: [1](https://github.com/bitcoin-dot-org/bitcoin.org/pull/972#discussion_r35388090) and [2](https://github.com/bitcoin-dot-org/bitcoin.org/pull/972#discussion_r35388128))

_translations/en.yml
harding: Agreed.  Thanks.

Dave-Aiello: Luke, you argued earlier that we should keep "instant peer-to-peer transactions" above because 0-conf is feasible among trusted friends / family.  The line in question here also implies sending to friends / family abroad, so we should actually change it from "in 10 minutes" to "instantly" here as well, rather than "hours".

harding: I'm just going to remove the time specification altogether.

404.html
luke-jr: "Instant transfers" + "Irreversible in hours instead of days"?

_translations/en.yml
```diff
@@ -68,13 +62,11 @@ en:
     mobile: "Mobile payments made easy"
     mobiletext: "Bitcoin on mobiles allows you to pay with a simple two step scan-and-pay. No need to sign up, swipe your card, type a PIN, or sign anything. All you need to receive Bitcoin payments is to display the QR code in your Bitcoin wallet app and let your friend scan your mobile, or touch the two phones together (using NFC radio technology)."
     international: "Fast international payments"
-    internationaltext: "Bitcoins can be transferred from Africa to Canada in 10 minutes. There is no bank to slow down the process, level outrageous fees, or freeze the transfer. You can pay your neighbors the same way as you can pay a member of your family in another country."
+    internationaltext: "Bitcoins can be transferred from Africa to Canada and there is no bank to slow down the process, level outrageous fees, or freeze the transfer. You can pay your neighbors the same way as you can pay a member of your family in another country."
```
Dave-Aiello: I'd suggest adding "quickly" either before or after "transferred" as a compromise.

404.html
petertodd: @luke-jr That kind of wording sounds reasonable.

Similarly, saying things like "payments in 10 minutes" is pretty close to the truth already - "irreversible payments in under an hour" for instance would be quite accurate.

petertodd:
I'm happy to ACK as-is.

Better wording for parts can be argued about later, where simple removals could be changed to replacements.

404.html
harding: I should note that this is the 404 page and that it redirects to the home page for anyone with Javascript enabled, so we probably don't need to spend much time optimizing it.  (Frankly, I think it should say "page not found" instead of being an old version of the index page.)

_translations/en.yml
pdaian: "transferred globally" also works.

felipelalli:
ACK

theymos:
It's long been known that Bitcoin itself would likely be incapable of handling zero-fee transactions and micropayments forever, but I don't think that it's accurate to say that low fees are just an accident of the current state of the network: Bitcoin does *make possible* technology like Lightning, which will likely allow all Bitcoin users to do cheap/fast transactions in the future. So while the removed text probably does promise too much, I don't think that references to *low* fees should be removed just because the technology for making this scale isn't quite there yet. And Bitcoin is way faster than other irreversible transaction systems in almost all cases.

saivann:
FWIW, I agree that replacing "no fee" with "low fee", and "instant" by "fast" is probably the best solution at this point than dropping complete texts altogether (Bitcoin still has great qualities). If users indeed have no choice but to pay higher fees in the future, these texts can be revisited to match reality.

Note, although maybe for a separate pull request, it may be worth clarifying that 3+ confirmations is now preferable for lightweight wallets on the "You need to know" page.

Ninkip2p:
@saivann FWIW, I agree that replacing "no fee" with "low fee", and "instant" by "fast" is probably the best solution at this point than dropping complete texts altogether (Bitcoin still has great qualities). If users indeed have no choice but to pay higher fees in the future, these texts can be revisited to match reality.

+1

_templates/press.html
thesoftwarejedi: Nothing is cheaper. 0 fee will go through with a possible delay.

thesoftwarejedi:
I disagree with @petertodd and the quick to ACK crowd. I bought coffee today instantly using bitcoin. That is a fact. The transactions **are** instant. Why change this wording at all?

Cash transactions are instant too, but the risk exists that the cashier will be robbed before they get that money into a bank. We don't therefore say that cash transactions aren't instant. 

As engineers, we have a tendency to try and be as literal and careful with this wording because with don't want to have to explain double spends and mining. We don't have to explain these things - they are robberies of the future and many companies are already in the business of addressing them, and educating merchants. 

Bitcoin DOES allow instant transactions, the only exception being the definition of "instant", for me it took about a second. Period.

carnesen:
> I suggest that replacement sections or more comprehensive revisions be submitted as a separate pull request.

The content that's there is all pretty good and accurate. Why is it better to cut it out wholesale like that and then have people rewrite it from scratch instead of just tweaking the existing wording?

Besides that Bitcoin does in fact offer low fees and instant transactions. So I largely disagree with these edits in either case.

_templates/press.html
mjamin: Possibly an impractically long delay.

harding:
@theymos and @saivann (quoting from Saïvann)

> If users indeed have no choice but to pay higher fees in the future, these texts can be revisited to match reality.

This site's marketing text is read by people who are deciding whether they want to start using Bitcoin.   If we tell them that fees are low without any additional qualifiers, they will assume that low fees are part of the system.  They may even assume, the way some people on Reddit seem to have assumed, that "low fees" mean "the current low fees" (which are extremely low, so that any fee rise looks big in comparison).

I worry that by advertising low fees, we're advertising something that isn't under our control, and which could make people feel like we misled them should fees rise significantly.  We're selling Bitcoin as a system, and I think we should be careful to only promise what we know we can deliver.

I think a similar situation applies to "fast".  Bitcoin is fast, but only when it comes to preventing chargebacks.  That's not typically a concern for consumers, and with Bitcoin and one-or-more confirmation requirements from merchants, Bitcoin can seem like a pretty slow payment method.  This is even more so if a user's transaction isn't confirmed in the next block or two, likely to become a more frequent occurrence if the queue of legitimate transactions grows.

In this case, I feel like we're advertising a best-case scenario that will disappoint the user when things don't go perfectly.

There are other good things about Bitcoin that we can reliably promise that are also appealing.  Perhaps more importantly, "low cost" and "fast" are not competitive with centralized payment systems that say "free" (for spenders) and which are so implicitly instant (sans chargebacks) that they don't even bother to mention it on their home page.  (I'm looking at the PayPal homepage as I write this.)

I'm happy to discuss this further, but I still feel like removing these texts now (and replacing them with something else later) is the best course of action.

kanzure:
> I bought coffee today instantly using bitcoin. That is a fact. The transactions are instant. Why change this wording at all?

Well, primarily because that's technically impossible. There are fundamental laws of physics that you'd have to break. You'll win a bunch of prizes in physics for doing so! You should feel honored or something? Word games aside, whether your merchant chooses to take zero-confirmation off-chain transactions is not something that can be enforced by the design of the system anyway.

404.html
Introshine: >Instant peer to peer transactions

I don't have the permissions to make edits, so I'm commenting here. I don't think this line is 100% false, it's just not good enough. 

I suggest:

>Directly verifiable digital cash payments - fast confirmation.

404.html
Introshine: >Low or zero processing fees

This is somewhat misleading, but there's a core of truth in it. I suggest changing it to:

>Low negational processing fees.

Because up to the latest block, fees are *low* (cents) but not "Free". Also, it's market driven.

_templates/press.html
Introshine: NACK 

It's a quote/citation, not written in stone. Bitcoin is one of the cheapest ways to move money around.

_translations/en.yml
Introshine: >The lowest fees out there
change to
>Negational, mostly low fees.

And the last sentence should stay there, with this adjusted

>......In most cases, fees are not strictly required but they are recommended for faster confirmation of your transaction..

change to

> .......In almost all cases, fees are required for a reasonable duration of confirmation of your transaction....

Introshine:
NACK

While most of the highlighted text is a bit misleading, I don't think they should be deleted. It would be better to re-write them to better match reality.

davout:
@thesoftwarejedi 

I bought coffee today using bitcoin and it wasn't instant. That is a fact. Thanks to this anecdotal evidence I can safely conclude that the transactions can not and are never instant.

saivann:
> If we tell them that fees are low without any additional qualifiers, they will assume that low fees are part of the system.

Re additional qualifiers: That's the reason why I suggest we use the "You need to know" page for that purpose (it is linked it prominently in all key pages), so the website can present current Bitcoin's qualities while also explaining it's risks and uncertainties. I wouldn't be against a short text here to summarize that fees are hardly predictable over the long term, and could rise significantly, either temporarily or permanently, depending on how the system scales up.

To quote a relevant text already on that page:

~~~
Bitcoin is an experimental new currency that is in active development.
Although it becomes less experimental as usage grows, you should
keep in mind that Bitcoin is a new invention that is exploring ideas that
have never been attempted before. As such, its future cannot be predicted by anyone.
~~~

thesoftwarejedi:
> whether your merchant chooses to take zero-confirmation off-chain transactions is not something that can be enforced by the design of the system anyway

You're right. So why not market through this site what a user will experience? Currently, the experience is instant, low cost transactions in a vast majority of cases. Why not just put an asterisk after "instant" and "low cost" and link out to the technical details? Currently the market meets my families expectations, and I have demanding teens buying a lot of steam games, Amazon shipments, Microsoft Xbox games, etc. they have not once complained about speed or cost. 

My point that I'm trying to make is that somehow we are feeling ethically responsible to disclose risk. Risks that exist with all currency. PayPal says instant. It is most of the time, until they freeze your bank accounts. They don't advertise those little nuances of the system because they're a PITA to explain.

felipelalli:
Less is more. The text promises too much.

pbreit:
I think low/no-cost and fast should continue to be highlighted as Bitcoin transaction advantages.

wamatt:
@Introshine "negational"?

that does not seem to fit the context. Perhaps you were after a different word?

harding:
@saivann I like the idea of adding a sentence about fees to the Experimental section on the You Need To Know page.  I'll see if I can think of something now and put it in a separate PR.  I'll also add the note about SPV wallets requiring extra confirmations at the same time.  Thanks!

But I don't feel like that ameliorates my concern that we're promising something that we can't be sure we can deliver.  Even if we assume everyone reads the You Need To Know page (unlikely), we're just moving the time when the user feels like they've been misled from months or years in the future to just a few minutes after they read some other page on the site that promises low fees and fast payments.

Admittedly, sooner is better for the user in the long run---but wouldn't it be better still to only make claims we can be highly confident about?  If low fees and fast payments are the only reason people would consider using Bitcoin, we're in trouble---centralized payment systems already provide consumers with zero spend fees and millisecond payment processing (plus lots of other features Bitcoin doesn't currently do well).

theymos:
@harding IMO if the fees for an end-user are especially high, then there's something *wrong* with Bitcoin, and that needs to be fixed somehow (probably via higher-layer protocols) as long as the fix doesn't compromise more important things like decentralization. Low fees aren't a guaranteed property of Bitcoin, but they're an important goal. This is true of many aspects of Bitcoin; for example, censorship-resistance and the ability to pay people worldwide aren't guaranteed properties of Bitcoin, but they *should* exist.

Instant transactions are possible if the recipient is willing to accept the fraud risk, which they often are. It's similar to credit card fraud risk, but disappears much more quickly.

I think that it's fair to advertise positive points about Bitcoin (the currency, not the base protocol) that exist now and that we can reasonably expect to still exist in 5-10 years, even if there might be some rough patches along the way.

I agree with this PR -- most of the removed stuff is awfully inaccurate/misleading -- but I think that a lot of this should be reintroduced with more careful language. For example, maybe replace the three points with something like "Fast (sometimes instant) peer-to-peer transactions | Worldwide payments | Be your own bank"

harding:
@theymos 

> I agree with this PR -- most of the removed stuff is awfully inaccurate/misleading -- but I think that a lot of this should be reintroduced with more careful language.

Are you requesting that be done in this PR, or are you fine with my proposal of submitting replacement marketing in a separate PR?

pbreit:
@harding one massive use case is cross-border transfers in which case speed and cost are two of the most important attributes. I don't see why you'd want to down-play or hide that.

theymos:
@harding A separate PR is fine with me.

dgenr8:
NACK

The changes in Bitcoin's usability are under active debate.  Changing the messaging at this point strays deep into the territory of taking a position in the debate.

It may be uncomfortable for some that the future they envision is far different than what is described -- and has been described for years -- on bitcoin.org.  That is not a reason to change the site now.

paulmadore:
FWIW, I have to agree that unless or until "free transactions" become the UX, it's not good to have information on the official page that insinuate they are what's going to happen. Everyone who's ever sent money another way is used to the idea of paying a fee. What if new materials are developed that illustrate this concept: "Competitive Fee Structure." 

Barring that, yes, this is a good idea. WYSIWYG is what we're going for, not false promises, right?

_templates/press.html
flound1129: "Is something cheaper now"?

Other than every alt?

davout:
> The changes in Bitcoin's usability are under active debate.

Not really.

thesoftwarejedi:
NACK For the reasons outlined above and need addressed. 

Complete removal of some **real** and core benefits of Bitcoin, while under pressure for block size attention, presents a very difficult to change stance on Bitcoin from a consumer angle. This is **exactly** the same difficulty being addressed re: block size because of a similar "temporary" change made. Yes, I'm comparing this change to [this one](https://bitcointalk.org/index.php?topic=287.msg8810#msg8810). 

For anyone in this thread to say that "we'll just add it back in later with different wording" is ignorant of the impact of this change.

Introshine:
@wamatt sorry I meant negotiable. English is not my native tongue.

harding:
@dgenr8 your NACK seems to focus on how this PR impacts the block size debate and its participants.  I don't care about that.  I'm concerned with how potentially misleading information on the website impacts users.

Mirobit:
NACK 
Free transactions should be removed But i think low fees should stay. Even with a 100x increase of the minimum fee the fee would be 0.2$ which is still pretty low for international payments. Basically nothing if you send large amounts. That bitcoin has a very small fixed fee compared too percentual fees of other payment systems is a huge advantage and should be marketed.

_translations/en.yml
```diff
@@ -31,12 +31,8 @@ en:
...
-    lowfee: "The lowest fees out there"
-    lowfeetext: "Bitcoin's high cryptographic security allows it to process transactions in a very efficient and inexpensive way. You can make and receive payments using the Bitcoin network with almost no fees. In most cases, fees are not strictly required but they are recommended for faster confirmation of your transaction."
     fraud: "Protection against fraud"
     fraudtext: "Any business that accepts credit cards or PayPal knows the problem of payments that are later reversed. Chargeback frauds result in limited market reach and increased prices, which in turn penalizes customers. Bitcoin payments are irreversible and secure, meaning that the cost of fraud is no longer pushed onto the shoulders of the merchants."
-    international: "Fast international payments"
-    internationaltext: "Bitcoins can be transferred from Africa to Canada in 10 minutes. In fact, bitcoins never have any real physical location, so it is possible to transfer any amount anywhere with no limits, delays, or excessive fees. There are no intermediate banks to make you wait three business days."
```
jonasschnelli: Why removing this?
You still can do a transaction in ~10mins.
Id recommend to keep this in. But maybe remove `or excessive fees`

jonasschnelli:
ACK.
It generally reflects the way of how bitcoin works today. So we should have text on the webpage of how it WILL work or how it WAS working.

nit "Fast international payments" should stay.

chriswheeler:
I think this goes a bit too far - It removes of whole chunks of useful content and even unrelated content simply because of its proximity to the 'bad' content. I understand that the plan would be to use separate pull requests to re-write that content, but I don't see the urgency in removing it immediately.

Why not simply switch 'zero fee' with 'low fee' and 'instant' with 'fast' (as has been suggested by others) which would make the current text reflect the current way that bitcoin payments usually work, then use separate pull requests to re-write anything which needs rewriting.

This pull request seems to assume that the 1mb block size limit will remain in place for the foreseeable future, and I assume is in response to @jgarzik 's comments about not communicating the new fee policies to users.

What it's actually doing is confirming that there has been a shift in policy and making predictions about future fees and confirmation times based on that shift in policy, which I think it at little premature at best.

_translations/en.yml
harding: I felt it had to much that was problematic:

1. "10 minutes" (implies a first confirmation only takes 10 minutes, when they can take much longer)
2. "bitcoins never have any real physical location, so it is possible to transfer any amount anywhere" (I get what's being said, but if you think about it, it doesn't really make logical sense: if something doesn't have a physical location, how can you transfer it somewhere?)
2. "no limits, delays, or excessive fees."  (Bitcoin certainly has limits and delays, and some people might consider $0.25USD per transaction excessive.)

I think a ground-up rewrite in a separate PR would be better.  I do think that International payments is one of the areas where Bitcoin excels.

jonasschnelli: > "10 minutes" (implies a first confirmation only takes 10 minutes, when they can take much longer)

But you can? If you know the current feeRate you can ensure (mostly with a fee less the 1USD) to get confirmed within the next block/~10min (at a chance above 95%).


Maybe we should also view the text from a advertising perspective. Let's don't be to particular about technical details: it's not a tech only documentation/description.

harding:
@chriswheeler this PR has no special urgency, and I probably won't merge it until a PR (or PRs) with replacement content are ready for review.

The purpose of having separate PRs is so that the possible endless bike shedding over new phrasing is separated from the decision to stop promoting low fees and instant transactions without additional qualifiers.

Regarding the block size debate, I made this PR on my own without any consultation with other people.  (Actually, I was procrastinating on another task at the time.)  This PR does not represent a policy shift, unless you think that I have the solo ability to shift consensus policies as Bitcoin.org's technical writer.

chriswheeler:
@harding ah OK, merging it at the same time as other PRs which re-write the removed content makes sense, sorry if I jumped to the wrong conclusion there.

I wasn't trying to imply that you have the sole ability to shift consensus policies, but you do seem to be making predictions about future fees and confirmation times. I think the bitcoin.org site content should be kept up to date with the current state of the system, rather than trying to guess what may or may not happen in future.

_translations/en.yml
harding: Sure, but you've been in #bitcoin-dev the same as me when random strangers start complaining about those one or two blocks a week that take 50 minutes or more.  An unexpected 40 minute or longer wait is a bad UX, so we should probably either provide an accurate expectation or not mention 10 minutes at all.

I agree that this isn't documentation, but good marketing sells the product as accurately as possible so that the customer gets exactly what they expected.  Barring that, it's better to promise less and deliver more.

jonasschnelli: Agree.
While thinking about this text, it came to my mind that the text not only focuses on bitcoin-core where a user could ensure over the UI to get a ~10min confirmation.

harding:
@chriswheeler if we promise something good that doesn't materialize, people may justifiably feel that we misled them.  On the other hand, if we don't promise that good thing and it is available anyway, people are likely to be pleasantly surprised.  So there's a definite asymmetry here that leads me to want to only promote things that I can be reasonably sure will be available or whose promotion contains sufficient qualifiers to ensure the visitor gets a fair impression.

chriswheeler:
@harding but your not promising something good that doesn't materialise, your describing the system in its current state. If that changes in future then update the site to reflect that. It's not like the site currently says "zero fees forever". 

It may be better to just add an asterisk with a reference to https://bitcoin.org/en/faq#how-much-will-the-transaction-fee-be - and expand that to explain that at some point in future fees may rise as the block reward decreases for miners, or due to other resource constraints.

If you're going down the route of only mentioning anything we can be sure will be available in future, you should also remove these sections:

image

Or, taking it to the extreme, rewrite bitcoin.org as a B2B focused website explaining how institutions can use Bitcoin as a settlement layer for their off-chain payment systems.

greenaddress:
I agree with the sentiment behind avoiding to mislead users. ACK.

@chriswheeler

> This pull request seems to assume that the 1mb block size limit will remain in place for the foreseeable future, and I assume is in response to @jgarzik 's comments about not communicating the new fee policies to users.

Just because bitcoin.org contains/contained some relatively misleading statements doesn't mean there has been any change in policy and doesn't mean that we should keep misleading statements forever. 

When statements are not clear or even outright misleading it is more than reasonable to fix them and I applaud the effort.

The fact that thus far there is no consensus on any of the proposed changes on the block size limit  should be enough to reasonably assume the 1MB limit will remain in place for the foreseeable future - this it is very reasonable to assume until consensus is reached on a different solution.

chriswheeler:
> When statements are not clear or even outright misleading it is more than reasonable to fix them and I applaud the effort.

I agree, but I believe the best fix in this case is s/zero/low/ & s/instant/fast/ rather than rm -f. Even that is being very cautions as it is perfectly possible to have zero-fee transactions at the moment, and instant (as in coffee) payments are also fairly common (especially when compared to card payments which are 'instant' as far as the consumer is concerned, and only actually settled once per day).

ywecur:
I don't understand. Have we all just decided that bitcoins future is one where these features are not main functions? Does this have anything to do with the block size debate?

btcdrak:
I generally agree with the sentiment of this PR, we should not be falsely advertising however, some of the deletions are a bit much. For example while small denomination microtransactions might be becoming less feasible fee wise on-chain, bitcoin still provides these using micropayment channels, that technology already exists today and is implemented and will continue to get better as OP_CHECKLOCKTIMEVERIFY is deployed along with OP_CHECKSEQUENCEVERIFY which together will empower Lightning payments. Even some argue LN is far away, micropayment are here today and there are already several implementations, Strawpay, Streamium, BlockCypher to name a few.

paulperegud:
NACK

Bitcoin transactions are the fastest out there. They are as close to instant and irreversible as it gets. There were no major cases of double spends AFAIK. Removing major selling point from the site just because of theoretical threat - this feels like pedantic zealotry.

ywecur:
@paulperegud Agreed

JRogiest:
@paulperegud <https://github.com/paulperegud> Agreed

czr5014:
i wish bitcoin stayed awesome, it is slowly losing its awesomeness.

instagibbs:
Sorry you have to put up with all this noise.

ACK, but I do think that this is a bit overkill. It's fast relative to some things, slow relative to others. Obviously nuance is hard on a newbie website, but I'd rather it not promise to be a floor cleaner and ice cream topping when it isn't.

jayd3e:
NACK

I'm really not sure why this is even necessary.  Bitcoin's transactions are faster and cheaper than anything else out there.  "Instant" is a marketing term used by all of the major payment processors and banks.  Don't take it too literally, it just means something is so fast it's indistinguishable from instant.  I've never used a wallet that didn't mask the confirmation process with an immediate success message of some kind, so I would say that the transactions ARE in fact instant.  It's impossible for anything to take 0 seconds, so why would we go to the added measure of being so overly literal?

Bitcoin does have "low to zero fees".  My zero fee transactions are stilling going through today.  It's premature to take out mentions of "low" and "zero" fees.  The more troubling thing, is that we DO in fact want to market Bitcoin as such.  B/c that's the vision that so many of us have invested tons of our time and energy into promoting and maintaining this ecosystem.  The idea that Bitcoin will be a global, fast, and low/zero feed payment system.  If you take away two of those things, Bitcoin becomes a lot less interesting.  I realize that we're not removing that functionality from Bitcoin as a technology, but by removing it from the marketing documentation, we're in fact removing that functionality from public awareness.

This doesn't represent the vision of Bitcoin so many of us want to see succeed.

instagibbs:
@jayd3e You are projecting your own vision onto the project. This is fine by itself, but you are requiring bitcoin.org to "toe the line" on your beliefs, rather than be neutral. There is no replacement text suggesting "high" fees, or anything of the like.

What makes this text problematic is clearly shown by the e-mail list discussions involving Jeff Garzik. He finds it irresponsible to change the economic reality away from what many had been sold on without plenty of advanced warning. If we are not allowed to even remove text promising to future HOLDers what we can't promise, that's a catch-22 for anyone who doesn't share your vision.

There is room to discuss what to replace it with, but the text is clearly not neutral as it is.

jayd3e:
@instagibbs How am I projecting my vision?  Bitcoin right this instant can support low fees(only a few pennies), and we have companies that are doing zero-confirmation(indistinguishable from instant) transactions.  If anything changing this text would be disingenuous and would be representative of someone's future vision.  I would also argue that those two aspects of Bitcoin are the only two that the general public is really going to care about, as privacy and anonymity are something that most people will collectively yawn about.

Also what makes the text problematic?  Do we have droves of people coming to the mailing lists and forums asking for this change?  I can safely say that this has never been a concern before by a large percentage of the user base.  In my experience it's far better to fix problems that exist due to large amounts of user feedback, as opposed to creating problems to fix.

saivann:
@harding I think describing current qualities of the system is not the same as making promises for the future. Nothing is garanteed in life, in particular when speaking about fees and new technologies.

There is reasonable doubt about the validity of "free / instant transactions" claims. However, it's worth thinking about where is the line. If we really start thinking of possible outcomes, we'd easily end up not saying much about security, decentralization, privacy, worldwide availability, freedom of transaction, because all of those things can be threatened by events we can't predict or control.

To me the best option remains to avoid making predictions, and not be afraid of displaying Bitcoin's features and potential, to the extent that these claims are reasonable and accurate, while also be straightforward about risks and uncertainties (by introducing and linking to the "You need to know / Secure your wallet / Protect your privacy" pages prominently and contextually).

This way I think we get a balanced website that is also a good introduction for newcomers.

instagibbs:
@jayd3e 

>The more troubling thing, is that we DO in fact want to market Bitcoin as such. B/c that's the vision that so many of us have invested tons of our time and energy into promoting and maintaining this ecosystem. The idea that Bitcoin will be a global, fast, and low/zero feed payment system. If you take away two of those things, Bitcoin becomes a lot less interesting.  

This is the "vision" of which bitcoin.org really has no business perpetrating as it is not intrinsic to the system. Nothing about the system guarantees low fees. Certainly not free. And nearly 100% of developers have doubts it will be the case in as little as a year unless radical, controversial changes are made. 

If we leave this text in, 6 months later we could be inviting yet more bag-holders to feel like they've been cheated because they were promised this vision.  

That's all I have to say on this.

harding:
@saivann 

> To me the best option remains to avoid making predictions

That feels fundamentally opposed to the objective of this PR (as stated in the second paragraph of the OP text), so that feels like a NACK to me.  Can you confirm that your intention here is to NACK, and that I'm not just misreading a less serious objection?

saivann:
> If we leave this text in, 6 months later we could be inviting yet more bag-holders to feel like they've been cheated because they were promised this vision.

@instagibbs This kind of thing has been a recurrent concern of mine for years. I think we'd likely have to drop any text / page that describe anything appealing about Bitcoin in order to completely adress that issue, but this would be at the cost of making the website a much less useful resource for introducing newcomers IMO, but that is a possibility.

@harding I don't want to be too blocking here :) , especially if you really feel strongly about it. But I personally think most of these texts should be rewritten first and it'd be easier to ACK. But otherwise, it's not the end of the world and I agree that in the absence of a more accurate replacement, it's better to drop them for now. A simple 's/Instant/Fast' 's/Zero or low fee/Low fee' would be fine for the front page IMO.

jayd3e:
@instagibbs once again you're using a lot of remarks with the word "could" in them.  I would say it's in our best interest to avoid predicting the future and solving problems that exist today.

instagibbs:
@saivann Can you point to other parts that are expected to not be true anymore in the near-ish future, assuming the core Bitcoin security model isn't broken? 

I'd be fine with a re-write instead. But status quo is bad and misleading to newcomers.

feld:
Love the bikeshed, but may I suggest a fresh coat of mauve? It would really brighten up the neighborhood.

chriswheeler:
@instagibbs I pointed out a couple here https://github.com/bitcoin-dot-org/bitcoin.org/pull/972#issuecomment-124462105 and I'm sure there are more.

If no solution to scaling is agreed upon, or developed, we very quickly get to a point where there is fierce competition for the available block space and only the wealthiest of users can afford to use the system, so most of the current use cases are no longer viable.

I'm sure that is not going to actually happen, because I'm sure a solution will be found (be it block size increases or something else). I think the problem is the the changes in the PR are based on the prediction that blocks will soon be full, and nothing will be done about it, so fees will increase.

saivann:
>I'd be fine with a re-write instead. But status quo is bad and misleading to newcomers.

Agreed

davout:
> I think the problem is the the changes in the PR are based on the prediction that blocks will soon be full, and nothing will be done about it, so fees will increase.

That's the case if nothing is done, which is the (sensible) default.

harding:
@saivann 

> I personally think most of these texts should be rewritten first and it'd be easier to ACK.

That seems to be a very popular opinion, so I'm going to close this PR in favor of making separate PRs with the suggested more surgical revisions.

Thank you everyone for your comments and suggestions.  There are some good ideas here, particularly in the diff comments, that I plan on using.

ABISprotocol:
Just saw this literally some five minutes ago and wish to add two points:

1) You need to add these pull requests open longer.
2) Apparently bitcoin developers were pushing for this as well which I think is problematic

I haven't read the reddit post(s) which developed on this, but apparently some cropped up on this matter, which I noticed because somebody mentioned it in the e-mail threads referring to this github issue.  

In sum, total NACK on this or on any pull request which has the intention of removal of mentions of "zero or low fees", "fast international payments", and "instant peer-to-peer transactions." 

Thank you.

saivann:
@ABISprotocol David closed his own pull request without merging anything, while preparing an alternative based on feedback he received. I think this can be criticized. If the other pull request is unsatisfactory, nothing prevents reopening this one. In other words, we still are in a WIP.

saivann:
Edit: I think this **cannot** be criticized, sorry

PR #973, opened 2015-07-24
Fixed broken url to script.h

jdb6167:
One line fix, but I noticed this URL was broken.


harding:
@jdb6167 good catch, thanks!  I'll merge from the command line in a moment.

PR #975, opened 2015-07-24
Combine /index, /<lang>/index, and /404

harding:
When working on #972, I discovered Bitcoin.org has three pages with nearly identical content:

- `/<language>/index.html` which provides the Bitcoin.org index page localized into each supported language
- `/index.html` which provides a fallback in case language redirection doesn't work.  This page is currently identical to the English localized index page.
- `/404.html` which provides the 404 error page, which includes a Javascript redirect to the localized index page plus fallback content that looks like an old version of the English localized index page.

This PR now builds all three pages from the same source so that if we change any text, we only have to change it once.  (No text changes are introduced in this PR, except that text displayed on the 404 page is now identical to the text displayed on the index pages.)

I tested this for English and Spanish.

### DIFF
--- a/404.html
+++ b/404.html
@@ -14,13 +14,4 @@ window.location.href='/'+lang+'/';
 </script>
 <!--Fallback page, don't allow indexing-->
 <meta name="robots" content="noindex">
-<h1>An open source P2P digital currency</h1>
-<p style="font-size:125%;">Bitcoin is a currency, a protocol, and a software that enables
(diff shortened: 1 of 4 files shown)


saivann:
That's indeed better, thanks!

harding:
@saivann thanks!

In the absence of critical feedback, this will be merged around 16:00 UTC Saturday.

PR #977, opened 2015-07-26
Build: Allow Rebuilding A Particular Commit

harding:
This PR updates the build script with two minor improvements:

1. When force building, the script no longer exits if you're attempting to the most recently-built commit.

2. The commit ID (commitish) of the most recently built commit is saved for use in other scripts

The intention here is to be able to create a cronjob for PR #976 that will force rebuild the site every night at midnight.  Point 1 makes that possible without the hard rebase hack @saivann and I currently use when manually rebuilding and point 2 prevents an automatic force build from allowing someone to bypass the signed commits by pushing a commit to master at exactly the right time.

I also switched the commit.txt command in the Makefile to use the same syntax as the build file, as I thought Saïvann's way was better.

This PR has not been tested, but I will test it when I manually update the file on the build server.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,7 +99,7 @@ build:
 	  | egrep -v 'sha256sums.txt' \
 	  | sort \
 	  | xargs -d '\n' sha256sum > _site/sha256sums.txt
-	$S git show --oneline > _site/commit.txt
+	$S git log -1 --format="%H" > _site/commit.txt
 
 ## Jekyll annoyingly returns success even when it emits errors and
 ## exceptions, so we'll grep its output for error strings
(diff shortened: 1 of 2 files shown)


saivann:
@harding As I understand it, the script would actually checkout to last commit from a repository provided as a command line argument (BUILD_COMMITISH). If so, what repository would you use here? I am not sure that I understand correctly what is in place to prevent unsigned commits to be built if this script is called with -force automatically.

Alternatively, maybe using a "-rebuild" option that does the same as "-auto", but skipping "$LASTLOCALCOMMIT == $LASTREMOTECOMMIT", would add less complexity and risk of messing up with "-force" (at the cost of always needing last commit to be signed).

harding:
@saivann the command line argument can be a repository or any other [commitish](http://stackoverflow.com/questions/23303549/what-are-commit-ish-and-tree-ish-in-git/23303550#23303550), but it's intended to be a particular commit.  For example, if I wanted to rebuild the site right now, I would run:

    update_site.sh force 09acaafbb0f265934e45f45acf99eb9e2570b787

The cronjob would be:

    update_site.sh force $( cat /bitcoin.org/last-successful-build-commitish )

Since that file is only updated after a successful build, that previous build would've had to have either been signed (verified during that build by `auto`) or manually `force` built by either you or me.

PR #981, opened 2015-07-26
Use "Low fees" and "Fast transactions"

saivann:
Live preview: 
http://bitcointest2.us.to/en/
http://bitcointest2.us.to/en/bitcoin-for-individuals#lowfee
http://bitcointest2.us.to/en/bitcoin-for-businesses#lowfee
http://bitcointest2.us.to/en/faq#what-are-the-advantages-of-bitcoin

This pull request isn't a complete solution to the problem outlined in #972 and is only meant to apply small changes everyone seem to consider "better than what we have now", so we can at least move on that front while more comprehensive changes are prepared. Several texts (in particular those about low fees) are still not appropriate and need to be fixed or replaced.

Given the previous strong support for this change, in the absence of critical feedback (feedback is still welcome), this pull request will be merged on July 29th.


harding:
ACK.  I think this is an improvement over the existing text, although I continue to believe that the revised claims are too bold and are setting users up to be disappointed.

1Hyena:
Deviating from the meaning of Bitcoin there, eh? :joy: 
At least in Bitcoin Cash it still applies that the transaction fees are near zero and they are fast.
You should also have replaced all references to Bitcoin with BlockstreamCoin. Because the Bitcoin I signed up for was never supposed to have high fees and slow confirmation times.

PR #982, opened 2015-07-27
Add Inside Bitcoins San Diego to event page

andreadezsi:
[Inside Bitcoins San Diego](http://insidebitcoins.com/san-diego/2015)


harding:
@andreadezsi thanks for submitting your event!  It should appear on the website within the next 15 minutes.

PR #983, opened 2015-07-28
Update settxfee description to match actual behavior

casey:
Currently, Bitcoin Core accepts 0 as an argument to settxfee, see settxfee() in https://github.com/bitcoin/bitcoin/blob/master/src/wallet/rpcwallet.cpp

I've updated the description of settxfee to reflect this.


harding:
Verified, looks like it was changed in https://github.com/bitcoin/bitcoin/commit/7d8ffac18612fe9a5faea3f8fb41442f9e7a3c1d

@casey thanks for submitting this fix! I'll merge it momentarily.

PR #987, opened 2015-07-29
Make no address reuse with change addresses a requirement for wallets

saivann:
This change was planned since a while and nearly all wallets avoid reusing change addresses by default now, including recently added wallets, so now seems like a good time to raise the bar.

The only exception is Hive for Mac, which might support it later. The Hive team is mostly focusing on Hive Web, which remains displayed on the website.


luke-jr:
ACK

crwatkins:
Good for me.

saivann:
In the absence of critical feedback, this pull request will be merged on July 31th.

harding:
ACK

PR #989, opened 2015-07-30
Update text on FAQ & Innovation pages

harding:
This PR continues the work started in aborted pull #972.  It is split into two commits:

1. No text changes; just wrap existing text to allow word diffing the next commit.

2. Actual text changes.

## Advantages of Bitcoin from FAQ

screenshot-btcorg localhost 2015-07-29 21-02-34

## Confirmations and fees from FAQ

screenshot-btcorg localhost 2015-07-29 21-19-23

## Micro payments from Innovation

screenshot-btcorg localhost 2015-07-29 21-22-53

### DIFF
--- a/_translations/en.yml
+++ b/_translations/en.yml
@@ -407,16 +407,19 @@ en:
 
     fee: "How much will the transaction fee be?"
     feetxt1: >
-      Transactions can be processed without fees, but users are
-      encouraged to pay a voluntary fee for faster confirmation of
-      their transactions and to remunerate miners. When fees are
-      required, they currently don't exceed a few pennies in value,
-      but this may increase over time. Your
-      Bitcoin client will pay what it thinks is an appropriate fee.
+      Transactions can be processed without fees, but trying to send
+      free transactions can require waiting days or weeks. Although fees
+      may increase over time, normal fees currently only cost a tiny
+      amount. By default, all <a href="#choose-your-wallet#">Bitcoin
+      wallets</a> listed on Bitcoin.org add what they think is an
+      appropriate fee to your transactions; most of those wallets will
+      also give you chance to review the fee before sending the
+      transaction.
 
     feetxt2: >
       Transaction fees are used as a protection against users sending
-      transactions to overload the network. The precise manner in which
+      transactions to overload the network and as a way to pay
+      miners for their work helping to secure the network. The precise manner in which
       fees work is still being developed and will change over time.
       Because the fee is not related to the amount of bitcoins being
       sent, it may seem extremely low or unfairly high. Instead, the
--- a/_translations/en.yml
+++ b/_translations/en.yml
@@ -297,16 +297,18 @@ en:
     advantages: "What are the advantages of Bitcoin?"
     advantagesli1: >
       <em><b>Payment freedom</b></em> - It is possible to send and
-      receive any amount of money instantly anywhere in the world at any
-      time. No bank holidays. No borders. No imposed limits. Bitcoin
+      receive bitcoins anywhere in the world at any
+      time. No bank holidays. No borders. No bureaucracy. Bitcoin
       allows its users to be in full control of their money.
 
     advantagesli2: >
-      <em><b>Very low fees</b></em> - Bitcoin payments are currently
-      processed with either no fees or extremely small fees. Users may
-      include fees with transactions to receive priority processing,
-      which results in faster confirmation of transactions by the
-      network. Additionally, merchant processors exist to assist
+      <em><b>Choose your own fees</b></em> - There is no fee to receive
+      bitcoins, and many wallets let you control how large a fee to pay
+      when spending. Higher fees can encourage faster <a
+      href="#you-need-to-know##instant">confirmation</a> of your
+      transactions. Fees are unrelated to the amount transferred, so
+      it's possible to send 100,000 bitcoins for the same fee it costs
+      to send 1 bitcoin. Additionally, merchant processors exist to assist
       merchants in processing transactions, converting bitcoins to fiat
       currency and depositing funds directly into merchants' bank
       accounts daily. As these services are based on Bitcoin, they can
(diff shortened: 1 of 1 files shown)


_translations/en.yml
```diff
@@ -367,11 +385,47 @@ en:
...
-    feetxt2: "Transaction fees are used as a protection against users sending transactions to overload the network. The precise manner in which fees work is still being developed and will change over time. Because the fee is not related to the amount of bitcoins being sent, it may seem extremely low (0.0005 BTC for a 1,000 BTC transfer) or unfairly high (0.004 BTC for a 0.02 BTC payment). The fee is defined by attributes such as data in transaction and transaction recurrence. For example, if you are receiving a large number of tiny amounts, then fees for sending will be higher. Such payments are comparable to paying a restaurant bill using only pennies. Spending small fractions of your bitcoins rapidly may also require a fee. If your activity follows the pattern of conventional transactions, the fees should remain very low."
+    feetxt1: >
+      Transactions can be processed without fees, but users are
+      encouraged to pay a voluntary fee for faster confirmation of
+      their transactions and to remunerate miners. When fees are
```
saivann: @harding How about:

~~~
Transactions can be processed without fees, but usually will require them for faster and
more reliable confirmation of your transactions and to remunerate miners.
~~~

(I feel current text fails to mention that user experience and security that comes with fast confirmations usually doesn't come without fees, so fees aren't only encouraged as a benevolent action for the network)

_translations/en.yml
```diff
@@ -367,11 +385,47 @@ en:
...
+      required, they currently don't exceed a few pennies in value,
+      but this may increase over time. Your
+      Bitcoin client will pay what it thinks is an appropriate fee.
```
saivann: @harding How about:

~~~
Your Bitcoin client will display what it think is an appropriate fee.
~~~

I feel like the current text suggests wallet will pay whatever fee they please in the background without informing the user. In reality, wallets typically show the transaction fee somewhere in the UI before the user clicks "Send".

_translations/en.yml
```diff
@@ -367,11 +385,47 @@ en:
...
+      Transaction fees are used as a protection against users sending
+      transactions to overload the network. The precise manner in which
+      fees work is still being developed and will change over time.
+      Because the fee is not related to the amount of bitcoins being
+      sent, it may seem extremely low or unfairly high. Instead, the
+      fee is relative to the number of bytes in the transaction, so
+      using multisig or spending multiple previously-received amounts
+      may cost more than simpler transactions. If your activity follows
+      the pattern of conventional transactions, you won't have to pay
+      unusually high fees.
```
saivann: @harding I thought some previous analogies and examples were still useful, how about re-including them along with your other changes:

~~~
Transaction fees are used as a protection against users sending transactions to overload the network.
The precise manner in which fees work is still being developed and will change over time. Because
the fee is not related to the amount of bitcoins being sent, it may seem extremely low or unfairly
high. Instead, the fee is defined by attributes such as the number of bytes in the transaction and
transaction recurrence. For example, if you are receiving a large number of tiny amounts, then fees
for sending will be higher. Such payments are comparable to paying a restaurant bill using only
pennies. Spending small fractions of your bitcoins rapidly may also require a fee. If your activity
follows the pattern of conventional transactions, you shouldn't have to pay unusually high fees.
~~~

saivann:
Except for my comments above, LGTM up to 8fac92a6. Thanks!

_translations/en.yml
harding: > the fee is defined by attributes such as the number of bytes in the transaction and
transaction recurrence.

I think this is misleading.  Miners tend to use Bitcoin Core's fee/priority evaluation code, which I believe works like this:

- Fees are evaluated purely based on fee-per-byte.  The age and value of the input (priority) plays no role here.

- If the transaction doesn't qualify for inclusion in the block based on its fee, its priority is evaluated.  Here the fee plays no additional role.

In other words, fee and priority are considered separately, but this sentence makes me feel like they're averaged together to allow high-priority transactions to pay lower fees for the same processing speed as lower-priority transactions.

> Such payments are comparable to paying a restaurant bill using only pennies.

I see where this analogy was going, but it breaks down for me.  I have personal experience refusing a transaction paid in an excessive number of pennies from when I worked at McDonalds in high school, 17 years ago.  I think that would be the usual response to that situation; I don't know of any business besides a bank that would charge more to process loose change.

> Spending small fractions of your bitcoins rapidly may also require a fee.

I think this could confuse people about how small "small fractions" are or how rapid "rapidly" is.  According to my node's `estimatepriority`, getting a free transaction confirmed within the next 25 blocks requires 331411429.8671501 priority.  So to send a 250-byte transaction with an input worth 0.1 BTC (10,000,000 satoshis) requires that input be 8286 blocks old: `331411429.8671501 / 10000000 * 250`

I wouldn't describe $30 USD worth of bitcoin as small or two months worth of waiting as rapid, so I think this sentence should be removed and that we should generally treat free transactions as something for experts and curious people only.

> transaction recurrence.

In case you revise, I'd suggest using a different term.  Recurrence is meant to be applied to similar events separated by time.  I get that receiving one Bitcoin transaction seems similar to receiving another Bitcoin transaction, but people tend to think of transactions based on what was exchanged---not the medium the exchange occurred through---so selling a pocket knife for 0.1 BTC in one transaction and a flashlight for 0.2 BTC in a different transaction doesn't seem like recurrence.

---

My suggestion would be to use the text in the commit, which does include examples, but I'm happy to continue working on revisions of this part of the paragraph.  (I'll be pushing updated text in a moment based on the rest of your commit comments.)

harding:
@saivann made changes in 7b8e26f addressing all but one of your comments, and left a reply to the un-addressed comment.  Thanks!

saivann:
LGTM 7b8e26f, thanks!

harding:
@saivann thanks for your review!

In the absence of critical feedback, this pull will be merged on Tuesday.

PR #991, opened 2015-07-30
Update quotes on Press page

harding:
This PR makes changes to the quotes section of the Press page.  It is, I hope, the last PR related to the changes started in aborted PR #972.  There are several commits:

1. Fix a CSS-based bug that prevents display of the open-quote on the left column.  This can be seen on the live site: https://bitcoin.org/en/press#quotes

2. Add a link to the [Bitcoin WikiQuotes](https://en.wikiquote.org/wiki/Bitcoin) page for even more quotes.  Although this page doesn't currently have many more quotes than we do, I hope that by linking to it, we'll encourage more people to contribute to it.

3. Replace quotes from Max Keiser and Timothy Lee with quotes from Paul Buchheit (GMail creator) and Brian Forde (MIT DCI director).  The replaced quotes described Bitcoin as, respectively, "cheapest" and "costless", which is only true if you exclude certain other payment systems (like physical cash) and network costs (like mining and verification).

4. Link the names of the quoted persons to the English Wikipedia articles about them.  If no English Wikipedia article was available, I did not link anything.

5. Link quotes to the source where the quote first appeared (as best as I could determine).  I couldn't find the source for some quotes, so no link was added in those cases.  I also corrected one quote based on the linked source quote.

6. Remove the quote that both Bill Gates and Eric Schmidt said.  The quote on Bitcoin.org is, "Bitcoin is a technological tour de force", but I feel that this is misleading given the qualifications they each put into the rest of the original quotes (as linked in the preceding commit):

    - (Schmidt) "Bitcoin is a technological tour de force, but what I don't know ultimately is whether it's going to be legal because almost all money structures need legality..."

    - (Gates) "It's a techno tour de force, but that's an area where governments are going to maintain a dominant role."

    I think they're approving of the block chain/POW consensus but not of the monetary aspect.  Since Bitcoin.org is promoting both aspects, I don't think it's fair of us to use their quotes.

After these commits, the fully-expanded section looks like:

screenshot-btcorg localhost 2015-07-30 18-53-53

### DIFF
--- a/_templates/press.html
+++ b/_templates/press.html
@@ -13,6 +13,7 @@ dialogs:
   materialquote: "Quotes"
   materialquotemore: "Show more quotes..."
   material_even_more_quotes: "For even more quotes about Bitcoin, please see <a href=\"https://en.wikiquote.org/wiki/Bitcoin\">WikiQuote.</a>"
+  material_source: "source"
   quotesatoshi: "With e-currency based on cryptographic proof, without the need to trust a third party middleman, money can be secure and transactions effortless."
   quotetonygallippi: "Bitcoin is Money Over Internet Protocol."
   quotedankaminsky: "Entire classes of bugs are missing."
@@ -34,7 +35,7 @@ dialogs:
   quotetuurdemeester: "The Babyboomers gave us the personal computer, Generation X gave us the Internet, and now Generation Y is building a new financial paradigm."
   quoterichardbrown: "Had you asked me five years ago, I would just say it was impossible. Bitcoin and cryptocurrencies solved this problem of coming to a consensus globally where you don't trust anybody else."
   quotefredwilson: "We believe that Bitcoin represents something fundamental and powerful ... It reminds us of SMTP, HTTP, RSS, and BitTorrent in its architecture and openness."
-  quotejohndonohoe: "Digital currency is going to be a very powerful thing."
+  quotejohndonohoe: "[Digital currency is going to be] a very powerful thing."
   quotefrancoisrvelde: "It represents a remarkable conceptual and technical achievement, which may well be used by existing financial institutions (which could issue their own bitcoins) or even by governments themselves."
 
 ---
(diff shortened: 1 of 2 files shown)


luke-jr:
Soft-dislike of linking Wikipedia in a way that implies it is a reliable source of information.

Also, Gavin is no longer with the Bitcoin Foundation.

harding:
@luke-jr I know you don't like Wikipedia for that reason, and I agree that too many people blindly trust it.  However, I think it has the advantage of being a neutral starting resource for research into the backgrounds of public figures.

Are you sure Gavin is no longer with the Foundation?  Last I heard, he was still its Chief Scientist even though the MIT Media Lab pays his salary.  At least that's what [this tweet](https://twitter.com/gavinandresen/status/591624879876767744) from him said in April.

_templates/press.html
```diff
@@ -19,16 +21,20 @@
...
+    by strengthening the capacity of governments to deliver more
+    responsive services and secure the rights of their citizens to
+    property, identity and increased financial inclusion.
+
   quoteandyskelton: "PayPal alone blocks access from over 60 countries ... Whatever the reason, we don’t think an individual blogger from Haiti, Ethiopia, or Kenya should have diminished access to the blogosphere because of payment issues they can’t control."
   quotetuurdemeester: "The Babyboomers gave us the personal computer, Generation X gave us the Internet, and now Generation Y is building a new financial paradigm."
   quoterichardbrown: "Had you asked me five years ago, I would just say it was impossible. Bitcoin and cryptocurrencies solved this problem of coming to a consensus globally where you don't trust anybody else."
   quotefredwilson: "We believe that Bitcoin represents something fundamental and powerful ... It reminds us of SMTP, HTTP, RSS, and BitTorrent in its architecture and openness."
-  quotejohndonohoe: "Digital currency is going to be a very powerful thing."
+  quotejohndonohoe: "[Digital currency is going to be] a very powerful thing."
```
saivann: @harding Are the square brackets intentional here?

saivann:
Except for the comment above, LGTM up to commit 5705f0a. Thanks!

_templates/press.html
harding: Yes.  Here's the quote in context from the linked source page:

> eBay president and CEO John Donahoe said that the company was keeping its eye on Bitcoin and that in the months and years to come, digital currency is going to be “a very powerful thing”.

harding:
@saivann thanks!  In the absence of critical feedback, this pull will be merged on Tuesday.

PR #993, opened 2015-08-02
Update instructions to starting up regtest

Steve12code:
Using Regtest mode as documented fails, you must first define rpcuser and rpcpassword before executing the generate 101 command.

### DIFF
--- a/_includes/example_testing.md
+++ b/_includes/example_testing.md
@@ -57,6 +57,14 @@ environment.
 Bitcoin server starting
 {% endhighlight %}
 
+Create a file in the .bitcoin directory names bitcoin.conf. Populate with
+the following:
+
+~~~
+rpcuser=anyuser
+rpcpassword=anypswd
+~~~
+
 {% autocrossref %}
 
 Start `bitcoind` in regtest mode to create a private block chain.


Steve12code:
Following the developer example I was unable to get the regtest option working without first populating bitcoin.conf.   Altered doc to reflect this.

harding:
@Steve12code this is already document near the top of the page:

screenshot-bitcoin org 2015-08-01 21-44-03

Setting `rpcuser` shouldn't be necessary, so it isn't documented.

Would you consider it satisfactory for me to link to the above instructions from the testnet and regtest sections?

Steve12code:
@harding a link back to the beginning and having the document mention it up there could work.

I do still have an issue when I don't specify rpcuser.  I just took rpcuser out of the bitcoin.conf file and it has stopped working for me again. I am getting 
"error: incorrect rpcuser or rpcpassword (authorization failed)"
and have to add it back in.

If that is not how things are supposed to work perhaps there is a specific order the tasks have to be done?

harding:
@Steve12code if you start `bitcoind` with `rpcuser` set to a non-default value and then remove `rpcuser` from the config, it won't work.  You have to shutdown `bitcoind`, remove the `rpcuser` setting, and then restart `bitcoind`.  If that's what you did, can you let me know what version of Bitcoin Core you're using and what platform you're on?

I will add the links tomorrow.  Thanks.

Steve12code:
@harding 

Thanks, doing that enabled me to get it functioning without rpcuser.

May I also ask what is the supported way of stopping bitcoind?
When I issue 'bitcoind stop' it refers me to utilize bitcoin-cli.
When I issue 'bitcoin-cli stop' I get 'error: couldn't connect to server'

I ended up doing this 'killall -9 bitcoind', which worked but I'm sure is not ideal.

harding:
@Steve12code `killall -9 bitcoind` sends SIGKILL, which is a bad idea except if the daemon is not otherwise responding for a long time.  If Bitcoin Core is forcibly shutdown in the middle of certain operations, you will need to reindex the entire block chain the next time you start---which can take hours.

If you must use `kill` or `killall`, try using it without the `-9` to send SIGTERM.  Bitcoin Core will handle this gracefully and shutdown correctly within a few seconds (maybe up to 30 seconds on a computer with really slow file system syncs).

If you started `bitcoind` with the same rpcpassword and (optional) rpcuser as are currently used in the bitcoin.conf, then using `bitcoin-cli` as the same user should work.  As you said, the command `bitcoin-cli stop` should shutdown Bitcoin Core (again, this will take a few seconds).

I'm going to update the docs with the links in a few seconds and will close this pull when that's done.  The website will be updated within 15 minutes after that.  If you see any problems, you can continue commenting here.

PR #1006, opened 2015-08-10
Adding French communities links

ffmad:
added [French reddit](http://www.reddit.com/r/BitcoinFrance/), [facebook group](https://www.facebook.com/groups/BitcoinFrance/) and irc [#cryptofr](https://webchat.freenode.net/?channels=cryptofr) (way more active people there than in #bitcoin-fr)

would it be possible to add the major meetups too ?


harding:
I checked the subreddit and the IRC chatroom, and they both seemed reasonable to me.  The Facebook group required login and I don't have an account, but we already link to other Facebook groups so I guess it's ok.  A preview of the page with the new links highlighted can be found below.

In the absence of critical feedback, this will be merged on Saturday.

screenshot-btcorg localhost 2015-08-13 15-40-50

Issue #1014, opened 2015-08-20
125% magnification breaks front-page styling


carnesen:
The front page of bitcoin.org shows as the authors intended at "normal" magnification (100%) and also at 110%. But for some reason in going from 110% to 125% the content width increases dramatically (far more than the extra 15% would imply) and also the list "Fast peer-to-peer transactions ... Worldwide payments" goes vertical. Screen shot attached.
<img width="1428" alt="screen shot 2015-08-20 at 09 38 58" src="https://cloud.githubusercontent.com/assets/5756743/9386098/9836a250-471f-11e5-8e0f-b7d4cfad80a6.png">

harding:
On Firefox 31.8 with a 1440px window, scaling steps appear to be proportional.  What @carnesen seems to be observing is the switch from the desktop layout to the mobile layout that occurs when the div.body box overflows the html body box.  I'm not sure this is an actual bug, but I'll tag it as such until @saivann has a chance to investigate.

saivann:
@carnesen I have opened pull request #1016 to fix one of the issues your seeing with the icons being suddenly left-aligned and not inline-displayed, please feel free to try the live preview and comment there. Otherwise what you are describing seems to be the intended behavior (switching to the mobile layout). Please leave a comment if you can think of any usability issue this can cause.

I am not aware of any way to allow desktop browsers to zoom on the desktop layout without triggering the mobile layout as it reduces the available width on the screen, and that is how mobile layouts enter into play thanks to @media CSS rules.

PR #1016, opened 2015-08-20
Apply more graceful mobile stylesheet to landing page icons

saivann:
This addresses at least one issue mentioned on #1014 , by allowing these elements to be displayed inline on the screen until the space is really stretch enough.

I should have left equivalent space for texts, so variable-width translations should still fit correctly in the available space. At least I have verified various ones without noticing any issue.

### DIFF
--- a/_less/ie.css
+++ b/_less/ie.css
@@ -139,30 +139,22 @@ body{
 .mainlist div{
 	zoom:1;
 	display:inline;
-	width:250px;
+	width:240px;
+	text-align:center;
 }
 .mainlist div div{
 	zoom:1;
 	display:inline;
 	line-height:1.5em;
-	text-align:left;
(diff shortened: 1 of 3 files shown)


harding:
LGTM.  Thanks!

saivann:
In the absence of critical feedback, this pull request will be merged on August 21th (tomorrow).

carnesen:
Yup, good one! Preview looks good now even at 175%.

saivann:
@carnesen Thanks for reporting the issue!

saivann:
Note; I just added some changes to the CSS for the RTL and IE6 layouts to provide them with the same display, the live preview will be updated in a minute.

harding:
utACK af646e4

PR #1017, opened 2015-08-20
Scaling Bitcoin Conference

sunnankar:


harding:
Thank you for submitting this important event!  It should be listed on the website within the next 15 minutes.

sunnankar:
Not just important but extremely important. Thanks for getting it merged quickly.

PR #1018, opened 2015-08-21
Add JSHint checks to the Makefile

saivann:
Live preview: http://bitcointest2.us.to/en/

If we're to add more javascript code to the website, this can help encourage a strict syntax and catch mistakes (I actually found and fixed one bug while preparing this pull request).

JSHint is configured to check only custom javascript files in js/ (not libraries) and uses the default configuration with a few relaxed tests.


harding:
Tested ACK: built & tested locally; checked HTML diff; browsed several update pages & checked JS functions.  Also reviewed code.

Note that this PR will create a merge conflict with #1009.  I suggest you merge this first and I'll rebase on top of it.

When you merge, the bundle on the build server should be automatically updated (it worked the last two times the Gemfile was changed) but I suggest that you monitor it just in case.

Finally, thanks!  This will be really useful particularly for me, since I write JS so infrequently.

saivann:
@harding Wow, thanks for your fast testing! Indeed, I planned to watch the live update closely when I will merge this pull request.

In the absence of critical feedback, this pull request will be merged on August 22th.

PR #1019, opened 2015-08-21
Allow code samples to overflow on mobiles in the documentation

saivann:

Some code samples cannot be read on mobiles. Although this solution isn't ideal as it makes the code overflow outside of the page, my impression is that it's better than cutting the content altogether.

### DIFF
--- a/_less/screen.less
+++ b/_less/screen.less
@@ -1003,7 +1003,7 @@ div.post {
         padding-top: 20px;
 }
 
-.highlight  { background: #ffffff; }
+.highlight  { background: transparent; }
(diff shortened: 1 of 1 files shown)


harding:
Tested ACK using Android 2.3.5 for mobile and FF 31.8 for desktop.  Overflowing on mobile is indeed much better, thanks!

saivann:
@harding Thanks!

In the absence of critical feedback, this pull request will be merged on August 23th.

carnesen:
Tested on Chrome on iOS and desktop. Is better indeed, thanks.

PR #1020, opened 2015-08-22
Allow build script to reuse pre-installed dependencies

saivann:
Currently, our build script needs to download and install dependencies on every build. This probably prolongs build time unnecessarily and introduces a bit more of security risk with no apparent benefit.

This little change allows the build script to stop deleting files listed in .gitignore when fetching recent commits, therefore allowing it to keep and reuse the bundle environment in ./vendor.

Note: This small change is untested.

### DIFF
--- a/_build/update_site.sh
+++ b/_build/update_site.sh
@@ -39,7 +39,7 @@ fi
 
 # Update local branch
 git reset --hard origin/master
-git clean -x -f -d
+git clean -f -d
 
 ## Whether to auto-build or force-build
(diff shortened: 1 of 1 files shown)


harding:
@saivann I thought we fixed that in #939 ?

saivann:
@harding Duh, obviously! Thanks, I somehow assumed that the build was still using the ./vendor path, and I guessed that I had only thought about the issue without opening a PR in the past... I should have looked more.

PR #1027, opened 2015-08-25
Update Inside Bitcoins Seoul date

muresanroland:


harding:
@muresanroland thanks!  The updated date should appear on the website within 15 minutes.

PR #1028, opened 2015-08-25
Remove Copay, Circle, Coinbase, BitGo, etc. for BIP101

gladoscc:
Many bitcoin services and industry stakeholders have recently came out in support of BIP101, [including BitPay, Blockchain.info, CIrcle, Kncminer, itBit, Bitnet, and Xapo](http://blog.blockchain.com/wp-content/uploads/2015/08/Industry-Block-Size-letter-All-Signed.pdf). They have pledged to "run code that supports this" by December 2015. This is in addition to ad-hoc support from [Coinbase](https://twitter.com/coinbase/status/595741967759335426) and [Erik Voorhees](https://www.reddit.com/r/Bitcoin/comments/3h5556/xt_is_not_only_on_topic_it_is_the_most_important/) (Coinapult). As per bitcoin.org's policy on ["contentious hard forks"](https://bitcoin.org/en/posts/hard-fork-policy), I have created a pull request to remove all of these services from the wallet page.

The bitcoin.org policy is states that it will not support wallets or services that "leave the previous consensus because of an intentional and contentious hard fork attempt." As the support for BIP101 from these players are indisputably in response to Bitcoin XT's hard fork, I believe there is no choice but to remove all of these wallets from bitcoin.org. We must prevent people from joining XT or supporting BIP101.

-TradeFortress

### DIFF
--- a/_templates/choose-your-wallet.html
+++ b/_templates/choose-your-wallet.html
@@ -191,51 +191,6 @@ wallets:
           privacyaddressreuse: "checkpassprivacyaddressrotation"
           privacydisclosure: "checkfailprivacydisclosurespv"
           privacynetwork: "checkfailprivacynetworknosupporttor"
-- copay:
-    title: "Copay"
-    titleshort: "Copay"
-    compat: "mobile desktop web android ios windowsphone windows mac linux"
-    level: 3
-    platform:
-      mobile:
-        text: "walletcopay"
-        link: "https://copay.io"
-        source: "https://github.com/bitpay/copay"
-        screenshot: "copay.png"
-        os:
-        - android
-        - ios
-        - windowsphone
-        - linux
-        check:
-          control: "checkgoodcontrolfull"
-          validation: "checkfailvalidationcentralized"
-          transparency: "checkpasstransparencyopensource"
-          environment: "checkpassenvironmentmobile"
-          privacy: "checkpassprivacybasic"
-        privacycheck:
-          privacyaddressreuse: "checkpassprivacyaddressrotation"
-          privacydisclosure: "checkfailprivacydisclosurecentralized"
-          privacynetwork: "checkfailprivacynetworknosupporttor"
-      desktop:
-        text: "walletcopay"
-        link: "https://copay.io/"
-        source: "https://github.com/bitpay/copay"
-        screenshot: "copay.png"
-        os:
-        - windows
-        - mac
-        - linux
-        check:
-          control: "checkgoodcontrolfull"
-          validation: "checkfailvalidationcentralized"
-          transparency: "checkpasstransparencyopensource"
-          environment: "checkfailenvironmentdesktop"
-          privacy: "checkpassprivacybasic"
-        privacycheck:
-          privacyaddressreuse: "checkpassprivacyaddressrotation"
-          privacydisclosure: "checkfailprivacydisclosurecentralized"
-          privacynetwork: "checkfailprivacynetworknosupporttor"
 - airbitzwallet:
     title: "Airbitz Bitcoin Wallet"
     titleshort: "Airbitz"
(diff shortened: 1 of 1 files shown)


harding:
Thank you for opening this pull request.  The Bitcoin.org site owners and maintainers discussed this yesterday by email, and several of us found that the letter you linked to was too ambiguous about the co-signer's planned actions.  Because of that ambiguity, I would prefer to wait for clearer statements of intent before we consider removal.

carnesen:
To be sure, there's no public indication that Coinbase supports BIP101. They didn't sign the letter. [The tweet referenced by OP](https://twitter.com/coinbase/status/595741967759335426) is from early May, weeks before BIP101 existed. More recently their director of engineering wrote an [impassioned plea for patience, condemning BitcoinXT as "dangerous and irresponsible"](https://www.reddit.com/r/Bitcoin/comments/3hp190/charlie_lee_nuclear_option_of_forking_the/cu9e4tj).

gladoscc:
Thanks for the info @harding

luke-jr:
@harding There is nothing ambiguous about the letter. They could not have been more explicit about switching to XT without specifying it by name (which would have worked against their intended goal, by making it seem distinct from Bitcoin).

However, these companies potentially represent an economic majority, and can very well force the economic consensus to recognise XT as Bitcoin. So it is not clear that the policy should apply anymore.

ABISprotocol:
Hello,

Full disclosure, I've been very vocal in opposition to XT for many reasons not worth repeating, and I've suggested support in the past for both Cameron Garnham's dynamic block size adjustment proposal (which would have been a soft-fork and may be implemented without changing any client code whatsoever) and more recently I've emphasized BIP 100 as a superior alternative to the variety of alternatives - BIP 100, 101, 102, SIPA, etc., [some of which can be seen here](http://bipsxdevs.azurewebsites.net/).  Actually, not "recently," but rather, [since June](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2015-June/008771.html) if not before, I've been hinting that BIP 100 is a good solution (along with my dire warnings about why one should not use XT).

With that fully disclosed, and my bias completely out there for the world to see, I want to add the following notes to this discussion:

There are two companies that are doing all of the following three things:
Backing the Windhover ID system ([Windhover press release](http://www.marketwired.com/press-release/21-top-bitcoin-digital-currency-companies-endorse-new-digital-framework-digital-identity-1959159.htm) | [Windhover regulated identity](https://idcubed.org/home_page_feature/privacy-and-compliance-by-design/) 
Helped fund Coin Center (which is pushing for a bitlicense in California at this very moment)
...and have signaled that they will adopt XT. 
These two companies are BitPay and Xapo.

These actions in combination to me represent a corporate assault on bitcoin users.  I believe rather than solely looking at one action (e.g. the company backing Windhover or indicating support for XT) one must look at a couple at least in combination to assess clearly the direction of the companies in question.  In my humble opinion, these companies are not acting in the best interests of the users.  I have called for a boycott of these companies and I would see no problem with them being removed from the wallet page.

Regarding the other companies I think you have to take a closer look on a more individual basis.  The remainder essentially can be evaluated as follows:
BitGo, BitPay, coinbase, itBit, and xapo all have funded the Coin Center, which as I've mentioned has taken a position in favor of a bitlicense bill in California.  ([Fortunately the [Electronic Frontier Foundation has helped to oppose this](https://www.eff.org/deeplinks/2015/08/license-kill-innovation-why-ab-1326-californias-bitcoin-license-bad-business), but it increasingly looks like if we can't get the bill killed in committee then it will come down to asking Governor Brown to veto it, which is a sorry state of affairs.)
Have these companies also all supported XT?  As mentioned, BitPay and xapo have come out in support of XT, for example.  However, it's less clear about Coinbase, [whose director of engineering has publicly opposed XT](https://www.reddit.com/r/Bitcoin/comments/3hp190/charlie_lee_nuclear_option_of_forking_the/cu9e4tj).
To be clear, I have strong objections to Coinbase (and long ago cancelled my account with them), in part because of how they have participated in financial censorship at the behest of banks.  However, this pull request seems to orient around whether some wallet company has supported XT, and it's not clear to me that Coinbase does, as @carnesen indicated above.

To me, it seems like it is in order to ask BitGo, Coinbase, and ItBit whether or not they plan to in fact use XT. (It seems obvious to me that BitPay and Xapo will.)  If they refuse to answer in a reasonable timeframe I would take it as a yes.  I would also recommend that you hold this pull request open for a good long while and see if they comment on it.

PR #1046, opened 2015-09-03
added instructions for setting up bitcoind on OS X

kevcooper:
The launch agent part of this guide is still waiting on [this PR](https://github.com/bitcoin/bitcoin/pull/6621)

### DIFF
--- a/en/full-node.md
+++ b/en/full-node.md
@@ -1107,10 +1107,47 @@ automatically started minimized in the task bar.
 #### Bitcoin Core Daemon {#osx-daemon}
 {:.no_toc}
 
-If you can provide instructions and screenshots for running the latest
-version of Bitcoin Core daemon on OS X Yosemite, please [open an
-issue](https://github.com/bitcoin-dot-org/bitcoin.org/issues/new) and we'll tell
-you what we need.
+The Bitcoin Core daemon (bitcoind) is not included in the .dmg file you may have downloaded to install Bitcoin-QT. Bitcoind, along with its support binaries, is instead included in the OS X .tar.gz file listed on the official Bitcoin Core download page. To download this file using Terminal, execute the following command:
+
+    curl -O https://bitcoin.org/bin/bitcoin-core-0.11.0/bitcoin-0.11.0-osx64.tar.gz
+
+{{verifyReleaseSignatures}}
+
+Extract bitcoind and its support binaries from the archive we just downloaded by running this command in Terminal:
+
+    tar -zxvf bitcoin-0.11.0-osx64.tar.gz
+
+Now we'll move the executables into your default path to make running and stopping bitcoind easier. To move the executables, run these commands (note that we have to use `sudo` to perform these commands since we are modifying directories owned by root):
+
+    sudo mkdir -p /usr/local/bin
+    sudo cp bitcoin-0.11.0/bin/bitcoin* /usr/local/bin/.
+
+To clean up the directory we've been working in, run:
+
+    rm -rf bitcoin-0.11.0*
+
+Before we can run bitcoind, we need to make sure that it has a place to store the blockchain and a config file that contains a username and password for the daemon. The commands below will set up your bitcoin directory and give bitcoind a default username and a random password (you do not need to remember the password for standard operation).
+
+    mkdir ~/Library/Application\ Support/Bitcoin
+    touch ~/Library/Application\ Support/Bitcoin/bitcoin.conf
+    chmod 600 ~/Library/Application\ Support/Bitcoin/bitcoin.conf
+    echo "rpcuser=bitcoinrpc" >> ~/Library/Application\ Support/Bitcoin/bitcoin.conf
+    echo "rpcpassword=$(cat /dev/urandom | env LC_CTYPE=C tr -dc a-zA-Z0-9 | head -c45)" >> ~/Library/Application\ Support/Bitcoin/bitcoin.conf
+
+You should now be able to start up your full node by running `bitcoind -daemon` in any Terminal window. If you need to stop bitcoind for any reason, the command is `bitcoin-cli stop`
+
+<div class="box" markdown="1">
+*Optional: Start Your Node At Login*
+
+Starting your node automatically each time you login to your computer makes it easy for you to contribute to the network. The easiest way to do this is to tell Bitcoin Core Daemon to start at login. In OS X, the way to start background programs at login is using a Launch Agent. Here is how to install a Launch Agent for Bitcoin Core daemon on your machine:
+
+    mkdir ~/Library/LaunchAgents
+    curl https://raw.githubusercontent.com/bitcoin/bitcoin/master/contrib/init/org.bitcoin.bitcoind.plist > ~/Library/LaunchAgents/org.bitcoin.bitcoind.plist
+
+The next time you login to your desktop, Bitcoin Core daemon will be automatically started.
+</div>
+
+{{installFinished}}
 
 ## Network Configuration
 
--- a/en/full-node.md
+++ b/en/full-node.md
@@ -1109,22 +1109,22 @@ automatically started minimized in the task bar.
 
 The Bitcoin Core daemon (bitcoind) is not included in the .dmg file you may have downloaded to install Bitcoin-QT. Bitcoind, along with its support binaries, is instead included in the OS X .tar.gz file listed on the official Bitcoin Core download page. To download this file using Terminal, execute the following command:
 
-    curl -O https://bitcoin.org/bin/bitcoin-core-0.11.0/bitcoin-0.11.0-osx64.tar.gz
+    curl -O https://bitcoin.org/bin/bitcoin-core-{{site.DOWNLOAD_VERSION}}/bitcoin-{{site.DOWNLOAD_VERSION}}-osx64.tar.gz
 
 {{verifyReleaseSignatures}}
 
 Extract bitcoind and its support binaries from the archive we just downloaded by running this command in Terminal:
 
-    tar -zxvf bitcoin-0.11.0-osx64.tar.gz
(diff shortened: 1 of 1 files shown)


en/full-node.md
```diff
@@ -1107,10 +1107,47 @@ automatically started minimized in the task bar.
...
-you what we need.
+The Bitcoin Core daemon (bitcoind) is not included in the .dmg file you may have downloaded to install Bitcoin-QT. Bitcoind, along with its support binaries, is instead included in the OS X .tar.gz file listed on the official Bitcoin Core download page. To download this file using Terminal, execute the following command:
+
+    curl -O https://bitcoin.org/bin/bitcoin-core-0.11.0/bitcoin-0.11.0-osx64.tar.gz
+
+{{verifyReleaseSignatures}}
+
+Extract bitcoind and its support binaries from the archive we just downloaded by running this command in Terminal:
+
+    tar -zxvf bitcoin-0.11.0-osx64.tar.gz
```
harding: This is a suggestion, not a request, but I always prefer to omit the -v (verbose) flag from tar so that I see any errors instead of having them buried in a flood of output.  The Bitcoin tarball should be small enough that the command will return almost immediately, so users won't be sitting there wondering, "is it doing anything?"

en/full-node.md
```diff
@@ -1107,10 +1107,47 @@ automatically started minimized in the task bar.
...
+
+Now we'll move the executables into your default path to make running and stopping bitcoind easier. To move the executables, run these commands (note that we have to use `sudo` to perform these commands since we are modifying directories owned by root):
+
+    sudo mkdir -p /usr/local/bin
+    sudo cp bitcoin-0.11.0/bin/bitcoin* /usr/local/bin/.
```
harding: This will also move bitcoin-qt to /usr/local/bin/.  Should the introductory part of the instructions recommend that the user uninstall the .dmg before installing using the tarball?

en/full-node.md
```diff
@@ -1107,10 +1107,47 @@ automatically started minimized in the task bar.
...
+
+To clean up the directory we've been working in, run:
+
+    rm -rf bitcoin-0.11.0*
+
+Before we can run bitcoind, we need to make sure that it has a place to store the blockchain and a config file that contains a username and password for the daemon. The commands below will set up your bitcoin directory and give bitcoind a default username and a random password (you do not need to remember the password for standard operation).
+
+    mkdir ~/Library/Application\ Support/Bitcoin
```
harding: Should there be a note here and for the `mkdir ~/Library/LaunchAgents` command below to ignore errors about the directory already existing?

harding:
Left a few comments, but otherwise looks good to me.  Thanks!

en/full-node.md
carnesen: Installing bitcoin-qt from the .dmg puts the Mac app bundle into the Applications folder. So there won't be a collision. Of course if the user is *running* bitcoin-qt installed from the .dmg they won't be able to run bitcoind at the same time with the default settings.

en/full-node.md
carnesen: It'd be good to remove from the instructions any hard-coded reference to a specific version of Bitcoin Core. Else we'll be in a position of having to update the instructions every time a new version is released.

Perhaps something like "In a web browser, navigate to https://bitcoin.org/en/download and click on the Mac OS X .tar.gz link to download the latest available version of the software. Next click on the downloaded .tar.gz in your downloads folder to expand the tarball into a directory of the same name."

Alternatively, @harding is there's a link available on the site for downloading the latest version of the OS X tarball? Or perhaps there's a site-wide macro whose value is the current version of Bitcoin Core?

en/full-node.md
carnesen: "mkdir -p" means "make the directory if it doesn't already exits". It's there for the /usr/local/bin command but needs to be added for the "~/Library/Application\ Support/Bitcoin"

en/full-node.md
```diff
@@ -1107,10 +1107,47 @@ automatically started minimized in the task bar.
...
+    touch ~/Library/Application\ Support/Bitcoin/bitcoin.conf
+    chmod 600 ~/Library/Application\ Support/Bitcoin/bitcoin.conf
+    echo "rpcuser=bitcoinrpc" >> ~/Library/Application\ Support/Bitcoin/bitcoin.conf
+    echo "rpcpassword=$(cat /dev/urandom | env LC_CTYPE=C tr -dc a-zA-Z0-9 | head -c45)" >> ~/Library/Application\ Support/Bitcoin/bitcoin.conf
+
+You should now be able to start up your full node by running `bitcoind -daemon` in any Terminal window. If you need to stop bitcoind for any reason, the command is `bitcoin-cli stop`
```
carnesen: -daemon is redundant as an argument to bitcoind.

en/full-node.md
```diff
@@ -1107,10 +1107,47 @@ automatically started minimized in the task bar.
...
+
+<div class="box" markdown="1">
+*Optional: Start Your Node At Login*
```
carnesen: There are currently two nearly-identical "Optional: Start Your Node At Boot" sections on the page. This would be the third. I believe that we could replace all three with a macro along the lines of:
```
<div class="box" markdown="1">
*Optional: Start Your Node At Boot*

Starting your node automatically each time your computer boots makes it
easy for you to contribute to the network.  The easiest way to do this
is to start Bitcoin Core daemon from your crontab.  To edit your
crontab on most distributions, run the following command:

    crontab -e

Scroll to the bottom of the file displayed and add the following line:

    @reboot bitcoind -daemon

Save the file and exit; the updated crontab file will be installed for
you. On most distributions, this will cause Bitcoin Core daemon to be
automatically started each time your reboot your computer.

If you're a expert system administrator and want to use an init script instead, see
[the init scripts directory in Bitcoin Core's source tree](https://github.com/bitcoin/bitcoin/tree/master/contrib/init).
</div>
```
I've never actually tried to run a cron job on a mac, but somebody else said it would work and the Mac documentation implies it will too.

en/full-node.md
harding: Yes, `{{site.DOWNLOAD_VERSION}}` should produce the current version number only (current output: 0.11.0)

The Mac filenames used to change fairly often, but I think we've sorted that out and they've been stable for the last few releases.

en/full-node.md
harding: Oh, hmm.  Actually, I remember the command as 'mkdir (and parents if necessary)' but I didn't realize until testing it right now that it doesn't emit an error if the directory already exists (plain mkdir does).  Thanks for the correction; please ignore my comment above.

en/full-node.md
harding: On Linux, it forks() it into a detached background process.  Is that not the case for OSX?

carnesen: On OS X, "bitcoind" with or without -daemon forks it into the background. I thought that was true on Linux too but just tried it and indeed -daemon is required there (or &). I'm happy to just leave the -daemon in there to be consistent with the Linux instructions.

kevcooper:
From [Apple's developer library](https://developer.apple.com/library/mac/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/ScheduledJobs.html):
    Note: Although it is still supported, cron is not a recommended solution. It has been deprecated in favor of launchd.

I believe it will work for now, but if apple decides to fully remove cron in the future we would have to add launchd instructions anyways.

On the two systems I've tested (both OS X 10.10.5), if the `-daemon` option is not included, bitcoind keeps control of the session and does not go into the background, so I have no idea why it's backgrounding for you @carnesen (maybe you have -daemon specified in your bitcoin.conf or something?)

en/full-node.md
kevcooper: These are good suggestions, I'll update to this:

    curl -O https://bitcoin.org/bin/bitcoin-core-0.11.0/bitcoin-{{site.DOWNLOAD_VERSION}}-osx64.tar.gz

and

    tar -zxf bitcoin-{{site.DOWNLOAD_VERSION}}-osx64.tar.gz

carnesen:
Yea I saw that article. Similar could be said about cron vs init on Linux

en/full-node.md
```diff
@@ -1107,10 +1107,47 @@ automatically started minimized in the task bar.
...
+    sudo mkdir -p /usr/local/bin
+    sudo cp bitcoin-{{site.DOWNLOAD_VERSION}}/bin/bitcoin* /usr/local/bin/.
+
+To clean up the directory we've been working in, run:
+
+    rm -rf bitcoin-{{site.DOWNLOAD_VERSION}}*
+
+Before we can run bitcoind, we need to make sure that it has a place to store the blockchain and a config file that contains a username and password for the daemon. The commands below will set up your bitcoin directory and give bitcoind a default username and a random password (you do not need to remember the password for standard operation).
+
+    mkdir ~/Library/Application\ Support/Bitcoin
```
harding: Suggest making this `mkdir -p` so there's no error if the user had previously installed and run Bitcoin Core GUI.

harding:
Preview of this PR rendered: http://dg4.dtrt.org/en/full-node#osx-daemon

I have no opinion on cron versus launchd on Mac.  I suggest we go with @kevcooper's text on that and let @carnesen open a separate PR later with the macro (although I would suggest continuing to provide more specific advice in each section than just visiting the Bitcoin Core contrib directory).

I'm going to tag this as Need More Info for now so it doesn't show up in the active pull requests. I'll untag it as soon as https://github.com/bitcoin/bitcoin/pull/6621 gets merged.  Thanks again!

kevcooper:
bitcoin/bitcoin#6621 has been merged

harding:
@kevcooper thanks for seeing that through!  The command `curl https://raw.githubusercontent.com/bitcoin/bitcoin/master/contrib/init/org.bitcoin.bitcoind.plist` now works, so I'm going to tag this as ready for merge.  If there's no critical feedback, it'll be merged on Monday.  Thanks again!

PR #1048, opened 2015-09-04
Conditionally build some pages differently based on environmental variables

harding:
This PR has three commits:

1. A new plugin which makes environmental variables (env vars) starting with `BITCOINORG_` available as template variables.

2. A conditional to the robots.txt file that prevents the site from being indexed if an env var is set to "preview".  This is useful for @saivann and myself who host previews of the site but don't want them to appear in search engine results.  Documentation for this is also added to README.md

3. A conditional on two of the content pages that changes how they will appear on cloned versions of the site (of which there are now several besides bitcoin.com).  Specifically:

    - The Bitcoin Core download page on cloned sites redirects to the language equivalent page on Bitcoin.org
    - The About Us page has all information replaced by a single sentence: "This site includes content originally published on Bitcoin.org, but it is not affiliated with Bitcoin.org."

    It is not the goal to prevent people from cloning the site but, rather, to ensure that cloned sites have accurate information about Bitcoin Core and that the contributors to Bitcoin.org don't get blamed for clone site problems.

### Previews

The above commits mean the site can be compiled in three different states, so a separate preview is provided for each:

- **No variables set:** this is what I expect clone sites to do.  Preview of the [About Us page](http://dg0.dtrt.org/en/about-us) and the [Download page](http://dg0.dtrt.org/en/download) (should automatically redirect to Bitcoin.org).  The [robots.txt](http://dg0.dtrt.org/robots.txt) and other site pages are unaffected.

- **BITCOINORG_BUILD_TYPE=preview** this is what I expect to use myself for preview builds in the future.  The [robots.txt](http://dg1.dtrt.org/robots.txt) denies all indexing but the [About Us](http://dg1.dtrt.org/en/about-us) and [Download](http://dg1.dtrt.org/en/download) pages are unaffected, as is the rest of the site.

- **BITCOINORG_BUILD_TYPE=deployment** this is what the build script will be running from now on ([as will Travis](https://travis-ci.org/harding/bitcoin.org/builds/78823051#L82)).  Except for inconsequential additional empty lines, all content is identical to the merge base, including the [robots.txt](http://dg2.dtrt.org/robots.txt), [About Us page](http://dg2.dtrt.org/en/about-us), and [Download page](http://dg2.dtrt.org/en/download)

---

I will update both the build server build script and the translation preview server build script at merge time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,6 +297,27 @@ run if the API site is running slow.
 
 For a list of languages, look in the `_translations` directory.
 
+#### Publishing Previews
+
+You can publish your previews online to any static hosting service.
+[GitHub pages](https://pages.github.com/) is a free service available to
+all GitHub users that works with Bitcoin.org's site hierarchy.
+
+Before building a preview site, it is recommended that you set the
+environmental variable `BITCOINORG_BUILD_TYPE` to "preview".  This will
+enable some content that would otherwise be hidden and also create a
+robots.txt file that will help prevent the site from being indexed by
+search engines and mistaken for the actual Bitcoin.org website.
+
+In the bash shell, you can do this by running the following command line
+before building you preview:
+
+    export BITCOINORG_BUILD_TYPE=preview
+
+You can also add this line to your `~/.bashrc` file if you frequently
+build site previews so that you don't have to remember to run it for
+each shell.
+
 ## Developer Documentation
 
 Most parts of the documentation can be found in the [_includes](https://github.com/bitcoin-dot-org/bitcoin.org/tree/master/_includes)
--- a/_templates/about-us.html
+++ b/_templates/about-us.html
@@ -5,6 +5,7 @@
 layout: base
 id: about-us
 ---
+{% if site.env.BITCOINORG_BUILD_TYPE %}
 <!-- Note: this file is built non-deterministically -->
 <h1>{% translate pagetitle %}</h1>
 <p class="summary">{% translate pagedesc %}</p>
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -297,6 +297,27 @@ run if the API site is running slow.
...
+[GitHub pages](https://pages.github.com/) is a free service available to
+all GitHub users that works with Bitcoin.org's site hierarchy.
+
+Before building a preview site, it is recommended that you set the
+environmental variable `BITCOINORG_BUILD_TYPE` to "preview".  This will
+enable some content that would otherwise be hidden and also create a
+robots.txt file that will help prevent the site from being indexed by
+search engines and mistaken for the actual Bitcoin.org website.
+
+In the bash shell, you can this by running the following command line
```
saivann: @harding s/you can this/you can do this/

saivann:
Thanks, very nice! Untested LGTM.

_templates/about-us.html
```diff
@@ -83,3 +84,16 @@ <h3 id="github">{% translate github %}</h3>
   </div>
   {% endfor %}
 </div>
+{% else %}
+{% comment %}
+<!-- if you've cloned bitcoin.org, feel free to fill in your own About
+Us below. It would be appreciated it if you link back to Bitcoin.org,
+but please make it clear that your site is not affiliated with
+Bitcoin.org. -->
```
saivann: @harding Just a suggestion, maybe we could also mention "and don't forget to update instructions in the README". I don't know if that's really going to be efficient in practice but who knows.

harding:
@saivann implemented both your comments as suggested (commits amended).  Thank you for your review!

Regarding the names and emails in the README.md, I don't know how much of a problem that is.  Except for Bitcoin.com, I don't think any of the other clone sites are hosting their clone source on GitHub, so nobody sees the README file.

Long-term, I think our README.md is getting awfully long and needs something like a table of contents.  I had the idea of moving most of it to a `toc` class document similar to the dev docs with all the nice features you've added for managing long pieces of documentation.  If we did that, we could mask names, emails, and links to our pull/issue trackers using the env vars.

harding:
In the absence of critical feedback, this will be merged on Tuesday.

harding:
Merged and build script updated.  I've verified that the main site [robots.txt](https://bitcoin.org/robots.txt) and [about us page](https://bitcoin.org/en/about-us) are correct.

Issue #1040, opened 2015-09-01
Remove Hive Wallet iOS


windsok:
Hive wallet for iOS is currently completely non functional - the only way to recover funds in a hive wallet is by importing the private key into another wallet (which i've had to do).

I would recommend that Hive for iOS be removed from wallets page until these issues are resolved.

https://twitter.com/hivewallet/status/635577092772524032

windsok:
Author of the wallet also recommends not to use it (for a different issue) here - https://github.com/hivewallet/hive-ios/issues/49

harding:
Hive for Mac was removed in PR #987 and (I just checked) is not displayed on the site:

https://bitcoin.org/en/choose-your-wallet

crwatkins:
@harding  #987 Removed Hive for Mac, not Hive for iOS.

harding:
Oh, bugger.  I need to learn to read.  Sorry.

harding:
@w-hive what is your opinion on delisting all of the Hive wallets from Bitcoin.org until you're able to address these issues?

w-hive:
Yeah for sure, pull iOS -- and thanks for asking. Unfortunately we're out of funds and no real development is going on right now, although the Web version has been updated, and should be fine. We'll be putting up a message about this shortly and pulling our iOS app from the Store.

If we get our house refurnished we'll pop a note in here and ask to be restored.

harding:
@w-hive thank you for your response and your candor.  I'll open a PR momentarily to remove the iOS wallet.

We look forward to the day we can re-list the client-side wallets.

PR #1049, opened 2015-09-04
Wallets: Remove Hive iOS

harding:
Per discussion with dev in https://github.com/bitcoin-dot-org/bitcoin.org/issues/1040#issuecomment-137864985


crwatkins:
LGTM.

PR #1051, opened 2015-09-06
Backend: cache events and contributor listings

harding:
When making changes to the site, Saïvann and I often diff the resulting HTML output with output from earlier known-good commits.  One problem with this method is that the HTML output isn't fully deterministic---some of it comes from external sites and can change massively between two builds that are only minutes apart, making reading the diff unpleasant and less useful than it could be.

This commit caches that external data between builds so that two builds on the same computer are very nearly deterministic.  As evidence, at the end of this message I offer inline the full diff of two sequential builds both on this commit.

Cached Meetup.com events, Bitcoin Core contributor listings, and Bitcoin.org contributor listings are retained for 24 hours before being refreshed.  Cached Geo-location data for conferences is retained until the _events.yml file is updated.  Caches are updated based on file timestamps, so they can be manipulated using `touch` or simply deleted to fetch new data.

This commit includes one small miscellaneous change to the Makefile to reduce the size of the /commit.txt file.  (The replaced command didn't work the way I thought it did; this command produces the correct output.)

```diff
diff -ru _site_a/en/rss/blog.xml _site/en/rss/blog.xml
--- _site_a/en/rss/blog.xml	2015-09-06 18:52:46.785203305 -0400
+++ _site/en/rss/blog.xml	2015-09-06 19:07:01.027678049 -0400
@@ -5,8 +5,8 @@
     <description>News about Bitcoin.org</description>
     <link>//</link>
     <atom:link href="//en/rss/blog.xml" rel="self" type="application/rss+xml"/>
-    <pubDate>Sun, 06 Sep 2015 18:45:28 -0400</pubDate>
-    <lastBuildDate>Sun, 06 Sep 2015 18:45:28 -0400</lastBuildDate>
+    <pubDate>Sun, 06 Sep 2015 18:59:39 -0400</pubDate>
+    <lastBuildDate>Sun, 06 Sep 2015 18:59:39 -0400</lastBuildDate>
     <generator>Jekyll v1.3.0</generator>
     
       <item>
diff -ru _site_a/sha256sums.txt _site/sha256sums.txt
--- _site_a/sha256sums.txt	2015-09-06 18:52:47.465211621 -0400
+++ _site/sha256sums.txt	2015-09-06 19:07:01.727686654 -0400
@@ -325,7 +325,7 @@
 34e1e82c7810f261baec99a73c0327c0a05a46d2869ed57840130daf5ea5fddd  _site/en/release/v0.9.3.html
 63b90f09f0def9780a9da8eeef0e4430e147b156314a0a1931ef17afa38920a5  _site/en/resources.html
 ea62e46ad3c8fc96824074ba21abc9bad9df2b145d2b230ae84a69b0d5ca8035  _site/en/rss/alerts.rss
-9f772391bb916ebd273d5e9d03d4d45b87f3d4f0a3440effc0537c10163f47d5  _site/en/rss/blog.xml
+4b6f77955753228f295068092928b87db7ea43a87c71b7f767c1ee7e2106ff7a  _site/en/rss/blog.xml
 49f3438da26309fc3c20522dae6cab65efb3c502327df6693933239fc53fe1c2  _site/en/rss/releases.rss
 d2b7db6b8bada8f96aeb505e0b284a28b83fda625c91aac48684dfde73cfd35e  _site/en/secure-your-wallet.html
 5e77e94a36c5698486b0d1fc383a7a7f874131d5be0b38973d6548758c89ba5c  _site/en/support-bitcoin.html
```

### DIFF
--- a/_plugins/contributors.rb
+++ b/_plugins/contributors.rb
@@ -106,9 +106,44 @@ module Jekyll
         return
       end
 
-      # Populate site.corecontributors and site.sitecontributors arrays
-      site.corecontributors = contributors('bitcoin/bitcoin',site.config['aliases'])
-      site.sitecontributors = contributors('bitcoin-dot-org/bitcoin.org',site.config['aliases'])
+      ## Create cache directory if it doesn't exist
+      if !File.exists?('_cache')
+        Dir.mkdir('_cache')
+      end
+
+      # Populate site.corecontributors and site.sitecontributors with
+      # data from GitHub.com. Store data in the cache and only
+      # re-retrieve the data if 86,400 seconds (24 hours) passes from
+      # the retrieval date or if the cache file is deleted.  For
+      # simplicity, updates on the two cache files are linked, so if one
+      # file has to be updated, they both get updated.
+      corecontributors_cache = '_cache/corecontributors.marshall'
+      sitecontributors_cache = '_cache/sitecontributors.marshall'
+      if File.exists?(corecontributors_cache) && File.exists?(sitecontributors_cache)
+        corecontributors_cache_age = (Time.now - File.stat(corecontributors_cache).mtime).to_i
+        sitecontributors_cache_age = (Time.now - File.stat(sitecontributors_cache).mtime).to_i
+      else
+        corecontributors_cache_age = Time.now.to_i
+        sitecontributors_cache_age = Time.now.to_i
+      end
+
+      if corecontributors_cache_age > 86400 || sitecontributors_cache_age > 86400
+        site.corecontributors = contributors('bitcoin/bitcoin',site.config['aliases'])
+        File.open(corecontributors_cache,'w') do |file|
+          Marshal.dump(site.corecontributors, file)
+        end
+        site.sitecontributors = contributors('bitcoin-dot-org/bitcoin.org',site.config['aliases'])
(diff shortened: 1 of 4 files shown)


_plugins/contributors.rb
```diff
@@ -106,9 +106,42 @@ def site_payload
         return
       end
 
-      # Populate site.corecontributors and site.sitecontributors arrays
-      site.corecontributors = contributors('bitcoin/bitcoin',site.config['aliases'])
-      site.sitecontributors = contributors('bitcoin-dot-org/bitcoin.org',site.config['aliases'])
+      ## Create cache directory if it doesn't exist
+      system 'mkdir', '-p', '_cache'
```
saivann: @harding How about the ruby equivalent `Dir.mkdir('_cache') if !File.exists?('_cache')`
(This line of code also occurs elsewhere)

saivann:
That solution is also elegant in that it prevents repeated API calls by default when testing! And probably improve performances quite a bit at the same time.

Thanks, LGTM!

harding:
@saivann amended commit to use the Ruby mkdir (didn't know about that, thanks!).  Thanks for reviewing!

In the absence of critical feedback, this will be merged on Wednesday.

PR #1057, opened 2015-09-12
Update whatcanidoforbitcoin site domain

andrewtian:
The previous domain whatcanidoforbitcoin.org recently expired. I moved the new site to whatcanidoforbitcoin.xyz.

Changes include updates to all references of the old domain.


harding:
I've confirmed the previous site is down, the new site is up and looks good to me, and @andrewtian was the same website author we dealt with before.

I'll be merging this momentarily so that we stop using the bad link ASAP.  If anyone sees any problems, I'll revert or provide an update.

Thanks for the update, @andrewtian!

PR #1058, opened 2015-09-12
Update community.html

funbucks:
Please add bitco.in to the community page.  Thank you.

### DIFF
--- a/_templates/community.html
+++ b/_templates/community.html
@@ -22,7 +22,7 @@ id: community
     <p>{% translate bitcointalk %}</p>
     {% if page.lang == 'fr' %}<p><a href="https://www.reddit.com/r/BitcoinFrance/">Communauté Bitcoin sur Reddit</a></p>{% endif %}
     <p><a href="http://reddit.com/r/Bitcoin/">{% translate reddit %}</a></p>
-    <p><a href="http://bitco.in/forum//">{% translate bitco.in %}</a></p>
+    <p><a href="http://bitco.in/forum/">{% translate bitcoin forum %}</a></p>
     <p><a href="https://bitcoin.stackexchange.com/">{% translate stackexchange %}</a></p>
   </div><div>
     <h2 id="social"><img src="/img/icons/ico_network.svg" class="titleicon" alt="Icon">{% translate social %}</h2>
--- a/_templates/community.html
+++ b/_templates/community.html
@@ -22,7 +22,7 @@ id: community
     <p>{% translate bitcointalk %}</p>
     {% if page.lang == 'fr' %}<p><a href="https://www.reddit.com/r/BitcoinFrance/">Communauté Bitcoin sur Reddit</a></p>{% endif %}
     <p><a href="http://reddit.com/r/Bitcoin/">{% translate reddit %}</a></p>
-    <p><a href="http://bitco.in/forum/">{% translate bitcoin forum %}</a></p>
+    <p><a href="http://bitco.in/forum/">bitcoin forum</a></p>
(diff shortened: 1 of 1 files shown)


harding:
You're hosting [a copy of our site](http://bitco.in/en/); why should we link to you when we already provide all of that content?

funbucks:
Thank you for your response.  Your input is welcome here, to the site administrator, bloomie.

http://bitco.in/forum/threads/link-from-bitcoin-org.45/#post-772

BloomieBTC:
Hi @harding, I am the admin of the new Bitcoin forum.  I was under the impression that the entire contents of bitcoin.org was released under the MIT license:

*Permission is hereby granted, free of charge, to any person ...  without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software*

I assumed this was done to encourage others to propagate information about Bitcoin as much as possible.  If that's not the case, please let me know.  Thank you.

harding:
@BloomieBTC with a few small exceptions, the entire site is MIT licensed and you're free to use the content however you'd like under those terms.  I have not seen any violations on your site.

My question is from the perspective of site visitors: why should we link to your copy of the content when we already link to our copy of the content?  (Especially since we have more control over our copy of the content and update it regularly.)

BloomieBTC:
@harding Ah, I understand your question now.  The proposed link is only to the new [Bitcoin Forum](http://bitco.in/forum/) that we're hosting, not the entire site.  I agree that it wouldn't make sense to link to an outdated version of a mirror from the original source!  

The forum, by the way, is the only Bitcoin community that encourages both Bitcoin Core and BitcoinXT development discussions and does not censor or ban members based on which side they take in the block size limit debate.  We have an extremely high signal-to-noise ratio, no promotions of pyramid schemes or other shady sites (unlike certain other forums), and virtually no trolling.  

We would be glad if you joined us and contributed to the discussion as well.

harding:
Temporarily ignoring my concerns about your hosting a copy of our site, I'm also troubled that Cypherdoc is a [moderator](https://archive.is/5XCLT) on your site given [the negative information](https://www.reddit.com/r/Bitcoin/comments/3kabxu/emin_g%C3%BCn_sirer_the_case_involving_bitcoin_scam/) about his previous activities.

Beyond that, I think maybe your forum is just too new, aparently being less than a month old and having fewer than 700 posts total---fewer than some large individual /r/bitcoin threads.  (And over 1/3 of them have been posted by Cypherdoc!)  Maybe we can close this issue for now and re-evaluate in three months?

BloomieBTC:
Thanks, I agree with re-evaluating later.  The forum is growing fast, and, again, the quality of the content is superb.  

I've also messaged cypherdoc about your concerns in case he needs to address them.  FYI, he joined the forum after his thread on another forum got shut down for discussing XT.  He then requested to moderate the sub-forum (currently called Economics) where he would restart the thread, and I thought that would be fair.

The invitation for you to join and share your knowledge with us stands.  I know you've contributed an enormous amount to Bitcoin, and we have fewer Core supporters on the forum right now than I'd like.  I personally can't stand the rift in the community and hope this will all be over soon.

wbnns:
@BloomieBTC Hello, after further consideration, it would be potentially harmful to Bitcoin.org to merge this PR and incorporate the link. This is because search engines like Google, for example, are very strict about duplicate content in their search engine ranking pages (SERPs). If links are added to other websites that have the same content, then we could face having our search results removed from the index altogether.

Currently we have many links that appear in Google's search results that lead new users to the site, providing very helpful information. Here is an example:

image

Thank you for taking the time to contribute.

PR #1060, opened 2015-09-15
Blockchain is one word

sandakersmann:
Blockchain was added to Oxford Dictionaries Online in August 2015.

http://www.coindesk.com/oxford-dictionaries-definitions-blockchain-miner/


harding:
There's a related pull here: https://github.com/bitcoin/bitcoin/pull/6666

To minimize time spend on low-yield tasks, I'm not a fan of changing any reasonable and consistent phrasing until there's strong evidence that our current usage is wrong.  I'm also not interested in debating this sort of *what color to paint the bike shed* kind of decision---so I'm only going to make this one comment about it.  If someone else decides to merge this, I'm ok with that.

Possible problems I see with the current pull request:

1. It failed the Travis CI checks.

2. I suspect it misses occurences of "block chain" that wrap across lines, such as:

        Satoshi Nakamoto invented the block
        chain.

3. It changes text in SVG images.  I would like assurances that someone manually checked each image to verify that this doesn't mess up the layout of any of those images.  Also, the corresponding PNG images need to be regenerated and optimized with optipng.

schildbach:
FWIW, a couple of month ago I replaced all blockchain > block chain in Bitcoin Wallet, after looking up on bitcoin.org what's the correct spelling…

sandakersmann:
I have manually checked each .svg image and can verify that the layout did not get messed up.

saivann:
The [wiki](https://en.bitcoin.it/wiki/Block_chain) also uses "block chain", while satoshi's paper only use "chain". Google suggests "blockchain" is around 5 times more popular (with "blockchain.info" excluded from search results). I guess both terms are fine and also don't worry much if we keep "block chain" on bitcoin.org, but "blockchain" indeed seems to becoming more or less the new "standard".

theymos:
Satoshi always used "block chain", never "blockchain", so if I had to pick one I'd choose "block chain". Though I use them interchangeably.

sandakersmann:
The problem with not having a standard is that people use different versions interchangeably. This produce a lot of unnecessary work down the road. The faster that we can establish a standard the better it is in terms of wasted work. Since blockchain was added to Oxford Dictionaries Online in August 2015, I think that we should go in that direction sooner rather than later.

http://www.coindesk.com/oxford-dictionaries-definitions-blockchain-miner/

saivann:
@sandakersmann I don't see where this produces a lot of unnecessary work (or any other serious issue). Language is distributed and evolves organically, it's never much something set in stone.

sandakersmann:
Schildbach said in his post that he have wasted work on Bitcoin Wallet, due to him looking for the correct version to write blockchain on bitcoin.org. Words that are in the dictionary does not evolve much.

saivann:
@sandakersmann To me that's a good example of the cost that goes with trying to all use a consistent terminology, and the risk of going back and forth, which you also suggest doing with bitcoin.org. Conversely, there is no extra work when developers pick what they prefer and stick to it.

sandakersmann:
It is extra work if developers just want to use the version that is correct according to the dictionary, but get the first try wrong due confusion over the standard.

sandakersmann:
Some projects are also a collaboration between several developers...

carnesen:
FWIW, Wikipedia uses the two-word variant "block chain", a convention that
I supported as an editor there.

sandakersmann:
Changing a page on Wikipedia is a lot easier than getting Oxford Dictionaries Online to change a word that they have added to the dictionary. People will also follow their lead and use blockchain, and so should we. This means some work now, but a lot less work in the future.

schildbach:
"Block chain" also has the advantage of being more distinct against blockchain.info.

carnesen:
@sandakersmann A dictionary defines words not phrases. Oxford Dictionary states that the word "blockchain" exists and means "A digital ledger in which transactions made in bitcoin or another cryptocurrency are recorded chronologically and publicly". That's a true statement. I claim it's also true that word is more commonly/properly written in its non-contracted phrase form "block chain". I suspect the editors of the Oxford Dictionary might be amenable to adding a few words to their definition explaining as much.

So I don't see the existence of the "blockchain" definition as evidence in favor of switching everything to "blockchain" here or in other documentation such as Wikipedia, which by the way is *very* difficult to make changes to for highly contentious and visible articles like [Bitcoin](https://en.wikipedia.org/wiki/Bitcoin). I'd bet good money that I'd have an easier time convincing the editors of Oxford Dictionary to add an explanatory note saying "block chain" is more common than the editors of Wikipedia to switch everything to "blockchain" preferentially.

saivann:
It's worth to note that some dictionnaries consider both "email" and "e-mail" as valid too, blockchain/block chain is likely a similar situation. I appreciate that you have submitted this but I think it's obvious that both will remain in use. I personally think it's better to not merge this pull request just so we don't spend time debating between two solutions that are good enough.

PR #1072, opened 2015-09-30
Fixed small grammar issues

i-rme:


saivann:
@i-rme Thanks! In the future, can you update translations upstream on Transifex?

https://www.transifex.com/bitcoinorg/bitcoinorg

This will prevent your work from being later reverted when translations are imported. I took care to update them for you already. Thanks again!

PR #1073, opened 2015-09-30
Fix double space in string

i-rme:


harding:
Is this causing any display problems?  If not, we probably want to leave it the way it is so that we don't have to mess around with Transifex.

saivann:
@i-rme Thanks, I have updated the string upstream on Transifex. There is no hurry with publishing that change as it shouldn't make any difference on the website (double-spaces don't display in browsers).

PR #918, opened 2015-06-26
About-Us: History, Staff Updates, & Signed Commits

harding:
This PR contains three somewhat related commits:

### 1—More Bitcoin.org History & Ownership Info

By multiple request, we've expanded the "Who owns bitcoin.org" section to cover more history and more information about who owns Bitcoin.org.  Text highlighted in green below is new or substantially revised:

screenshot-btcorg localhost 2015-06-25 20-04-41
 
### 2—Staff Updates

We simplified titles in the Maintainer section:

screenshot-btcorg localhost 2015-06-25 20-09-25

Along with the title changes, I (@harding) will be focusing more time on documentation and, shortly after this PR is merged, we will be giving Craig (@crwatkins) edit/commit access to help him more efficiently continue his excellent work on the Choose Your Wallet page.

We've also happily been able to move Will (@coderwill) from the Inactive Contributor section back to the position of Translation Coordinator:

screenshot-btcorg localhost 2015-06-25 20-15-12

Finally, we've added a section for the domain co-owners who wish to be publicly listed:

screenshot-btcorg localhost 2015-06-25 20-35-34

### 3—Signed Commits For Auto-Building

The final commit doesn't affect the site layout at all, but is related to how the site gets updated.  The site is currently automatically re-built whenever a new commit is pushed to the repository master branch.  Once this PR is merged, that will only be the case if the tip of the commit tree is signed by one of the authorized keys below:

    pub  2048R/D0DE98BD 2014-05-14 David A. Harding <dave@dtrt.org>
    sub  2048R/FB3E1276 2014-05-14
    sub  2048R/F29EC4B7 2014-05-14
    pub  2048R/1B5577B0 2013-03-21 Saivann Carignan <saivann@gmail.com>
    sub  2048R/3AC31877 2013-03-21
    pub  2048R/2346C9A6 2011-08-24 Wladimir J. van der Laan <laanwj@gmail.com>
    uid                            Wladimir J. van der Laan <laanwj@visucore.com>
    sub  2048R/890DE427 2011-08-24
    pub  2048D/DAB591E7 2010-08-22 Michael Marquardt <michael_m+pgp@mm.st>
    uid                            theymos <theymos+pgp@mm.st>
    sub  2048g/D3027A4F 2012-08-29 [expires: 2015-08-29]

Saïvann and I can still login to the build server and trigger a manual build for unsigned commits, for example when Craig merges a new wallet.

[...]

PR #1077, opened 2015-10-02
Enable Google analytics

Cobra-Bitcoin:
Adds Google Analytics to the site so that visitor numbers and behavior can be accurately measured. This is necessary as Bitcoin.org is currently looking to work with new advertising partners to help fund the site's expenses.

### DIFF
--- a/_includes/layout/base/html-head.html
+++ b/_includes/layout/base/html-head.html
@@ -16,5 +16,15 @@ http://opensource.org/licenses/MIT.
 {% if page.lang == 'bg' or page.lang == 'el' or page.lang == 'ko' or page.lang == 'hi' or page.lang == 'pl' or page.lang == 'sl' or page.lang == 'ro' or page.lang == 'ru' or page.lang == 'tr' or page.lang == 'uk' or page.lang == 'zh_CN' or page.lang == 'zh_TW' %}{% lesscss sans.less %}{% endif %}
 <script type="text/javascript" src="/js/base.js"></script>
 {% if page.id != 'download' %}<script type="text/javascript" src="/js/main.js"></script>{% endif %}
+<script>
+  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+  })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
+
+  ga('create', 'UA-46627233-1', 'auto');
+  ga('send', 'pageview');
+
+</script>
 <link rel="shortcut icon" href="/favicon.png">
 <link rel="apple-touch-icon-precomposed" href="/img/icons/logo_ios.png"/>
--- a/_translations/en.yml
+++ b/_translations/en.yml
@@ -642,9 +642,9 @@ en:
     title: "Privacy - Bitcoin"
     pagetitle: "Privacy"
     datacollect: "Data collected"
-    datacollecttxt: "Bitcoin.org collects anonymized server logs. These logs include IP addresses with replaced last byte, time of the visit, requested page, user agent and referer url. Bitcoin.org does not collect data using cookies."
+    datacollecttxt: "Bitcoin.org collects anonymized server logs. These logs include IP addresses with replaced last byte, time of the visit, requested page, user agent and referer url. Bitcoin.org uses Google Analytics to get visitor traffic information. Google Analytics collects cookies, IP addresses, operating system information, and screen resolution. Bitcoin.org does not have direct access to the information collected by Google Analytics."
     datause: "Usage of data"
-    datausetxt: "Collected data may be used to provide transparent public stats."
+    datausetxt: "Collected data may be used to provide transparent public stats. Data collected through Google Analytics may be used to accurately determine the number of visitors to the site, and traffic sources. Data obtained through Google Analytics may be publicly shared with potential advertising partners for business purposes."
   protect-your-privacy:
     title: "Protect your privacy - Bitcoin"
     pagetitle: "Protect your privacy"


harding:
Technical and concept ACK.  I think @saivann needs to comment on what the lawyer told him last year about cookies regarding the privacy policy we had written.

For the technical review, I verified the JS code provided is the same code as shown on Google's website [here](https://developers.google.com/analytics/devguides/collection/analyticsjs/) (with our particular tracking ID added).  I do think that this means Google could conceivably load or replace content on our pages, but I'm not personally concerned about that.

jonasschnelli:
I tend to see this in NACK territory.
Allowing Google to track bitcoin.org visitors is something i personally dislike.

Would `webalizer` or a similar non-third-party and fully-opensourced solution not be enough?

harding:
@jonasschnelli you may want to read https://github.com/bitcoin-dot-org/bitcoin.org/pull/853 where we dropped our self-hosted stats system because it couldn't deal with the crazy redirect traffic we received.  I think we're hoping that Google Analytics has already implemented all the code in their system to deal with garbage data like that, whereas we just don't have the manpower to maintain our own statistics system to the same quality.

In addition, potential advertisers may be more likely to trust Analytics data than stats from our logs (which we could fake).

Regarding privacy, I also dislike allowing third parties to track visitors (especially large organizations like Google who have access to big data to run correlations on), but the site needs revenue to pay for its server costs (at the very least) and so we need to prove the value of the site to potential advertisers.

saivann:
@Cobra-Bitcoin According to the lawyer I spoke with, we need a formal privacy policy if we enable Google Analytics (the current text on the site does not meet the definition of a privacy policy, and there is additional requirements than just writing the page), otherwise the site wouldn't be compliant with the law (since we would track cookies here, it's not just EU laws).

For the record, while I personally find Google Analytics to be very useful and even submitted it in the past, this was previously (and repeatedly) rejected by a fair amount of contributors.

gmaxwell:
How can we ensure that google (or someone who has compromised or ordered them to do so) cannot replace the content (including links to the downloads)?

If this were done would there be a way to load it only in a hidden iframe so that it would not have access to the page content (e.g. merely tracking and not tracking and manipulation)?

Would it be possible to not emit this to clients which send the do not track header?

Ninkip2p:
NACK- too dangerous in terms of the ability to load in JavaScript and manipulate content, links to wallets etc.

Cobra-Bitcoin:
@jonasschnelli Just to add to what @harding said: A lot of our visitors come from Google, so they already have a good idea of who is visiting bitcoin.org.

@saivann Can you email me all the requirements? I'll work on making a more formal privacy policy. I'm also assuming we would need a cookie disclaimer to comply with EU laws.

@gmaxwell I understand your concerns, but I don't think we can prevent that risk without significantly reducing the quality of our tracking. We already trust businesses like our hosting provider, domain registrar, SSL CA, GitHub, and more to behave properly. As long as the binaries continue to be signed, I think risks like these are something we can live with.

It might be possible to not emit this to clients which send the do not track header. It would have to be done through JavaScript though as the site doesn't have any other way of dynamically modifying content.

harding:
Forgive my pedantry, but for the record we don't trust GitHub any more AFAIK (provided there's no remotely-exploitable bug it git).  That's because, as of #918, the tip of each branch must be GPG signed by an authorized committer in order for it to be automatically built.  On the other hand, you can add a second hosting provider to that trust list because we use a separate VPS from a separate provider to build the site web content.

And, of course, that trust list should probably mention the several of us who have commit access or even direct server access.  I think we've demonstrated trustworthiness in the past, but I doubt we're any more able to resist legal or illegal compulsion than Google, and we too can have our computers compromised.

saivann:
@Cobra-Bitcoin Just sent you the information.

Re: Third party risk; In general we try to reduce that risk, I think we can't use the fact that we already trust third parties as an argument to make it worse, we have disabled Youtube and locked GitHub out for that reason. I think the real point is, does Google Analytics provide enough value to justify that risk. I am personally sort of neutral here.

Re Binary signed; It is worth to note that signatures are very rarely used. I think this also isn't a practical argument for Google Analytics.

harding:
@Cobra-Bitcoin how would you like to proceed?  If this were a regular feature request, I'd say that it had too much opposition and I would suggest closing it.  On the other hand, this is yours and @theymos's domain, so I think the final decision is up to you two.

If you're working on implementing some of the suggestions above to limit the security and privacy downsides, let me know and I'll tag this as Needs More Info so it doesn't appear in my regular queue of issues needing attention.

Thanks.

Cobra-Bitcoin:
@harding I'm working on implementing some of the suggestions above. I'll make a new pull request for that once I'm done since it involves some significant changes to the UI.

Thanks to everyone for their feedback and thoughts. I'm closing this pull request now.

laanwj:
A bit late, but I'm happy the decision turned out to not do this. Google Analytics is a giant, opaque, spying apparatus, and a central point of failure of the web.

mateuszrybin:
Would a possible solution be to track, let's say 20% of the visitors? 

**Currently without a tracking system bitcoin.org:**

- Have much harder time attracting possible advertisers without the ability to provide necessary data. 

- No ability to adapt if bitcoin.org was receiving 50% of its traffic from China, because you just don't know what language you need  to focus on. 

- no ability to improve the content and know for sure that its actuality improving the time people spend on the site. 

**The possible downsides to tracking are:**

The possibility for google to change content. 

longer loading time. 

**Possible solution**
I would say that the reward is higher than the risk, but a possible solution would be to hedge the risk by only loading google analysis on 20% of the visitors.

With 7 million people per quarter, bitcoin has high enough traffic to be able to extract a percentage from the total pool and still with 99% confidence say that the number represent the whole pool. 

I do not know if 20% would provide 99% confidence level, but somebody should be able to calculate that if the need to be precise exist:
https://en.wikipedia.org/wiki/Confidence_interval

jonasschnelli:
I agree that tracking can be useful. But why off-site via google? What is wrong with analyzing apaches access log? I think you can also extend apaches log (add preferred browser language, etc.). Do we really need browser window sizes, etc and taking all the downsides?

Your three bullet points are achievable with apache log parsing through *webalizer* or similar.

saivann:
Nginx log parsing is what we did previously and I even wrote a script for
it which was used to provide public stats :
https://github.com/saivann/bitcoinstats, and which would filter obvious
bots and DoS (note *not* DDoS)

The thing is bitcoin.org was targetted by junk traffic which made this data
baseless. Tons of hacked websites were redirected at bitcoin.org, as well
as additional suspicious traffic in addition to casual DDoS attacks.

This crossed the limits of what could be easily filtered as obvious bot
traffic without constant work. And stats became useless.

Cobra expressed hope that Google analytics would be better at filtering
this data with their large scale traffic analysis and algorithms, but
there's no garantee that it would be good enough.

mateuszrybin:
An other solution would be to make it time based instead. Include google every full last week of the month to sample to traffic.

wilbns:
Hello, how about [Piwik](http://piwik.org/) along with our own [referral spam blacklist](https://github.com/piwik/referrer-spam-blacklist)?

saivann:
@wilbns That's cool. No idea if it would actually work well in our case (although this can't be worse than server logs). This said, the privacy policy would remain a pre-requisite to using Piwik if it tracks cookies and/or if we need to log full IP addresses.

PR #1097, opened 2015-10-21
add info about minrelaytxfee default increase

achow101:
added information from the bitcoin-dev email regarding the increase of the default for the minrelaytxfee.


saivann:
@achow101 Thanks for your PR, I just updated the release notes.

PR #1101, opened 2015-10-22
correct ecc equation

mruddy:
Fixed a typo. The ECC equation had x and y mixed up.

References:

- Section 2.4.1 of http://www.secg.org/sec2-v2.pdf
- https://en.wikipedia.org/wiki/Elliptic_curve_cryptography


harding:
Not sure how I managed to do this, or how the other reviewers missed it.  Thanks for this and your other PRs lately!

Issue #1061, opened 2015-09-16
Endianness reversed


bedeho:
The "Hash Byte Order" section does a good job of trying to highlight the issue of txid/blockid byte order confusion,but the titles of the last table unfortunately gives the opposite interpretation what it should be.

As written, RPC byte order of block ids have leading zeros, this would be recognized as big endian, as these digits are most significant - given the very fact that hashing involves minimizing this value. However, the titles on the table have this reversed in the parentheses (“Big Endian”) and (“Little Endian”).

harding:
The quoted text is the phrasing I found other documentation using, which is why it says:

> Off-site documentation such as the Bitcoin Wiki tends to use the terms big endian and little endian as shown in the table below, but they aren’t always consistent. Worse, these two different ways of representing a hash digest can confuse anyone who looks at the Bitcoin Core source code and finds a so-called “big endian” value being stored in a little-endian data type.

I think maybe the best solution is to remove the labels "big endian" and "little endian" so that the examples speak for themselves.  Does that sound like an adequate solution to you?  (If so, I'll update the doc.)

bedeho:
I would recommend removing it. I have never seen RPC byte order be referred to as little endian anywhere else. There is indeed at least one possible contradiction one the bitcoin wiki on this

https://en.bitcoin.it/wiki/Block_hashing_algorithm states 
"The output of blockexplorer displays the hash values as big-endian numbers; notation for numbers is usual (leading digits are the most significant digits read from left to right)."

https://en.bitcoin.it/wiki/Protocol_documentation
"Note: Hashes in Merkle Tree displayed in the Block Explorer are of little-endian notation."

One refers to merkle tree hashes, and one refers to hashes in general. Perhaps byte order varies with different fields/values?

acityinohio:
Yeah, I agree with you both (@bedeho and I were diving into endianness in a few exchanges), I think maybe removing the labels "big endian" and "little endian" make sense, and providing another concrete example (perhaps a block header hash, since its hash value always has to be less than the target difficulty, and as the wiki link illustrates, can provide a nice anchor to describing byte order).

carnesen:
[...]

Issue #580, opened 2014-09-27
Wrong Endianess?


SergioDemianLerner:
Location: https://bitcoin.org/en/developer-reference#getblocktemplate

The text reads:
Each object in the array contains the rawtransaction data in hex and the hash of the data in little-endian hex.

I looked at the code of base_uint.GetHex() in uint256.cpp and the output hex value is in big-endian format (the internal representation is in little-endian format and GetHex() reverses it).

Am I interpreting it wrong or is the documentation wrong?

PR #1102, opened 2015-10-22
rational to rationale and little and big endian mixup fix

mruddy:
Two small fixes:

1) Fixed a typo "rational" -> "rationale"

2) Fixed a mixup between big and little endian.

About the big and little endian mixup, here's my reasoning:

https://en.bitcoin.it/wiki/Block_hashing_algorithm:
"Note that the hash, which is a 256-bit number, has lots of leading zero bytes when stored or printed as a big-endian hexadecimal constant, but it has trailing zero bytes when stored or printed in little-endian."

Example, double sha256 hash of block 300,000 header:
> echo -n '020000007ef055e1674d2e6551dba41cd214debbee34aeb544c7ec670000000000000000d3998963f80c5bab43fe8c26228e98d030edf4dcbe48a666f5c39e2d7a885c9102c86d536c890019593a470d' | xxd -r -p | sha256sum -b | xxd -r -p | sha256sum -b
5472ac8b1187bfcf91d6d218bbda1eb2405d7c55f1f8cc820000000000000000

The result is internal byte order (verified by checking the previous block hash of block header 300,001) and little endian (trailing zeros).

Also, as a double check, this verifies that the definition of "internal byte order" is consistent with the first example in the table and should be labeled as little endian:

> echo -n '00' | xxd -r -p | sha256sum -b | xxd -r -p | sha256sum -b
1406e05881e299367766d313e26c05564ec91bf721d31726bd6e46e60689539a

### DIFF
--- a/_includes/devdoc/bitcoin-core/api-intro.md
+++ b/_includes/devdoc/bitcoin-core/api-intro.md
@@ -12,23 +12,23 @@ http://opensource.org/licenses/MIT.
 
 {% autocrossref %}
 
-Bitcoin Core RPCs accept and return hashes in the reverse of their
-normal byte order. For example, the Unix `sha256sum` command would display the
-SHA256(SHA256()) hash of mainnet block 300,000's header as the
-following string:
+Bitcoin Core RPCs accept and return the byte-wise reverse of the computed
+SHA-256 hash value. For example, the Unix `sha256sum` command displays the
+SHA256(SHA256()) hash of mainnet block 300,000's header as:
 
-    5472ac8b1187bfcf91d6d218bbda1eb2405d7c55f1f8cc820000000000000000
+    > /bin/echo -n '020000007ef055e1674d2e6551dba41cd214debbee34aeb544c7ec670000000000000000d3998963f80c5bab43fe8c26228e98d030edf4dcbe48a666f5c39e2d7a885c9102c86d536c890019593a470d' | xxd -r -p | sha256sum -b | xxd -r -p | sha256sum -b
+      5472ac8b1187bfcf91d6d218bbda1eb2405d7c55f1f8cc820000000000000000
 
-The string above is also how the hash appears in the
+The above result is also how the hash appears in the
 previous-header-hash part of block 300,001's header:
 
 <pre>02000000<b>5472ac8b1187bfcf91d6d218bbda1eb2405d7c55f1f8cc82000\
 0000000000000</b>ab0aaa377ca3f49b1545e2ae6b0667a08f42e72d8c24ae\
 237140e28f14f3bb7c6bcc6d536c890019edd83ccf</pre>
 
-However Bitcoin RPCs use the reverse byte order for hashes, so if you
+However, since Bitcoin RPCs use the byte-wise reverse for hashes, if you
 want to get information about block 300,000 using the `getblock` RPC,
-you need to reverse the byte order:
+you need to reverse the requested hash:
 
     > bitcoin-cli getblock \
       000000000000000082ccf8f1557c5d40b21edabb18d2d691cfbf87118bac7254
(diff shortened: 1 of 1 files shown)


harding:
Regarding rational→rationale, if you split that off into a separate commit or PR, I'll apply that right away.

For endianness, we were discussing that in #1061 and I think we're waiting for @carnesen to provide a patch.  If that doesn't happen soon, I think I'm going to put forward my solution of just dropping the endianness labels and they're a source of never-ending confusion and debate.

carnesen:
My bad, I think I mostly figured out what's going on under the hood so far
as endianness in the hashes (at a point in the algorithm, a sequence of
bytes is interpreted as a big-endian integer) but never did write anything
up. @harding or anyone else please feel free to jump in with a PR for that
section.

mruddy:
@harding just broke out the typo into #1103.

I quickly looked through the Bitcoin Core code and the FIPS PUB 180-4 (that specifies SHA-256) and I haven't made up my mind on the endian-ness wording yet. The spec probably defines the endian-ness, but it is a matter of constructing a few sections together to get that interpretation and I need to re-read it to make sure.

mruddy:
Here is my current thinking on the endianness topic.

First, we must determine whether the SHA-256 result should be considered a 256-bit integer or simply an array of 32 bytes.

If considered to be a 256-bit integer, then based on sections 2.2.1, 3.1 2, and 6.2.2 of FIPS PUB 180-4 (http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf) and Bitcoin Core's implementation of SHA-256 (https://github.com/bitcoin/bitcoin/blob/c719cefc417cc578f48b33069b764339a61054ce/src/crypto/sha256.cpp#L174-L181) I think that a fair case could be made that it is a big-endian value.

For our purposes though, I think it is more natural to think of the SHA-256 result as being simply an array of 32 8-bit bytes (aka a blob).
I say this because the Bitcoin Core code treats the (previous) block, Merkle root, and transaction hashes as being of type uint256 which, from a comment in the code, "is called uint256 for historical reasons only. It is an opaque blob of 256 bits and has no integer operations.".
Treating the hash as a blob also makes the most sense when considering how the proof-of-work check against the target threshold is carried out (https://github.com/bitcoin/bitcoin/blob/c719cefc417cc578f48b33069b764339a61054ce/src/pow.cpp#L100). The POW check basically just treats the blob as a single 256-bit value stored as a little-endian when comparing against the target threshold.

Basically, depending on your perspective, you could correctly consider the same sequence of bytes in memory as being either a blob, a single big-endian value, or a single little-endian value.
That's too damn confusing. I think the endian terminology should be avoided and "reverse of the computed hash" style terminology should be used instead.

Thus, I essentially agree with Sergio's comment https://github.com/bitcoin-dot-org/bitcoin.org/issues/580#issuecomment-57158555

I've adapted my proposed commit. Looks like @harding was doing some work concurrently. However you want to handle this is fine with me. I might have to fix some formatting in this latest commit anyways.

harding:
@mruddy thanks!  I did indeed just open https://github.com/bitcoin-dot-org/bitcoin.org/pull/1105 -- what do you think about the proposal of dropping the endian references in the text and then summarizing the issues in a Bitcoin Wiki page, which Bitcoin.org will then link to?

harding:
I just reviewed this updated PR, and I'll probably adapt some of it into #1105 (particularly the terminology).

luke-jr:
FWIW, more endian "fun" coming soon... apparently versionbits will number the bits:

```
7 6 5 4 3 2 1 15 14 13 12 11 10 9 8 23 22 21 20 19 18 17 16 31 30 29 28 27 26 25 24
```

I don't care enough to argue for fixing this, so I guess it'll be left for documentation to deal with. :(

harding:
@luke-jr wow.  And I thought nBits was weirdest data structure in Bitcoin.

mruddy:
@harding Yes, dropping endian references in this DevDoc and then linking out to the wiki for more detail could make sense. The FIPS pub and code references probably make more sense in a more detailed wiki entry or appendix rather than in-line. I'll be happy to review what you do for #1105 if you want me to close this PR. It's funny how I initially thought this PR would be a quick update and then how much more complex it really was in the code and FIPS spec when I went down the rabbit hole.

mruddy:
handled by #1105

PR #1105, opened 2015-10-23
Dev Docs: drop mention of endianness in hash byte order

harding:
This tries to quickly solve some of the issues raised in #1061 and #1102 about the description of endianness in the hash byte order text.

In my opinion, simply thinking about hashes as having endianness tends to confuse developers, since hashes are usually treated as byte arrays rather than numbers.  This PR simply drops the references to endianness in the hopes of minimizing confusion to new developers as well as allowing established developers to spend less time arguing about whether header hashes in Bitcoin are little or big endian.

I'm 100% open to a better patch that describes all the relevant considerations so far discussed, although I think it might be better to put that on the Wiki and then link to it from here.  (That's a suggestion; feel free to PR it here if you think it's important.)

Screenshot of the revised section:

2015-10-23-100237_615x874_scrot

### DIFF
--- a/_includes/devdoc/bitcoin-core/api-intro.md
+++ b/_includes/devdoc/bitcoin-core/api-intro.md
@@ -43,20 +43,13 @@ target. Whatever the reason for reversing header hashes, the reversal
 also extends to other hashes used in RPCs, such as TXIDs and merkle
 roots.
 
-Off-site documentation such as the Bitcoin Wiki tends to use the terms
-big endian and little endian as shown in the table below, but they
-aren't always consistent. Worse, these two different ways of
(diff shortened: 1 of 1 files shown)


harding:
@mruddy cherry-picked and adjusted your commit https://github.com/mruddy/bitcoin.org/commit/9f1885beae3e1f32b2072be2501fb5d6154ee3d8 to this PR.  Please let me know if that looks ok.

Current rendered section:

2015-10-28-114739_619x876_scrot

mruddy:
LGTM.
I'll go close my PR #1102.

harding:
@mruddy thanks!

In the absence of critical feedback, this PR will we merged on Friday.

Issue #1068, opened 2015-09-23
opcode or "op code"?


laanwj:
In the developer reference we refer to "op codes" or "OP codes":

- https://bitcoin.org/en/developer-reference#serialized-blocks
- https://bitcoin.org/en/glossary/op-code

In what I've read before, e.g. assembly manuals, the word was always "opcodes".  Wikipedia seems to confirm this https://en.wikipedia.org/wiki/Opcode same for the bitcoin wiki: https://en.bitcoin.it/wiki/Script

Is there a specific reason to put it as two words?

harding:
The reason that it's two words is that I didn't know it should be one word.  I'll open a PR to fix it as soon as a get a chance.  Thanks!

laanwj:
If everyone agrees this is the way forward I'm willing to make the changes as well

carnesen:
I did a lil Googling on the subject and "opcode" does seem to be more
standard than "op code".

harding:
I think there are some images that need to be updated too, which might be easier for me to do (the images are all generated with GraphViz).  However, if someone wants to do a quick s/// on the _includes directory, we can get that merged quickly and clean up any edge cases later.

harding:
#1107 makes this change.  Thanks for the suggestion!

PR #1107, opened 2015-10-23
Dev docs: change "op code" to "opcode"

harding:
As suggested by @laanwj in #1068

I checked, and I couldn't find any images on the site that used "op code", so this is a pure text replacement (done semi-manually to ensure correctness, not with sed).


jonasschnelli:
ACK. Increase search capabilities and constancy.

laanwj:
ACK, thanks!

harding:
@jonasschnelli @laanwj thanks for the reviews!

In the absence of critical feedback, this pull will be merged on Tuesday (tomorrow).

PR #1110, opened 2015-10-24
Added CCN.la to Resources.html

paulmadore:
CCN  has covered Bitcoin news since the middle of 2013. June, July, or August. Don't recall. Not long after CoinDesk got started.


wbnns:
@paulmadore Thank for submitting this. Looks good to me and am OK w/ adding unless there is consensus otherwise.

wbnns:
Unless others object, this will be merged on Sunday, December 8th.

paulmadore:
And then what happened? Khan came back from retirement to give a directive? What is this nonsense where CoinTelegraph, which has numerous bitcoiners complaining about nonpayment for work completed, is allowed but CCN, which has none of the above except from one single BCH supporter who has not done much with himself since, and whose statements were long ago proven false?

Issue #1106, opened 2015-10-23
Block version 4 will likely be BIP 65, not BIP 62


JustinDrake:
See [here](https://twitter.com/petertoddbtc/status/657563877123735552).

Source File: _includes/devdoc/ref_block_chain.md

This paragraph is outdated:

```
Version 4 blocks will likely be introduced in the near-future as specified in draft BIP62. Possible changes include:
```

harding:
Heh, I should've caught that when I updated the block version 3 text.  Thanks!  I'll update this shortly and see if I can work in a mention of the pending versionbits stuff for future upgrades.  (This is will be PR'd since it isn't a simple correction.)

harding:
@JustinDrake when you have a moment, can you please review #1118 which should correct this?  Thanks again for reporting!

PR #1118, opened 2015-10-26
Dev Docs: update v4 blocks text; mention version bits

harding:
As reported in #1106, our text for v4 blocks had become outdated.  This corrects it and also mentions IsSuperMajority() and version bits as soft fork management methods, although I don't currently have the time to write up a more detailed description of either method.

Preview with this commit applied:

2015-10-26-121035_627x321_scrot


JustinDrake:
Can't spot anything obviously wrong.

harding:
@JustinDrake thanks!

In the absence of critical feedback, this will be merged on Friday.

PR #1121, opened 2015-10-29
Remove Ninki Wallet

Ninkip2p:
With some sadness we are going to discontinue the service from 31st December 2015.

Step 1: Remove links from bitcoin.org and remove Create Wallet functionality from apps.
Step 2: Communicate discontinuation plan with existing users and request move funds to alternative wallet.
Step 3: Discontinue the service and remove apps on 31st December 2015.

Users will still be able to recover after this or at anytime in the future using our recovery tool which will remain on GitHub.


gmaxwell:
Sad to hear anyone closing down; thank you for having an orderly process for this.

The recovery tool doesn't appear to have any copyright/license information. Would it be possible to get a license on it that clearly allows redistribution and modification so that if the github page goes offline it could be archived elsewhere (e.g. on bitcoin.org if needed)?  The recovery tool depends on chain.com, so there is another failure point but at least with the code and the ability to update it, it could be fixed in the future if it breaks, if required.

Ninkip2p:
@gmaxwell Hi, I have added the GNU general public license.

harding:
@Ninkip2p I too am sad to hear about your shutting down, and I second @gmaxwell's appreciation for you doing so in an orderly way.  If there's anything Bitcoin.org can do for you, please let us know.

crwatkins:
@Ninkip2p I'm sorry to hear that you are discontinuing service on your fine wallet.

In a quick test of the current recovery utility, I had a few problems.

1.  I would suggest slightly refining the instructions in the README.md to target a panicked non-technical user with explicit click-by-click steps quoting the exact text of the labels in the utility (e.g. sometime you press next, sometimes you click on a button)

2.  My Chrome wallet balance was 0.0012 while the recovery program offered to restore 0.0002.  Checking the recovery code very quickly, I would have expected 0.0011 after fees.  If this is not what you would expect, I would be glad to debug offline with you.

3.  chain.com rejected the transaction with `{"message":"Transaction rejected.","code":"CH202"}`.  Again, I would glad to debug offline with you.  I'll look at it again myself later today.

greenaddress:
@crwatkins regarding (3) it is probably because the recovery wasn't updated to enforce Low S and new core 0.11.1 rejects all transactions with High S.

I submitted a PR to the repo with a backport that fixes it

https://github.com/Ninkip2p/ninki-recover/pull/1

@Ninkip2p Indeed it is sad to see this fine wallet service being discontinued.

Ninkip2p:
Thanks everyone! I'll spend some time in the coming weeks to get the recovery tool up to a better standard. @greenaddress thanks for the low-s fix. I suspect I will have to add a more sophisticated scanning mechanism for the bip32 chains also.

schildbach:
I wonder if [bitcoinj / wallet-tool](https://github.com/bitcoinj/bitcoinj/blob/master/tools/src/main/resources/org/bitcoinj/tools/wallet-tool-help.txt) could be used for recovery? If Ninki's wallets use BIP32 this should be possible.

crwatkins:
@schildbach Using bitcoinj/wallet-tool  should be possible as I have done it (during the wallet review process) with pybitcointools.  If anyone were to try it, it would be helpful for them to know that Ninki uses a non-standard BIP39 implementation and Ninki uses a separate BIP32 path for each contact.

@greenaddress Great idea regarding (3).   As it turns out, the transaction that was "rejected" by chain.com was actually broadcast and mined.  I don't have the original transaction, so I can't tell you whether A) the problem was indeed high-s and chain.com fixed it for me or if B) the transaction was perfectly fine and chain.com was just screwing with me.  In any case, that is an important fix to get into the recovery utility.

@Ninkip2p regarding issue (2) the problem was that the recovery tool did not properly scan the bip32 chains as you suspected and thus did not find my entire balance.  I increased the gap from 5 (tracked by knodeCounter in your code) to 20 and it found the rest of my balance.

Ninkip2p:
I have merged the low-s fix and also adjusted the gap depth to 50

PR #1122, opened 2015-10-29
Releases: Bitcoin Core 0.10.3

dexX7:
This adds the Bitcoin Core 0.10.3 release notes to bitcoin.org.

I tried to mirror for pull requests for other releases, and for the date I used the date of the release tag. Links were wrapped with `<`, `>`.

The magnet link is missing, and I'm not sure where/how to find the torrent.

### DIFF
--- a/_releases/0.10.3.md
+++ b/_releases/0.10.3.md
@@ -1,0 +1,195 @@
+---
+# This file is licensed under the MIT License (MIT) available on
+# http://opensource.org/licenses/MIT.
+
+## Required value below populates the %v variable (note: % needs to be escaped in YAML if it starts a value)
+required_version: 0.10.3
+## Optional release date.  May be filled in hours/days after a release
+optional_date: 2015-10-14
+## Optional title.  If not set, default is: Bitcoin Core version %v released
+optional_title: Bitcoin Core version %v released
+## Optional magnet link.  To get it, open the torrent in a good BitTorrent client
+## and View Details, or install the transmission-cli Debian/Ubuntu package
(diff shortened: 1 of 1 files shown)


harding:
@dexX7 thanks for doing this!  I think this is the first time we've added the release notes for a release that's not the primary release, so I'm going have have to do some manual checking that everything is working correctly before merging this.  (It should work, @saivann upgraded the site to support this a few months ago, but I want to be safe.)

Regarding the magnet link, that only gets shown for the primary release pages (e.g. /en/download), so it's not used here.  You can leave it blank as you did.

dexX7:
Oh, I see. I just noticed there are no release notes for 0.9.4 or 0.9.5 either. However, I believe it would be nice to have the release notes not only on GitHub, but also listed on https://bitcoin.org/en/version-history, basically for completeness and as reference.

If it works, and if desired, adding the other missing release notes may be done in a follow up.

harding:
I've verified that the various lists and download links all seem to point to the correct places with this commit applied, so this LGTM.  (See screenshots below.)

In the absence of critical feedback, this will be merged on Tuesday.

2015-10-31-215558_493x124_scrot

**Download page still pointing to 0.11.1; I checked several links and they looked good too.**

2015-10-31-215610_449x195_scrot

**Main list of releases; 0.10.3 notes page checked for correct rending**

2015-10-31-215629_690x284_scrot

**List of releases on main Bitcoin Core page also in same order**

saivann:
LGTM 7ec7965

@dexX7 Sure adding missing release notes sounds like a good thing to me. Thanks for your help!

PR #1124, opened 2015-10-29
Alerts: drop banner for UPnP vulnerability

harding:
The red banner at the top of the site went up on Oct 12th.  I suggest we take it down on Sunday, Nov 1st.

The alert notice text will still be available at all the same URLs on the site as before, so anyone who is still catching up on their bitcoin-dev email from two weeks ago will still be able to read it.

CC: @laanwj


saivann:
LGTM

laanwj:
Yes, makes sense

PR #1087, opened 2015-10-12
Enable Google Analytics (2)

Cobra-Bitcoin:
Adds Google Analytics to the site so that visitor numbers and behavior can be accurately measured. This is necessary as Bitcoin.org is currently looking to work with new advertising partners to help fund the site's expenses.

This addresses some of the concerns raised in https://github.com/bitcoin-dot-org/bitcoin.org/pull/1077. The following has been done:
* A formal privacy policy has been written.
* A cookie disclaimer is shown to visitors as per the EU cookie law.
* Visitors that have Do Not Track enabled are not tracked with Google Analytics.

A stripped down version of https://github.com/silktide/cookieconsent2 has been used to display the cookie disclaimer.

If you are personally concerned about Google Analytics being on the site, I recommend you switch on Do Not Track or you install a browser extension that prevents Google Analytics from tracking you.

PR #1136, opened 2015-11-13
Include Support for Selling Ads

Cobra-Bitcoin:
This pull request introduces code that allows Bitcoin.org to show ads which will be located at the top of the site. The ads aren't shown to mobile visitors.

Note: this pull request itself does not contain *any* ads. This just sets up the infrastructure necessary for us to sell ads when we need too.

Technical feedback is appreciated. I'd prefer that we avoid getting into debates about how Bitcoin.org should be funded in this pull request. I don't think that belongs here.


jonasschnelli:
Related to: https://github.com/bitcoin-dot-org/bitcoin.org/pull/1087

ad-auth.txt
```diff
@@ -0,0 +1,3 @@
+Cøbra (https://bitcointalk.org/index.php?action=profile;u=249495) is authorized
+to sell advertising slots for this website (https://bitcoin.org).
+
```
luke-jr: This should not be merged until such authorization has consensus. (Yes, this isn't a technical comment, but neither is this part of the change technical nor merely adding infrastructure.)

jonasschnelli: Agree with @luke-jr

harding:
> I'd prefer that we avoid getting into debates about how Bitcoin.org should be funded in this pull request.

Where can people post to try to change your mind?  I stand by my [earlier comment](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1087#issuecomment-147583676) that I think this is ultimately the domain owners decision (as is all content on this site), but that I also think seeing advertisements makes new people less willing to contribute.  By your not having a debate over the decision and trying to persuade people who doubt this is a good decision, it may also make existing contributors less likely to continue contributing.

On the technical side, I will try to find some time to test this today or on Saturday.

theymos:
I feel like putting it at the top is a bit intrusive. Maybe we should try it in the footer first and see if this generates enough ad income.

There should also be some sort of disclaimer. Currently (unless the ad text will say otherwise), it isn't even clear that it's an ad.

Cobra-Bitcoin:
Thanks for the feedback @theymos. I'll work on changing it to show on the footer, and also add a disclaimer.

PR #1142, opened 2015-11-17
Fix typo

staeff:


MarcoFalke:
Is indeed a typo. utACK 20a77fa

wbnns:
@staeff Thanks for catching this typo. I've fixed this upstream on Transifex so that it will be automatically resolved the next time the German translation strings on the site are updated.

PR #1164, opened 2015-12-17
Add note about future traceability to privacy concerns

anduck:
Note about future traceability.


harding:
This PR adds the following sentence to a related text:

> As blockchain is permanent, it's important to note that something not traceable currently may become trivial to trace in the future.

I think it should say, "As the block chain".  The article "the" seems to be required by the grammar here, and "block chain" with a space is how we spell it on Bitcoin.org for consistency.

Otherwise this looks good to me.

anduck:
Alright. Fixed it.

PR #1180, opened 2015-12-27
Revert "Merge pull request #1178 from Cobra-Bitcoin/remove-coinbase"

hoffmabc:
This reverts commit 7d1cdd94651461ff13ad4ed10b05b2374690fac2, reversing
changes made to cc79a1ee7de9a2a6366898b7ace3a33d9853d24e.

PR #1181, opened 2015-12-28
Remove the wallets supporting Bitcoin XT from "Choose your wallet" page

achow101:
From the [letter that they signed](https://blog.blockchain.com/wp-content/uploads/2015/08/Industry-Block-Size-letter-All-Signed.pdf), Bitpay, Circle, Xapo, and Bitgo are supporting Bitcoin XT. As stated in the letter,
>We[the signers] support the implementation of BIP101

The only implementation of BIP101 is BItcoin XT, and they are also promising to

>support BIP101 in [their] software and systems by December 2015

December 2015 is almost gone, and they have already promised to upgrade to XT.

Since Coinbase was already removed for supporting Bitcoin XT, then these other wallets should be removed since they also support Bitcoin XT, as seen by the letter they all signed.

BTW, Copay is Bitpay's wallet.


theymos:
This was discussed previously, when that letter was issued. We thought that the letter was not clear enough (ie. it's not an absolute *promise* to do anything) and that it was pretty unlikely that they would actually leave Bitcoin for XT. It now seems even less likely that they (most of them) will do so. If more information shows that they have left Bitcoin or have solid plans for doing so, then they should definitely be removed.

As I said then, I do believe that these services were/are acting amazingly irresponsibly in this and other ways, and it may be appropriate to remove them for this reason. But I don't think "they've left Bitcoin" is currently an accurate reason to remove them.

CelestialWalrus:
ACK. They're on XT not on Bitcoin.

ABISprotocol:
Hello, I also believe that Bitpay, Circle, Xapo, and Bitgo have acted irresponsibly, but I don't want to advocate arbitrary removal either.  For a more detailed explanation of what I mean by this, please see: https://github.com/bitcoin-dot-org/bitcoin.org/pull/1180#issuecomment-167734880

luke-jr:
@theymos FYI It was clarified by @coblee on the Coinbase-specific PR (since merged) that Coinbase has also *not* left Bitcoin at this time.

coblee:
@theymos Do you actually think that Coinbase has left Bitcoin?

jgarzik:
What is "amazingly irresponsible" is punishing a bitcoin company for experimenting.  This bitcoin.org action - not Coinbase - has caused end user confusion, with some users asking me if Coinbase BTC is still spendable just like regular BTC.

All this does is further cement the notion that Bitcoin Core and Bitcoin.Org are disconnected from users in the field and what they want and need.

theymos:
@coblee No, I tend to believe you when you say that they haven't.

coblee:
@theymos then why not revert the PR that remove Coinbase wallet?

theymos:
As I said earlier in this thread:

> I do believe that these services were/are acting amazingly irresponsibly in this and other ways, and it may be appropriate to remove them for this reason.

Of the services who signed this statement, Coinbase is probably by far the worst. It has demonstrated its irresponsibility on a number of occasions such as promoting XT (which is suicidal in two ways: it's both a terrible change and a hostile hardfork), holding customer deposits hostage without advance notice, tracking BTC "taint" of customers, etc. Bitcoin.org is not a list of all Bitcoin wallets - it's a list of Bitcoin wallets that people should actually use, and nobody should use Coinbase as a wallet. Therefore, while I tend to think that Coinbase is probably a Bitcoin company right now (based mostly on your word...), I would be opposed to adding it back to the wallets page.

coblee:
> promoting XT 

Coinbase has never promoted XT on our site, blogs, or twitter. Our CEO has expressed his personal views but we have not pushed XT on our users in any way. We are still running Bitcoin even if we have experimented with XT.

> holding customer deposits hostage without advance notice

We don't do that. We let the customer withdraw their BTC and fiat if we close their account.

> tracking BTC "taint" of customers

Being regulated, we have to follow local laws and regulation.

Claiming that nobody should use Coinbase based on the above is a bit extreme. Of course, being a private website, you have the right to promote whatever wallets you want to or not. And I guess we all have a right to not tell new users to go to bitcoin.org because the site is unfairly biased.

coblee:
Also, notice that we did not sign the industry letter last year. And that was on purpose.

luke-jr:
> Bitcoin.org is not a list of all Bitcoin wallets - it's a list of Bitcoin wallets that people should actually use, and nobody should use Coinbase as a wallet.

All webwallets ought to be removed then.

coblee:
> All webwallets ought to be removed then.

I wouldn't be against a unilateral removal of all webwallets if you hold the principles, of users holding their own keys and the freedom of uncensorable bitcoin transactions, important. And any wallets that don't give you that should not be promoted.

But being biased to not list Coinbase, but do list other webwallets is just that: biased.

luke-jr:
@coblee My concern with webwallets is that the security necessary to protect them doesn't exist. If you could give me reason to think Coinbase could protect funds better than MtGox, for example, I would be surprised. But I fear we're going off-topic in this thread, so perhaps if we want to continue discussing not-listing-webwallets it ought to be moved to a new issue or IRC or something.

coblee:
@luke-jr Security is hard to prove. You can only prove you aren't secure when you get hacked. Time with no hacks gives a bit more confidence, but that's obviously not enough. We do have >98% of customers funds in cold storage protected by m/n keys. And our hotwallet (the <2% of coins) is insured. Though no one can be sure the insurance will pay out until it happens... just ask BitPay.

I've always wanted to prove that we do indeed have all the coins that we claim to have, but it's very tricky. We did not previously sign a message with our cold storage keys, so it's very hard to prove we own those addresses without gathering the m/n keys. And doing that is always risky. And we don't want to publicize exactly how many BTC we own and paint a bigger target on our backs. I know there are ways we can show solvency without exposing total amount, but this is not high on our priority list.

coblee:
As for protecting funds better than MtGox, MtGox has bitcoins from cold storage automatically replenish the hot wallet when it's low. That's ridiculous and definitely not something Coinbase does or can do. For the few times it happens (when the hot wallet is low and we need to get coins from cold storage), we first do an investigation to make sure the outflow of coins is legit and not because of a hack. Almost always, it's a user withdrawing a lot of coins maybe to sell at another exchange. And when we have confirmed that it's safe, we have a process to unlock cold storage funds. This requires keys from multiple locations around the world. So it's as safe as we can make it.

Is it fool-proof? Nothing is. But at least I sleep pretty well at night and not worry about it... too much. :)

Azulan:
NACK

Bitcoin.org is no longer a reliable source of bitcoin information. Look to platforms and services not controlled by @theymos and his thugs.

CelestialWalrus:
> We are still running Bitcoin even if we have experimented with XT.

XT is Bitcoin. You've meant Core, @coblee

dabura667:
NACK, especially on removing Copay.

Ok, I get it. Petty arguing >> want to "get back" at them >> hurt their revenue stream >> remove their wallets... makes sense as a way of punishing something you see as wrong. (if not a tad childish)

But Copay is an open source wallet that is making 0 dollars for Bitpay, and they are basically funding its development out of complete charity at this point.

Copay allows the user to access their full node (Bitcoin Core, I may add) to verify all incoming transactions directly with little more effort than setting up a VPS with a full node on it (or your home network if you want to, I guess)

Removing Copay as a recommendation is detrimental to the open source community. (ok, this whole debacle is, who are we kidding... but that aside)

As for the other ones, still just as childish.

paulmadore:
> I wouldn't be against a unilateral removal of all webwallets if you hold the principles, of users holding their own keys and the freedom of uncensorable bitcoin transactions, important.

This is what should be done.

lichtamberg:
Can you please stop to act so childish? The most important thing on bitcoin.org should be to teach new users how to use bitcoin safely. Coinbase and Copay do a great work in this area and the question about xt or not is really displaced here. please just stop it.

gregnie:
NACK

Bitcoin.org is no longer a reliable source of bitcoin information.

gappleto97:
NACK

Bitcoin.org is no longer a reliable source of bitcoin information.

Even if it were, it's hypocritical of you to remove Coinbase and not BitPay, who have are working on their own client that people here would call a "hostile hardfork" (despite the fact that, like XT, it has safeguards to keep it from hard forking unless there's broad support for it).

hrishikeshio:
NACK

Bitcoin.org is not a reliable source of bitcoin information anymore.

harding:
Closing per https://github.com/bitcoin-dot-org/bitcoin.org/pull/1181#issuecomment-167465906

PR #1186, opened 2015-12-28
Add disclaimer to bitcoin-core index page

btcdrak:
There is some confusion in the community that Bitcoin Core and bitcoin.org are the same project
and thus share the same policy and views. This disclaimer is to make it clear that Bitcoin Core is
a separate project and just occupy /bitcoin-core on the website for the main purpose of hosting
downloads.

PR #1187, opened 2015-12-29
Core: mention separation of concerns from Bitcoin.org

harding:
This commit makes two changes.  First it modifies the footer on the Bitcoin Core pages:

2015-12-29-012709_611x97_scrot

This drops the [Press](https://bitcoin.org/en/press) page and the [Blog](https://bitcoin.org/en/blog) page and changes the About link to the page described below.  Beneath the copyright statement, a short message tells readers that the Bitcoin Core pages are not a normal part of the site; the link here also points to the page described below.

The second change is a new page that describes the relationship between Bitcoin Core and Bitcoin.org in detail:

2015-12-29-012731_655x900_scrot

All the text in both changes is just draft text, so please feel free to make suggestions.  In addition, I'm open to combing this with #1186 if a more visible disclaimer is desired for the main page.  What I tried to aim for with this PR is a short and unobtrusive disclaimer on every Bitcoin Core page as well as a separate page where we could go into depth about the separation between the two projects.

Related to this PR, I also created a new tag for the site for Bitcoin Core issues to help people (like me) who want to focus on them: https://github.com/bitcoin-dot-org/bitcoin.org/labels/Core

### DIFF
--- a/_includes/layout/base-core/footer-license.html
+++ b/_includes/layout/base-core/footer-license.html
@@ -1,0 +1,10 @@
+{% comment %}
+This file is licensed under the MIT License (MIT) available on
+http://opensource.org/licenses/MIT.
+{% endcomment %}
+
+<div class="footerlicense">© Bitcoin Project 2009-{{ site.time | date: '%Y' }} {% translate footer layout %}<br>
+Bitcoin Core pages on Bitcoin.org are <a
+href="/en/bitcoin-core/about-site">maintained separately</a> from the
+rest of the site.
+</div>
--- a/_includes/layout/base-core/footer-menu.html
+++ b/_includes/layout/base-core/footer-menu.html
@@ -1,0 +1,16 @@
+{% comment %}
+This file is licensed under the MIT License (MIT) available on
+http://opensource.org/licenses/MIT.
+{% endcomment %}
+
+<div class="footermenu">
+  <a href="/en/alerts" class="statusmenu{% if site.STATUS == 1 %} alert{% endif %}">Network Status</a>
+  <a href="/{{ page.lang }}/{% translate legal url %}">{% translate menu-legal layout %}</a>
+  {% case page.lang %}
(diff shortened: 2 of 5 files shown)


jonasschnelli:
Concept ACK. Thanks @harding!

Nevertheless the separation is only virtual because the power/centralization is not really separated. The domain/site owners can always change the sources of the Core pages. But okay and somehow unavoidable IMO

btcdrak:
ACK

saivann:
utACK, would be nice to have +1 from domain owners @theymos @Cobra-Bitcoin .

CodeShark:
ACK

kanzure:
ACK, this looks like a reasonable thing to mention on the site.

instagibbs:
concept ACK

morcos:
Yes I like this, but agree with @saivann that I'd like to know that the domain owners see it this way.

eragmus:
Concept ACK.

theymos:
ACK

harding:
Thanks everyone.  Scheduling this for merge late tomorrow (Wednesday).

luke-jr:
No offence intended, but I don't think that it's relevant whether or not the domain owners see it this way...

saivann:
@luke-jr Domain owners have been more active lately and doing more maintainer-level decision making, and technically speaking, they have control over the domain, which is the highest level of control someone can have over bitcoin.org . Thus, making sure they agree with that separation, especially given that it's only a symbolic one, basically just some general agreement, seems rather important to me.

jrmithdobbs:
Agreed @saivann. I think merge should be postponed until the domain owners chime in.

harding:
@jrmithdobbs: One of the owners ACK'd [here](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1187#issuecomment-167902926), which is why I scheduled the merge.

jrmithdobbs:
Missed him chiming in. My bad.

laanwj:
ACK, sounds good to me

PR #1188, opened 2015-12-29
Updated requirements in full-node.md

Jorgeminator:
Updated disk space and bandwidth requirements due to blockchain growth.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -209,21 +209,21 @@ kramdown:
     coderay_css: style
 
 text:
-  ## Values last updated 2015-08-26
+  ## Values last updated 2015-12-29
   ## All variable names must indicate unit type for easy translation of adjacent text,
   ##   such as: subsidy_in_decimal_bitcoins or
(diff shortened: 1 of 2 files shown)


MarcoFalke:
ACK bc8ad6e

jgarzik:
ACK as written, though my preference would be 80 G.

harding:
PR updated to use site-wide variables for chain size (initial download) and disk space needed (higher than strictly necessary to allow for moderate growth).  For the disk space needed, this was bumped up to 80G per @jgarzik's suggestion.  Preview:



We can do some rewriting here when 0.12 is released and it becomes possible to use pruning without disabling the wallet.

Since the current values are out of date and the general feedback seems positive, I'm marking this as merge scheduled for Thursday.  Please comment if you have unaddressed concerns.

MarcoFalke:
> We can do some rewriting here when 0.12 is released and it becomes possible to use pruning without disabling the wallet.

Also there is `-maxuploadtarget` in `0.12`.

PR #1191, opened 2016-01-05
Dev Docs: clarify null data conditions

harding:
PR #1129 pointed out that the null data transaction details were out of date.  Other comments on that PR asked that the tx type be better explained.  I submitted a PR to the repository of that PR which provided what I think was a better explanation, but I haven't heard from that author in two months, so I'm submitting it separately here.

### DIFF
--- a/_data/glossary/en/null-data-transaction.yaml
+++ b/_data/glossary/en/null-data-transaction.yaml
@@ -7,8 +7,10 @@ required:
     title_max_40_characters_no_formatting: Null Data (OP_RETURN) Transaction
 
     summary_max_255_characters_no_formatting: >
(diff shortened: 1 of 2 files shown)


jaromil:
How about referencing the term "null-data-transaction" introduced in the glossary also in the description of the OP_RETURN? the term is not named in the latter.

harding:
@jaromil are you talking about something in this diff, or something not changed on the site?  Looking at the diff, it seems every line in the added text is within 7 or fewer lines of the name "null data".

jaromil:
Yes the diff, but nvm my bad, I see now it is in all caps in the title, I was just looking at the description.

Issue #1147, opened 2015-11-29
Create template for developer weekly IRC meetings


btcdrak:
@G1lius have been kindly writing summaries of the weekly IRC dev meetings and posting them to various subreddits and the mailing list to help assist general users be aware of what Bitcoin Core developers are working on.

[example](https://www.reddit.com/r/Bitcoin/comments/3tyfaa/bitcoin_dev_irc_meeting_in_laymans_terms_20151119/)

I suggested to @harding we could create a place on bitcoin.org where @G1lius could submit a PR and have them published on bitcoin.org.

I'm opening this issue so it can be tracked.

harding:
The version of Jekyll we use to build the site only supports one blog, which we currently use for site news.  The newest version supports multiple blogs, and I already have an patch that @saivann and I have played around with to upgrade us to that version.  It's a fairly big change, so I'll try to PR that individually in the next few days; once it looks like that's going to be merged, I'll work on creating a section for the weekly meeting notes and maybe other similar info.

@btcdrak good suggestion, thanks!  @G1lius thanks for making the notes, I've enjoyed reading them!

btcdrak:
Closing. See https://github.com/bitcoin-core/website/issues/18#issuecomment-171316814

harding:
Reopening.  I continue to think @G1lius's meeting summaries would make a good addition to Bitcoin.org and I just recently in PR #1193 provided the necessary changes to begin support hosting them.

luke-jr:
ACK reopening. The meetings *are not* specifically about Bitcoin Core.

btcdrak:
I am agnostic toward reopening the ticket but @luke-jr's statement is not accurate. Every IRC meeting is is attended by contributors to Bitcoin Core where the meeting topics cover Bitcoin Core pull-requests, issues and release schedules as all the previous meeting notes will corroborate. While the meetings are open, they are very much about the Bitcoin Core project's ongoing work.

harding:
I (like Luke) am also under the impression that the meetings are not specifically about Bitcoin Core.  I agree that they tend to be dominated by Bitcoin Core concerns, but I think that's only natural because Bitcoin Core is doing so much highly-innovative-but-review-needing work and because so many Bitcoin Core developers are already hard core IRC users.

---

[...]

PR #1193, opened 2016-01-07
Upgrade to Jekyll 3.x

harding:
Update to Jekyll 3.x

This commit updates the site from Jekyll 1.3.0 to 3.0.1 (anagram unintended).  This catches us up on a couple years of technical debt from not upgrading and makes it easier to fulfill feature requests like #1147.

A diff of the HTML compared to the current site is provided below. I'm sorry it's so large: as part of the upgrade, the syntax highlighter was changed from Pygments to Rouge and so every code span and block on the site was changed.  (And, gosh, we had a lot of 'em.)

There should be no changes to any javascript, CSS, or links.  I think there's only one user-visible change to the site: the Bitcoin Core releases are now shown in date order rather than version order (but the highest version number is still the version offered for download on the Download page).

- Preview site: http://dg0.dtrt.org/
- Full HTML diff: http://dg0.dtrt.org/diff.html

Explanations of all the changes in the commit are available in the commit message.  Sorry it's one giant commit---everything seemed to be interdependent.  Site developers should be able update to the latest Ruby gems after checking out this branch by running, `bundle install`

I'm aiming to merge this early morning on Saturday the 16th. (I want to merge on a Saturday so I'm available to fix any upgrade issues.)


btcdrak:
ACK

harding:
Note to owners/other maintainers: I'm beginning the upgrade now; please don't make any other merges until I post here that I'm done.  If you need me, I'll be checking IRC periodically.

harding:
Upgrade complete.  I watched the build and didn't see any problems in the logs, and I've spot checked a few dozen pages in several different languages without seeing any problems.  If you do spot a problem, please post here or open a new issue.  Thanks.

PR #1200, opened 2016-01-10
Making clear that segwit is equivalent to a block size increase

NicolasDorier:
Segwit is a block size increase, lots of readers of this FAQ care more about the block size than all the goodness of segwit.
So I think this should be clearer.

### DIFF
--- a/en/bitcoin-core/capacity-increases-faq.md
+++ b/en/bitcoin-core/capacity-increases-faq.md
@@ -29,7 +29,7 @@ specific improvements described in the [roadmap][].
 | Feb 2016 | 0.12.0 | [libsecp256k1 verification][] |
 | Feb 2016 | | Segregated witness feature complete & ready for general review |
 | Mar 2016\* | 0.12.x | Deploy OP_CHECKSEQUENCEVERIFY (BIPs [68][BIP68] & [112][BIP112]) + [BIP113][] as first [BIP9][] versionbits soft fork |
-| April 2016\* |  0.12.x |  Deploy segregated witness |
+| April 2016\* |  0.12.x |  Segregated witness deployment including [block size increase](#segwit-size) |
 | 2016 | | Weak blocks, IBLTs, or both |
 
 \* Dates with an asterisk are when we expect to release soft fork-ready code. The code will not be released until it has been well reviewed, and the actual fork will take time to activate ([BIP66][] activated in July 2015 after a few months; [BIP65][] activated in Dec 2015 after only 5 weeks).


harding:
I agree mentioning the effective block size increase here would be good, but I suggest that we leave the amount unspecified (since it varies depending upon usage) and link to the [answer on that page](http://bco-master/en/bitcoin-core/capacity-increases-faq#segwit-size) which goes into more detail.  My suggestion:

    | April 2016\* |  0.12.x |  Segregated witness deployment including [block size increase](#segwit-size) |

NicolasDorier:
I think there is a kind of pavlovian reaction concerning the three words "Block size increase", which is why I think it is more important for them to appear before segwit. (Even if I agree though that segwit is more exciting than the increase in reality)

harding:
Putting "Block size increase" at the beginning of the line makes me feel like we're promising an increase in April, when I conceive of this line as saying that Bitcoin Core will included the necessary segwit code in April and that it will then be up to miners to signal their readiness and activate the actual soft fork.

I know this is splitting hairs, and I apologize, but I don't want to over-promise.  I'm hoping that putting "block size increase" in blue (as a link) will be almost as good at drawing reader attention as putting it at the beginning of the line.

What do you think?

jgarzik:
Block size increase via segregated witness is economically different from increasing core block size.

Please don't conflate the two.  That is factually incorrect.

harding:
@jgarzik that's why I suggested adding a link to the detailed answer: https://bitcoin.org/en/bitcoin-core/capacity-increases-faq#segwit-size

NicolasDorier:
> Block size increase via segregated witness is economically different from increasing core block size

@jgarzik can you explain ? I agree that for profiting of the increase in segwit, we need segwit adoption in wallets, is that what you mean by "economically different ?". If so I agree.
But if segwit get adopted in wallets, you effectively have the same economical result than a core block size bump.

NicolasDorier:
@harding I updated the patch to use your proposition. I'm not convinced by it though, but that's better than what there is now.

harding:
@NicolasDorier thanks!  ACK a272cb5

Aquentus:
@NicolasDorier SW may have the benefit of a blocksize increase, but it is not a blocksize increase proposal nor has it been presented as such. The blocksize increase is a side benefit. 

If instead you wish to present it as mainly a blocksize increase proposal via a complicated softfork, that is perfectly fine, but that is not what has been stated so far. 

Moreover, with core now having its own website, this content should be moved out of bitcoin.org to bitcoin core's website, thus making any such update pointless.

As such I lean towards a NACK.

NicolasDorier:
@Aquentus the fact that the blocksize increase is a side benefit does not mean it should not be stated. Lots people coming to this FAQ are pissed of because they think nothing is done to improve the block size. Which is wrong, this is a segwit is a block size. Not the "Core block size" but a block size increase nevertheless, which increase max tx rate almost as much as a dumb bump.

Azulan:
NACK

While bitcoin.org is no longer a reliable source of bitcoin information, continuing down that path with outright deception would be even worse. Walk up the stairs with your feet, not your hands.

NicolasDorier:
@Azulan can you tell why segwit is not a block size increase ?

jl2012:
@jgarzik Any segwit user will immediately get a ~50% discount of their fee. The only reason for not switching to segwit is the absolute fee discount is still too low and therefore lack of incentive to upgrade

instagibbs:
ACK. It is an increase, and in fact provides a smooth(er) ramp as the non-segregated portion fills up.

petertodd:
ACK https://github.com/NicolasDorier/bitcoin.org/commit/a272cb5a5479dda7d1dc404677019adfe83db58c

That's a fair description.

harding:
I'm going to merge this with some other stuff later today, but this and #1202 will be the last content updates of this page I plan to make on Bitcoin.org.  Future PRs regarding the FAQ should be sent to https://github.com/bitcoin-core/website

Thanks everyone for your comments

luke-jr:
@jl2012 Fees are miner policy, not something determined by the consensus rules. So it's not appropriate for us to talk about discounting fees at all.

Azulan:
@NicolasDorier Creative accounting is the basis of the legacy financial system. It's not something we should enable and advocate for bitcoin.

PR #1202, opened 2016-01-10
Capacity Increases FAQ: more prominently link to roadmap

harding:
I'm told some people were confusing the FAQ itself for the roadmap; this should help remove the confusion.  This becomes the first question/answer in the document:

2016-01-10-104013_634x193_scrot

### DIFF
--- a/en/bitcoin-core/capacity-increases-faq.md
+++ b/en/bitcoin-core/capacity-increases-faq.md
@@ -19,6 +19,14 @@ Other versions: [Italiano](/it/bitcoin-core/capacity-increases-faq) \| [简体�
 1. toc
 {:toc}
 
+## What is the roadmap? {#roadmap}
+
+**[Capacity increases for the Bitcoin system][roadmap]**, written by Gregory
+Maxwell and published 7 Dec 2015 to the bitcoin-dev mailing list. A
+[statement][] supporting the roadmap has been signed by developers who
+([according to GitHub][commit stats]) generated more than [90% of all
+commits][commit spreadsheet] to Bitcoin Core in 2015.
+
 ## What specific technologies are included in the roadmap, and when can we expect them?  {#roadmap-dates}
 
 New technology will be deployed when it is ready and has been tested.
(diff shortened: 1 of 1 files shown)


Aquentus:
NACK. Bitcoin core is getting it's own website where they can set their roadmap. Bitcoin.org should be a neutral site explaining to newbies the bitcoin ecosystem in the least confusing way possible.

As you know, individuals who already participate in bitcoin, and therefore run nodes or businesses or mine etc, don't really visit bitcoin.org because they already know much of the information it contains. You are not therefore able to influence any opinion except for confusing newbies who probably don't understand any of the statements made in that roadmap.

Bitcoin is changing and that is for the better. While previously there was only one team of developers who could have been coerced or corrupted, different teams are now forming, all with the best interest of bitcoin in mind and all teams share a lot in common, in the process drastically increasing bitcoin's resilience. 

Bitcoin.org should focus on making things easy and simple for newbies to understand, and if that includes multiple implementations then explain that too. Not fool itself in thinking that anyone who has any influence cares whatever about what it contains.

kanzure:
ACK, because the FAQ previously linked to the roadmap, the FAQ was written regarding that roadmap, and I have noticed users confused about "what roadmap", even though the FAQ already linked to the roadmap in at least one other location.

At most I could support a NACK on mentioning the number of git commits. It's useful context, and I wish I had a better way to phrase that, but I don't. For this reason I am OK with leaving it in (or even removing it) at the moment.

@aquentus, you have made a factually false statement (that "there was previously only one team with a seemingly working Bitcoin implementation"). This github issue thread is not a good place to try to rewrite history. Sorry.

Your NACK sounds more like a NACK regarding hosting any content about Bitcoin Core at all. This pull request is about a particular change to some existing content about Bitcoin Core.

Aquentus:
Thanks Kanzure. I have refined my Nack to point out that there is no point in updating the content when core is in the process of launching its own website where they will host their own content and describe the road map and their philosophy.

Further thanks @kanzure for the nits in regards to different implementations. I have updated my post to make it clear that my intention was to state one "main" team which defacto acted for all purposes as the only team in having some 99% use it's client.

That is changing however and that is for the better. The system is now becoming more robust from outside or inside attacks where a corrupt individual or group of individuals won't be able to cause much damage. It is the same shit as from just one exchange - mtgox - to the multiple exchanges we have today.

The main point however, you are correct to say @kanzure, is that bitcoin.org should be neutral and aim to explain in the clearest way possible what bitcoin is and the bitcoin ecosystem to newbies, and leave roadmaps to client focused websites.

harding:
I'm going to merge this with some other stuff later today, but this and #1200 will be the last content updates of this page I plan to make on Bitcoin.org. Future PRs regarding the FAQ should be sent to https://github.com/bitcoin-core/website

PR #1205, opened 2016-01-12
Move Bitcoin Core project statements

btcdrak:
The Bitcoin Core project now has it's own official website.


Aquentus:
You'll need to rebase I think? Otherwise ACK.

Bitcoin is changing to a world of multiple implementations which will make it more robust from internal or external attacks. As such, bitcoin.org, as an entry point for newbies, should try and explain the ecosystem in the clearest and simplest way possible, with no bias whatever, so that new entrants can fully understand bitcoin.

Cobra-Bitcoin:
Maybe you should use bitcoincore.org as a domain instead of bitcoinco.re? It looks much better. I think @gavinandresen owns bitcoincore.org at the moment. I don't think he'd be too opposed to transferring it to @laanwj.

It looks like this new website doesn't clone all the Bitcoin Core specific content that @harding worked so hard to create. It also isn't appearing in any search results for the term "bitcoin core". The site needs more content and some time to establish itself, because at the moment it looks very unprofessional and bare-bones.

I also don't agree with how you're redirecting parts of bitcoin.org to this new site with redirects. Rather than redirects to strange new sites, which are sudden and not what the user was expecting, it's better to edit the original statement page with something like "This statement has been moved to http://bitcoinco.re/statement_link_here". That way the user can decide for themselves to navigate to a new site. The site's use of CloudFlare was annoying too, especially for a Tor user such as myself.

_config.yml
```diff
@@ -73,6 +73,11 @@ redirects:
   /ar/إدعم-ألبت-كوين: /ar/support-bitcoin
   /ar/مفردات: /ar/vocabulary
   /ar/تحتاج-أن-تعلم: /ar/you-need-to-know
+  /en/bitcoin-core/2016-01-07-statement: https://bitcoinco.re/en/2016/01/07/statement
+  /en/bitcoin-core/capacity-increase: https://bitcoinco.re/en/2015/12/21/capacity-increase
+  /en/bitcoin-core/capacity-increase-faq: https://bitcoinco.re/en/2015/12/23/capacity-increases-faq
```
harding: This is capacity-increase**s**-faq with a s.  All the translated versions are singular.  (Sorry for the inconsistency.)

harding:
I like the idea of adding links to the top of the statements pointing to the new site rather than using redirects.  I'm not totally opposed to using redirects to off-site content, but I agree with @Cobra-Bitcoin that's not what the user expects, and I think we should maybe save that for at least a few months down the line when the new site is much better than the existing subsite.

I don't use Tor for browsing, so I'm not sure what this CloudFlare issue is, but if the new site is not usable / easy to use on Tor, I consider that an important bug.

(Side note: I am going to work on porting over the existing content.  Hopefully I can find some time for that this weekend.)

ACK on adding a link to the new website from the bitcoin.org/en/bitcoin-core/ page.

btcdrak:
When content has moved, it has moved, otherwise you are creating duplicate content which is bad for SERPs. The correct action is to notify by use of redirect responses. The better option would be to adapt the "redirects" plugin to do redirects by displaying a page with meta tags refresh with a 301 redirect and a delay. That way the user is notified "content has moved and will be redirected in 5 seconds (otherwise click here)".

harding:
I was thinking we could maybe add a banner to the Bitcoin Core pages similar to the banner we use for alerts (but not red/yellow like alerts) that points people to the new pages.

As for duplicated content, I think it's sad that can hurt SEO but I would hope that if one version of the content links to an alternative version of basically the same content, that would give search engines a hint about which one is the canonical content.

btcdrak:
@harding Well if you wont move the content, put meta-refresh in the HTML header at least, but I am unsure that helps the SEO case since 301 HTTP headers are not emitted.

`<META http-equiv="refresh" content="10;URL=http://example.com/page">`

harding:
@btcdrak I'm saying that I like @Cobra-Bitcoin's idea of people actually clicking a link to go to a new site, at least for the moment while our subsite is still useful and up-to-date.

jonasschnelli:
ACK

wbnns:
Hi all, closing this for now since it's over a year since the last discussion (a banner was added) and there are a significant number of merge conflicts.

PR #1209, opened 2016-01-14
Updated to use site variables for blockchain size

Jorgeminator:


harding:
Thank you very much for opening this PR.  Unfortunately, the site variables can't be used in the translation YAML, as seen in the screenshot below.

2016-02-10-083251_890x181_scrot

PR #1211, opened 2016-01-17
Bitcoin Core: add banner redirecting some pages to BitcoinCore.org

harding:
This is an alternative to #1205 that adds banners to pages that are now mirrored on BitcoinCore.org.  The banners look like this and redirect to the specific page:

2016-01-16-193839_1220x418_scrot

In addition, a link is added to the News on the main Bitcoin.org Bitcoin Core page pointing to the new website:

2016-01-16-194303_584x146_scrot

Full preview: http://dg2.dtrt.org/en/bitcoin-core/

### DIFF
--- a/_includes/layout/base-core/pagetop-moved-notice.html
+++ b/_includes/layout/base-core/pagetop-moved-notice.html
@@ -1,0 +1,10 @@
+{% comment %}
+This file is licensed under the MIT License (MIT) available on
+http://opensource.org/licenses/MIT.
+{% endcomment %}
+
+{% if page.moved_url %}
+<div class="banner-message success">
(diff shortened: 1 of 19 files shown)


btcdrak:
I would suggest a different colour, maybe yellow, orange or red.

harding:
@btcdrak this good?

2016-01-16-223953_1098x330_scrot

(I don't want to use orange or red since we use them for network alerts)

btcdrak:
@harding, yes, it's much clearer now thanks!

FWIW I still object to not having redirects since the content has moved, but ACK on this PR anyway.

harding:
Updated PR and live preview with yellow color.  If there are no objections, I'll merge in a few hours.

PR #1224, opened 2016-01-23
Remove givebtc.org from resources

cpmx:
On the resources page (https://bitcoin.org/en/resources) directories section the link for givebtc.org redirects to http://bitcoinembassy.ca


harding:
@FrancisPouliot is it ok to remove this link, or is there somewhere else we should redirect to?

FrancisPouliot:
yes, site was taken offline. Please remove. I forgot to notify

PR #1225, opened 2016-01-23
Updating Blockchain size from "20GB" to "65GB".

jonathancross:
Accounts for all variations in the 27 translations.

FYI: Here is the shell commands used to update (may come in handy next time):

    cd bitcoin.org/_translations
    for F in *;do /usr/bin/perl -p -i -e "s/20GB/65GB/g" $F;done
    for F in *;do /usr/bin/perl -p -i -e "s/20 GB/65 GB/g" $F;done
    for F in *;do /usr/bin/perl -p -i -e "s/20 /65 /g" $F;done
    for F in *;do /usr/bin/perl -p -i -e "s/20ГБ/65ГБ/g" $F;done
    for F in *;do /usr/bin/perl -p -i -e "s/20ＧＢ/65GB/g" $F;done

Yes, it *could* have been simplified, but it took care of all crazy variations with no mistaken matches  ;-)


greenaddress:
ACK

FelixWeis:
Since the block chain size is constantly growing, it would be good to include a date every time it is mentioned so "[...] storage for the full block chain size (over 65GB, **as of January 2016**) [...]"

see also https://en.wikipedia.org/wiki/Wikipedia:As_of

jonathancross:
@FelixWeis That would in fact be nice, but unfortunately it would require many new translations.
In #1234 there is a suggestion of trying to make the number a variable that can be changed.

jonathancross:
Hi all.
There is much we could do to make it more dynamic, add additional info, etc. but the current info is just wrong and misleading.

Any reason not to merge this while discussing additional improvments?

MarcoFalke:
Agree with @jonathancross and @laanwj: This should be merged soon as it does not seem anyone came up with a dynamic solution in the last months.

Ping @harding

harding:
Untested ACK.

Issue #1248, opened 2016-02-27
Developer Glossary Definition: Double Spend


VincentPiper:
Currently, Bitcoin.org's double-spend definition states: "A transaction that spends the same input as spent in another transaction.".

The original meaning before Bitcoin existed was something to the effect of: "a failure mode of digital cash schemes, when it is possible to spend a single digital token twice." (https://en.wikipedia.org/wiki/Double-spending) (In the Whitepaper, Satoshi refers in this meaning of double-spend). Ultimately, the creation of the blockchain made double-spending by this definition, impossible. Yet, the community still uses this term but in the new defined way.

The important term in both definitions is the word "spend" or "spent", but they are used in different meanings. In Bitcoin's Double-Spend Definition, it seems to refer to and in theory is interchangeable with just a "push" or a "broadcast" and does not seem to convey whether it is in a block or not. It actually seems to convey a state of limbo, like pending in the mempool.  But before Bitcoin, double-spend's definition seems to refer to an actual finalized pay out or disbursement and the doubling of that coin. (In addition, the terms are also used in the context of being recorded in the blockchain, like when a block explorer shows outputs that are "spent" or are "unspent".) There seems to be no single definition of "spend" or "spent", when it is to be used, and for what purpose. The reader is left to interpret the intended meaning.

I request that we change the name of the term of double-spend, to something else. I would propose something like a "double-push" or a "double-cast", so to differentiate from the original meaning. (I would prefer this option and would vote for "double-cast".)

But I also understand that it may not be possible since everyone is most likely set in their ways of Bitcoin terminology. (In addition, I believe I am filing this request in the proper venue, since Developers and Users would defer to the Bitcoin.org’s definitions as official or more official than others.) In that case, I request that the Developers Glossary Definition of Double Spend be made more specific so that it would say something as to the following:

**_Double Spend: "The transaction that used the same input as an already broadcasted transaction. This attempt of duplication_**[1]**_, deceit_**[2]**_, or conversion_**[3]**_, will be adjudicated when only one of the transactions is included in the blockchain."_**

[1] duplication, as in regards to double spend's original definition, as used in the Whitepaper.
[2] deceit, as in regards to a double spend against merchants or citizens.
[3] conversion, is my made up term for legitimate double spending, like in Opt-In-RBF cases.

Thank you for your consideration.

MarcoFalke:
[...]

PR #1251, opened 2016-02-29
Update to Glossary Definition: Double Spend

VincentPiper:
Please see the following issue for context: #1248 
TLDR: I believe the definition for "double-spend" should not contain the words "spend" or "spent" 
and should be more specific and clear to differentiate from the pre-Bitcoin definition of double-spend.

I am not proficient in Github or programming in general, so hopefully this pull request is 
appropriate and done correctly. If this is successful, I will also attempt to create a new glossary term
for "spend/spent" and others that may be helpful.

### DIFF
--- a/_data/glossary/en/double-spend.yaml
+++ b/_data/glossary/en/double-spend.yaml
@@ -7,8 +7,10 @@ required:
     title_max_40_characters_no_formatting: Double Spend
 
     summary_max_255_characters_no_formatting: >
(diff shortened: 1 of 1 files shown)


wbnns:
Unless others object, this will be squashed and merged on Sunday, December 18th.

@VincentPiper Thank you for taking the time to articulate your perspective in #1248 and for learning how to submit a pull request to make the change. Also, for fixing the typo you found.

PR #1256, opened 2016-03-08
Fix incorrect link to segwit bip

ikeboy:
Link to bip is dead, Core version updated the link in https://github.com/bitcoin-core/bitcoincore.org/commit/7bd3ad9484802dd9a623e5586f8a2da3f92a2cd6 but this page still has the dead link.


MarcoFalke:
Concept ACK

harding:
ACK (untested, but I'm highly confident this is correct)

PR #1261, opened 2016-03-14
Clarify solo mining version applicability

cloudnthings:
Minor clarification that getwork is only applicable for v0.9.5 and below.


harding:
ACK.

luke-jr:
Probably should just remove mention of getwork...

cloudnthings:
True. Either or. I'm not fussed as it is a very minor point.

PR #1277, opened 2016-04-09
add HSTS preload requirement for wallets

jameshilliard:
Requiring HSTS preloading reduces the risk of MITM attacks for users who have not previously visited a particular site before and is especially useful for preventing downgrade attacks on TOR users or users who use browsers without persistent storage. bitcoin.org itself currently doesn't use HSTS preloading and it would be a good idea to enable it there as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -719,7 +719,7 @@ Basic requirements:
 - No concerning bug is found when testing the wallet
 - Website supports HTTPS and 301 redirects HTTP requests
 - SSL certificate passes [Qualys SSL Labs SSL test](https://www.ssllabs.com/ssltest/)
-- Website serving executable code or requiring authentication uses HSTS with a max-age of at least 180 days and is included in the [HSTS preload list](https://hstspreload.appspot.com/)
+- Website serving executable code or requiring authentication uses HSTS with a max-age of at least 180 days
 - The identity of CEOs and/or developers is public
 - Avoid address reuse by using a new change address for each transaction
 - If private keys or encryption keys are stored online:
@@ -751,6 +751,7 @@ Optional criteria (some could become requirements):
 - Does not show "received from" Bitcoin addresses in the UI
 - Uses deterministic ECDSA nonces (RFC 6979)
 - Provides a bug reporting policy on the website
+- Website serving executable code or requiring authentication is included in the [HSTS preload list](https://hstspreload.appspot.com/)
 - If user has no access over its private keys:
   - Full reserve audit(s)
   - Insurance(s) against failures on their side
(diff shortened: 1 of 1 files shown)


jameshilliard:
Some of the web wallets listed already enable this:
Coinbase: enabled
Bitgo: enabled
Coinapult: enabled
Coinkite: not enabled
Greenaddress: not enabled
Xapo: not enabled
Circle: not enabled

Cobra-Bitcoin:
Thanks for bringing this to my attention. I'll look into enabling HSTS preloading for bitcoin.org (we should at least do it ourselves before making it a requirement for any wallets).

crwatkins:
I think this is a good idea and I believe that it continues our (sometimes tedious) HSTS campaign which is a good thing.  Thank you very much for submitting it, and thanks for reviewing the existing listings.

In the past, when we have introduced new requirements that some of our existing listings did not meet, we have placed them in our optional criteria section, strongly encouraging the criteria for new listings and requesting that existing listings update their offerings.  A good example is that we initially introduced address rotation as optional and then moved it to required.  It may be interesting to note that in that case, we did not list any new wallets that did not meet the criteria while it was it was in the optional section.

@jameshilliard, I would recommend moving the new preload criteria to the optional section for the time being and later upgrade it to required.  This will allow us to officially recommend it to our existing listings while giving them adequate time to research the implementation and any possible implications.

jameshilliard:
@Cobra-Bitcoin Any update on getting HSTS preloading enabled for bitcoin.org in light of [this](https://bitcoin.org/en/alert/2016-08-17-binary-safety)?

README.md
```diff
@@ -719,7 +719,7 @@ Basic requirements:
 - No concerning bug is found when testing the wallet
 - Website supports HTTPS and 301 redirects HTTP requests
 - SSL certificate passes [Qualys SSL Labs SSL test](https://www.ssllabs.com/ssltest/)
-- Website serving executable code or requiring authentication uses HSTS with a max-age of at least 180 days
+- Website serving executable code or requiring authentication uses HSTS with a max-age of at least 180 days and is included in the [HSTS preload list](https://hstspreload.appspot.com/)
```
wbnns: @jameshilliard Hey, sorry that this PR hasn't been resolved just yet. Can you please move this to the Optional section as per the [comment](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1277#issuecomment-231561648) @crwatkins left?

jameshilliard:
@wbnns I've moved HSTS preloading to the optional section

crwatkins:
Thanks @jameshilliard.  LGTM.

wbnns:
Unless others object, this will be merged on Wednesday, December 21st.

@jameshilliard Thank you, sir!

PR #1280, opened 2016-04-14
Adding a new bitcoin exchange directory.

bitcoindirectory:


Cobra-Bitcoin:
This site needs more time to establish itself before we consider linking to it.

PR #1293, opened 2016-05-04
Example formating

uberj:
Add in commas and comments to match other example send() calls


harding:
ACK.  (Don't know how I missed that; thanks!)

PR #1299, opened 2016-05-18
Update "Null Data" section: Core 0.12.0 released

Flowdalic:


Flowdalic:
Basically the same as #1299, but without keeping the "It is expected", since it doesn't make any sense (at least to me).

PR #1300, opened 2016-05-18
improve wording

g4p:
The wording "almost just like with physical money" is awkward. Removed the extraneous "just": "almost like with physical money"


cloudnthings:
ack.

PR #1304, opened 2016-05-25
Add NBitcoin to developer documentation

nopara73:


harding:
1. I think the link should point [here](https://programmingblockchain.gitbooks.io/programmingblockchain/content/); it took me a bit to figure out how to read the book with the current link.

2. The book seems very sparse, with many of the entries in the table of contents being empty.  (I do appreciate that it has sections about [P2WPKH](https://programmingblockchain.gitbooks.io/programmingblockchain/content/other_types_of_ownership/p2wpkh_pay_to_witness_public_key_hash.html) and [P2W* over P2SH](https://programmingblockchain.gitbooks.io/programmingblockchain/content/other_types_of_ownership/p2w_over_p2sh.html).)

My suggestion would be to keep this PR for a few weeks or months to make sure more content is being added to the book, and if it is, then merge once the link is updated.

@nopara73 as an aside, as a writer I understand the utility of putting in empty chapters, but as a reader I wanted to point out that it's rather disappointing when you discover that book doesn't have as much content as it claims.  My recommendation would be for you to remove the stub chapter titles in the release version of the text.

NicolasDorier:
@harding this book is the same one I wrote more than 1 years which was only available in pdf. (https://aois.blob.core.windows.net/public/Blockchain%20Programming%20in%20CSharp.pdf)

@nopara73 moved it to gitbook, as a result, I will stop my pdf edition. The book is rather complete, actually it is around 100 pages. The empty chapters need to be removed though, I added them long time ago as I wanted to write about but they are not relevant anymore.

Please consider merging once I removed the empty chapter, the content of the book is mature enough.

harding:
The PDF does make the book look much more complete; I must've missed seeing some content when I was skimming the gitbook.

I dropped my merge access here, so the current site maintainers will have to make the decision to merge or not.  I do still think the link should be updated per my previous comment before a merge.

NicolasDorier:
@harding I just removed the missing chapter.
Actually this is just an impression, the content is the same. Except the first chapter which was rewritten more clearly.

It seems smaller because the summary on the right is only 2 level deep.
It makes me think I should probably do something so unimportant chapter does not take so much space in the left summary.

nopara73:
Please do not forget to deal with this PR.

laanwj:
ACK, it's good to link the various disperate sources of documentation about
bitcoin together

PR #1317, opened 2016-06-26
Update Airbitz to [validation: "checkpassvalidationspvp2p"]

paullinator:
Airbitz has been able to connect to any Electrum/Stratum server since Nov 2015 and randomly connects to 90% non-Airbitz owned, publicly available servers

Also update Android and iOS screenshots.


crwatkins:
Hey Paul, thanks for the update!   Great idea on including Electrum servers.  Does your wallet still download a list of SPV servers from an Airbitz server upon startup (the old getinfo api call)?  Do you still include Obelisk/bs servers in that list or is it now entirely Electrum?  How do you create your list of Electrum servers?  Is it directly from the \#electrum IRC channel or do you curate the list?

paullinator:
Hi @crwatkins The wallet does download a list from an Airbitz server. It does it once per day and caches the results. We do still include Obelisk/bs servers, but our next release will exclude them for now until some optimizations are made to Libbitcoin. We grab the list from those shown in the Electrum client but prune out any that appear to be non-functional.

crwatkins:
Thanks @paullinator I agree that the Airbitz validation score should be changed, but I propose that we change it to `checkpassvalidationspvservers`, "...uses SPV and random servers from a list" which is the scoring used for Electrum.  If you agree with this and would like to update the PR, I will gladly recommend the merge.

paullinator:
@crwatkins I'm sorry. I just realized that I picked the wrong validation type completely. I meant to copy Greenaddress which is non-SPV but from an open server list. I accidentally copied Greenbits instead. I will submit a new pull request which selects checkpassvalidationservers instead.

PR #1149, opened 2015-12-01
Keepkey

dstanchfield:
This pull request is for adding KeepKey wallet to the "Choose Your Wallet" section.

PR #1319, opened 2016-06-27
Add Opendime

nvk:
Adds Opendime now that it is version 1.1.0 

Release notes: https://opendime.com/downloads#1.1


btchip:
(reg. disclaimer, CTO of Ledger speaking, so technically "competition", but willing to grow a healthy and diverse hardware ecosystem here)

I think this PR is a good opportunity to think about a new category (that we'd also support creating), maybe "Bitcoin Digital Bills" or "Bitcoin Physical Bearer Tokens" considering that compared to other hardware wallets, Opendime is mostly a different animal : 

- It is "sealed", allowing the user to physically check if the coin has been spent or not
- It is not designed to be reused
- It is not protecting the end user against malware

peter-conalgo:
:+1: on all those points, very true. It is a new category and it's goals and security model are very different from a traditional hardware wallet. However, I'd like to merge as-is first, because it is hardware and so people will look for it there. Once a new category is made, I'd be happy to move it over.

nvk:
Sure, I think this expands what traditional hardware wallets were able to do;

1. Opendime never requires software install, or app download. When unsealed, like a paper wallet anyone can sweep it anywhere. 
2. It prevents the original giver/sender from stealing the coins from the receiver with a backup.
3. It can be loaded into a offline computer, and the sweeping can be done with a phone wallet QR code reader.

So yes, it is different, but can be powerful in a different way.

btchip:
> So yes, it is different, but can be powerful in a different way.

Yes, that's my point. It might also be confusing for users to list it as a hardware wallet when a key functionality of hardware wallets is to protect users from malware and this doesn't, unless used in special operating conditions.

A hardware wallet is defined as follows on bitcoin.org website (https://bitcoin.org/en/secure-your-wallet#hardwarewallet)

> Hardware wallets are the best balance between very high security and ease of use. These are little devices that are designed from the root to be a wallet and nothing else. No software can be installed on them, making them very secure against computer vulnerabilities and online thieves. Because they can allow backup, you can recover your funds if you lose the device.

EricLarch:
If the creation of a new category is not possible because too early, it would be then more precise to amend the description "Opendime is a new kind of hardware wallet" in something like "Opendime is a new kind of hardware bearer bond". This would reduce the confusion, and make a clear distinction.

crwatkins:
This is an excellent discussion.  I'm hesitant to recommend the creation of a new category for a single listing at this time.  While @btchip [points out](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1319#issuecomment-229065103) some inconsistency between the bitcoin.org definition of hardware wallets and Opendime, I would be inclined, at least for the time being, to include Opendime in the existing hardware category, with perhaps a more precise description as suggested by @EricLarch.

As the market evolves, I would definitely support some changes to the categories, and it still remains my [intention](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1149#issuecomment-163090820) to propose a change to the reviews of hardware wallets to evaluate combinations of hardware and software as a wallet system in the future.

nvk:
We agree with @EricLarch @crwatkins to add extra information describing the different functionality.

peter-conalgo:
:+1: Concur. @EricLarch @crwatkins @nvk

crwatkins:
I have reviewed Opendime based on the current wallet requirements criteria and my evaluation is below. I evaluated Opendime using the same standard criteria that would be used for any hardware wallet based on the premise that all of our currently listed hardware wallets are basically key stores with a signing engine and Opendime is basically a key store (perhaps as a community we will refine the classification of these types of devices in the future such that different criteria apply). The summary is that because of the recent release (just shy of three months), insufficient usage based feedback, and a bug that I noted in the review, I cannot at this time recommend Opendime for listing as it may be too new for general use. I would be happy to re-review Opendime after the issues have been addressed and sufficient usage has been identified. Note that requirements criteria also provide for an independent audit as an alternative to usage feedback.

Note that if this device continues to be thought of as a hardware wallet, the lack of open source makes a successful review much more difficult as one of the requirements is for either open source or detailed blackbox testing.  In addition, the lack of open source makes many aspects much harder to verify in a device like this.  A few more notes on this subject will follow the review.

---------

## Opendime

### Version v1.1.0

### Review Version 2016080201

The wallet list is based on the personal evaluation of the maintainer(s) and regular contributors of this site, according to the criteria detailed below.

**NOTE** The hardware devices used for testing were provided to bitcoin.org by Coinkite at no cost

Basic requirements:

**FAIL** There has been a fair amount of discussion on [Reddit](https://www.reddit.com/r/Bitcoin/comments/4cmlz5/opendime_worlds_first_zerotrust_physical_bitcoin/) and [Bitcointalk](https://bitcointalk.org/index.php?topic=1452987.0) but there has only been one [review](https://medium.com/@beautyon_/exquisite-opendime-ad1195a2790e#.lfh48aw9k) found with little technical analysis.

**PASS** No indication

**PASS** No indication

**NOTE** Closed source.  No indication.

**PASS** No indication.

**FAIL** May 3, 2016 https://twitter.com/Coinkite/status/727628795663011841

**FAIL** A concerning bug was found.  Because of web browser caching, the QR code displayed for an Opendime private key was incorrect (the private key from a different Opendime, cached by the browser, was displayed instead).  This has been reported and Conkite is planning a fix in v1.1.1

**PASS** http://opendime.com redirects
**FAIL** http://www.opendime.com does not redirect 

**PASS** https://opendime.com rating: A


**FAIL** no HSTS


**PASS** Opendime claims to be made by Coinkite, a long standing member of the community with their principals well known and a legal contact at https://opendime.com/legal

**NOTE** A more direct reference to details about the company or individuals is recommended

**NOTE** Wallet is single use and has only one address

**NOTE** Hardware has no change address

**N/A**

**N/A**

  - Allows backup of the wallet

**FAIL** No method to back up wallet

**FAIL** No method to backup/restore

**FAIL** Source code is not open

**N/A**

- For hardware wallets:

**NOTE** This hardware does not sign transactions

**PASS** Firmware upgrades are not possible

**PASS** Hardware accepts user supplied entropy to generate key

**FAIL** Closed source.  No detailed specification for blackbox testing, but some tools for verification


Optional criteria (some could become requirements):

**NOTE** No independent security audit

**N/A** No UI

**N/A** Does not sign transactions

**PASS** Provides link to support email address support@opendime.com 

**N/A**

  - Full reserve audit(s)
  - Insurance(s) against failures on their side

  - Supports HD wallets (BIP32)

**N/A** This is a single address wallet

**NOTE** There are instructions for recording the entropy supplied for initialization, but this is not sufficient to reproduce the key

**N/A** No backups

  - On desktop platform:

**N/A**

- For hardware wallets:

**PASS** Firmware is not downgradeable (nor upgradeable)

------------------

### A few more thoughts on Opendime

Opendime is a new and innovative device which does not share some of the characteristics of traditional hardware wallets.  One difference is the lack of ability to backup the device (the feature might be considered a bug in this case).  My concern with the lack of backup is that there may be uncorrectable failure modes (temperature? static discharge? x-ray? over voltage? physical shock?) that average users are unable to anticipate and Opendime is unable to correct.  Opendime is sometimes compared to a bearer bond, however the average user is likely to be able to anticipate and understand the physical failure modes of a paper based bearer bond better than an Opendime.

I'm also concerned that it would be extremely difficult, or perhaps impossible for even an expert to identify a modified or counterfeit Opendime.  A rogue Opendime that chooses predictable keys would not be able to be identified until a post mortem after the device is "unsealed", which could be too late to recover funds.  

With closed source software, I can only speculate on other attack scenarios, and not easily nor independently, confirm or deny hunches.  For example, in an effort to prove knowledge of the private key for an address, the device can be called upon to sign short chosen plaintext messages.  This is unlikely to expose a vulnerability, but it is difficult to confirm. Likewise, a "first owner" of the device, one who seeds the private key, does so using a known entropy seed and a known serial number.  The only part unknown to the first owner is the entropy provided by Opendime's internal random number generator which is closed source, leaving some speculation as to the actual amount of unknown entropy in the private key. The "second owner" (one that "unseals" the device expecting to be presented with the private key) has no way to verify in advance that when unsealed, the device (authentic or counterfeit) will indeed reveal the private key.

I suspect that these issues will likely either become less important (perhaps solved) or more obvious as time goes on and Opendime gets into more hands and more usage scenarios are discovered.

nvk:
1. Half of these considerations are features.
2. We test before we ship, hence no people complaining on reddit.
3. Maybe in some months we make a new PR.

crwatkins:
@nvk wrote:

> Half of these considerations are features.

I was hoping we all could take what might have been the fastest route to listing by using existing criteria rather than a potentially long drawn out process of adding a new category.  On that, I failed.

> We test before we ship, hence no people complaining on reddit.

I'm painfully aware of this catch-22 criteria; more new wallets than not run into it.  It stems from the fact that the community doesn't have the resources to do in-depth evaluations and must partially rely on external feedback.  For that reason, there is an alternative to the criteria to provide an independent audit.  That's a substantial effort that I don't believe any new wallet has undertaken, but it is available.

> Maybe in some months we make a new PR.

That sounds like a good idea.   We can also pursue new product listing categories on bitcoin.org.  I would support that effort.

> I suggest Bitcoin.org find a better method to include new product categories to avoid stifling innovation.

Anyone in the community is welcome to submit a PR to add new product listings (or for that matter, make any change in any form), but I'm painfully aware of how much work that is.  First you have to have a well thought out idea that will past the muster of many smart and many opinionated (some of them being the same) people, and then you actually have to do the work to make the changes.  @nvk I know I'm not telling you anything new, having just submitted this PR yourself.

I highly encourage people to do just that, while understanding it can be a lot to ask from from a community composed entirely of volunteers.

nvk:
It's not like we are new here, Coinkite was one of the very early wallets listed on Bitcoin.org. 

This is incredibly frustrating because the only objection came from the competition. I think its an overall loss for users and the further development of Bitcoin use. 

Right now, I think is better that we spend our time making new things than battling to have them listed. Maybe we try again in the feature.

Cheers,

btchip:
> This is incredibly frustrating because the only objection came from the competition

I'm afraid there's a slight misunderstanding here. It was not an objection, it was a suggestion, actually to help your product getting traction. 

If it gets listed as a hardware wallet, you'll be compared to a hardware wallet, likely not to your advantage - and it'll be exactly the same thing the other way, if you compare a hardware wallet to a physical bearer token, most will fail the price and physical validation test.

nvk:
Here is what you guys are missing out https://www.youtube.com/watch?v=9UFF9d3Y1BY

nvk:
It's been almost 6 months,  thousands sold and no reported problems. 

Even looking at it very conservatively I think its time to get merged 👍

crwatkins:
I base my recommendations for wallet listings on the wallet criteria (see [above](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1319#issuecomment-237137696)).  I was sorry (for all of us, considering the amount of time that I devote to such a review) that Opendime does not meet all of the hardware wallet criteria (perhaps I should have figured that out sooner, but I was being optimistic that it might be similar enough).   I'll mention again that I think that I made a mistake trying to review it according to the hardware wallet criteria, and I apologize for any of your time wasted during the review process. I see the following paths forward:

1. We (the community) change our listing criteria for hardware wallets.  I do not prefer this option because I think that our criteria works well for our "Choose your wallet" section.  I also believe that this would not serve our target user base well because "Choose your wallet" is step 2 in "Getting started with Bitcoin."

2. You could change the Opendime product to meet the listing criteria.  That seems neither practical nor useful.

3.  We could create new listings for new types of devices (as originally proposed above by @btchip, and panned by the frequently lazy yours truly).  These listings could be finely focused, or could be open to all sorts of Bitcoin devices such as Bitcoin ATMs and paper wallet printers and such.  The listings may be under the wallets section or perhaps more likely in a new separate section. This could allow for different criteria than our current wallets section.  As I mentioned [above](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1319#issuecomment-237241785) I would support that effort.

To summarize my personal view, after very detailed review, I believe that Opendime does not fit into the category of "Getting started with Bitcoin: Choose your wallet" and should be listed in a different category.

nvk:
See this is where most would disagree. A huge reported user case is giving bitcoin to new users for the first time. They don't need to have a wallet yet in order to received it.

nvk:
Pressed enter too fast...

I think having another category is good, and we should support and inform about all kinds of uses. And I would support such category, and would be happy to have Opendime listed there as well.

I do think easy on-raping with Opendimes or Paper wallets belongs on getting started. 

All in all, I defer to you the maintainer. 

If there was another category, when do you think we can have it up and how would it look like?

crwatkins:
As I mentioned [above](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1319#issuecomment-237241785) this site is reliant on volunteers like you and me to to do the work.  Anyone who wants to can propose a PR and I highly encourage it.

wbnns:
@nvk Hi, apologies for the delay - we're currently looking into what might be an optimal place to add this to the site. Since it is a one-time use product, we're not sure it fits the definition of a wallet.

Additional comments and feedback from other contributors and community members are welcome.

nvk:
@wbnns 👍

dooglus:
> I do think easy on-raping with Opendimes or Paper wallets belongs on getting started.

I'm unaware of such a feature but perhaps it would be better left undocumented. People love surprises.

wbnns:
Hello, just an update - PR #1464 has been opened which will close this PR and add Opendime to the Resources page.

PR #1323, opened 2016-06-30
Clarify that Pubkey Scripts are called scriptPubKe

Flowdalic:
Readers could get the impression that the script for fulfilling the conditions is called scriptPubKey in code (which is not true).

### DIFF
--- a/_data/glossary/en/pubkey-script.yaml
+++ b/_data/glossary/en/pubkey-script.yaml
@@ -10,7 +10,7 @@ required:
         A script included in outputs which sets the conditions that must
         be fulfilled for those satoshis to be spent.  Data for
         fulfilling the conditions can be provided in a signature script.
-        Called a scriptPubKey in code.
+        Pubkey Scripts are called a `scriptPubKey` in code.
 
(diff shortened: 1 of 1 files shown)


_data/glossary/en/pubkey-script.yaml
```diff
@@ -10,7 +10,7 @@ required:
         A script included in outputs which sets the conditions that must
         be fulfilled for those satoshis to be spent.  Data for
         fulfilling the conditions can be provided in a signature script.
-        Called a scriptPubKey in code.
+        Pubkey Scripts are called a `scriptPubKey` in code.
```
harding: The rephrasing is a good idea, thanks!  Could you drop the backtics?  Two reasons: (1) they won't be rendered by markdown everywhere this is used, which is why this field is called `summary_max_255_characters_no_formatting` and (2) scriptPubKey is used in most Bitcoin documentation without being rendered as a `code` string.

Flowdalic:
Removed backticks

harding:
ACK.  Thanks again!

PR #1326, opened 2016-07-02
Move Saivann Carignan into inactive contributors

saivann:
Small update to the contributor page as I have not been active since a while.


crwatkins:
I would like to extend a big thanks to Saïvann for all his work here for the community, and for all the help he has given me personally with bitcoin.org.

saivann:
@crwatkins Thanks :)

PR #1327, opened 2016-07-04
Move avoiding address reuse criteria from optional to required

crwatkins:
I recommend making the optional "avoid address reuse" criteria mandatory.  All currently listed wallets meet this criteria.

In the future, I may recommend adjusting the privacy scoring since this criteria is the only requirement for a passing score.


greenaddress:
Seems reasonable to me. ACK in principle.

harding:
ACK.  I think this is really important.  Thanks!

shangzhou:
ACK.I think every wallet should make the address change automatically when receive BTC.

Issue #1329, opened 2016-07-07
[BUG] Knotz has the same screenshot with Core


mcccs:
![screen shot 2016-07-07 at 18 32 04](https://cloud.githubusercontent.com/assets/20167871/16659003/22e320aa-4471-11e6-9684-8e0d61c0c478.png)

luke-jr:
Uh? Clearly shows a different icon in your screenshot...

greenaddress:
@luke-jr I think he means the screenshot, it states "Bitcoin Core - Wallet" rather than "Bitcoin Knots - Wallet"

mcccs:
> @luke-jr I think he means the screenshot, it states "Bitcoin Core - Wallet" rather than "Bitcoin Knots - Wallet"

Yes, this is what I meant.

luke-jr:
Not sure why the titlebar matters so much...

MarcoFalke:
@mcccs Please update the screenshots to the most recent versions by creating a pull request.

mcccs:
I don't have Linux, can I take a screenshot on Mac or Windows ?

MarcoFalke:
It should be easy to get Ubuntu running in a virtual machine.

mcccs:
I did, and unzipped tar.gz
The problem is, I can't find the installer (or stand-alone)

There's a file called Bitcoin-QT but it  has no extensions and I can't launch it.

MarcoFalke:
Press `CTRL + ALT + T` to open a terminal. Then something like

```
cd Downloads/bitcoin
./bin/bitcoin-qt &
```

to run the binary.

PR #1330, opened 2016-07-08
Add Bitcoin Knots screenshot

mcccs:
Resolves: #1329


jonasschnelli:
ACK

mcccs:
Also, there's a line that says:

>         link: "http://bitcoinknots.org/"

Should I replace it with `https`? I checked, https works too; but http page doesn't redirect to htttps.

luke-jr:
ACK.

I use http because https is a false sense of security (at least in this case). I want people to use PGP to verify the integrity of downloads instead.

PR #1344, opened 2016-08-18
Expanding 2016-08-17-binary-safety alert

BashCo:
Encouraging users to cross reference keys across multiple sources beyond Bitcoin.org. Providing one additional source for Wladimir's signing key, and providing a link to the gitian signature repository so that signatures from other developers can be used for verification as well.


harding:
LGTM.

theymos:
I think that it is an improvement, but I wonder if that might make visitors even more confused. Especially the last line is like, "If you don't do HORRIBLY_COMPLICATED_STEPS, then you're insecure." Not sure that this can really be improved without compromising security, though.

MarcoFalke:
LGTM

PR #1368, opened 2016-09-19
Update sv.yml

LevinSwe:
Change transaction to confirmation row 666. Sweng translation follow:
Remove: Each Transaction exponentially decrease the risk for removed transaction.
Replace: Each Confirmation exponentially decrease the risk for removed transaction.


Mirobit:
This has been fixed on [Transifex](https://www.transifex.com/bitcoinorg/bitcoinorg/) (_bekräftelse_) but it seems that the Swedish language file hasn't been updated in a while.

wbnns:
@LevinSwe Thanks for submitting this fix. @Mirobit Thank you as well for confirming the change on Transifex. This will be fixed along with [several other Swedish translation strings](https://www.diffchecker.com/wXs2s1DS) in the near future.

PR #1385, opened 2016-10-06
Version message Relay field is optional

jnewbery:
Update the version documentation to show that the relay field is optional.


wbnns:
@jnewbery Thanks!

PR #1392, opened 2016-10-13
Alert system retirement notice

achow101:
As per discussion on the -core-dev IRC, here is the post for the retirement of the alert system. The actual alert will happen tomorrow. This post should go up today before the actual alert as a pre-warning warning.

IRC Discussion of this post: https://botbot.me/freenode/bitcoin-core-dev/2016-10-13/?msg=74784693&page=6

### DIFF
--- a/_alerts/2016-10-14-alert-retirement.md
+++ b/_alerts/2016-10-14-alert-retirement.md
@@ -20,7 +20,7 @@ The network wide Alert system was created by Satoshi Nakamoto as a means of info
 information regarding Bitcoin. It has been used in the past to inform users about important network events such as
 accidental blockchain forks. However, the Alert system also represents a large source of centralization in Bitcoin.
 The holders of the singular Alert Key can at any time send an alert which could affect the entire network. As more
-developers join, the Alert Key is given to others, but cannot be taken away from those who have left. This has lead
+developers join, the Alert Key is given to others, but cannot be taken away from those who have left. This has led
 to the Alert Key potentially falling into the hands of malicious actors who could use it to disrupt the network. Because
 there is only one Alert key, it is not possible to prevent former developers from sending an alert nor is it possible
 to identify who sent an Alert.
--- a/_alerts/2016-10-14-alert-retirement.md
+++ b/_alerts/2016-10-14-alert-retirement.md
@@ -1,0 +1,62 @@
+---
+## This file is licensed under the MIT License (MIT) available on
+## http://opensource.org/licenses/MIT.
+
+title: "Alert System Retirement"
+shorturl: "alert-retirement"
+active: true
+#banner: "Alert system is being retired (click here to read)"
+#bannerclass: "alert"
+---
+
+## Summary
+
+The network wide Alert system is being retired. **_No Bitcoins are at risk and this warning may be safely ignored._** 
+Upgrade to the newest version of your wallet software to no longer see the alert.
+
+## Reasons for Retirement
+
+The network wide Alert system was created by Satoshi Nakamoto as a means of informing Bitcoin users of any important
+information regarding Bitcoin. It has been used in the past to inform users about important network events such as
+accidental blockchain forks. However, the Alert system also represents a large source of centralization in Bitcoin.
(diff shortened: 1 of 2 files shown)


achow101:
This post should go up today before the actual alert as a pre-warning warning.

IRC Discussion of this post: https://botbot.me/freenode/bitcoin-core-dev/2016-10-13/?msg=74784693&page=6

harding:
This LGTM (note: I don't have merge access; either @Cobra-Bitcoin or @laanwj will have to merge).  Here's what the alert will look like on the Alerts page:



I confirmed the alert is listed in the RSS feed.  Here's a full page preview of the rendered Alert:

screenshot-127 0 0 1 4000 2016-10-14 05-22-27

harding:
I also confirmed that the redirection from `bitcoin.org/alert-retirement` to this alert will work when this is published.

gmaxwell:
LGTM. I'm not sure we're actually sending it the 15th.  If we're RCing a new release ASAP, it would be a little unfortunate to cause the few remaining users on older versions to upgrade to something that is immediately outdated.  Page should go up in advance (though perhaps not highlighted on the site). @laanwj ?

harding:
As background information for whatever decision is made:

This PR as of commit https://github.com/bitcoin-dot-org/bitcoin.org/pull/1392/commits/1b8bcce537b04d236f7f75f5f87b70d1b1b99b88 is emphasized on the site only through a warning icon in the footer on every page:



Clicking that link will take the reader to the Alerts page where this alert is prominently featured (see screenshot above).

The alert is also listed in the alerts RSS feed.

There is no colored banner at the top of ever page as we've done in the past for serious situations.  (The part of the YAML header that would cause that to be rendered has the text written but is commented out; I assumed this was so it can be uncommented when the pre-final alert is actually sent.)

---

If this page's YAML header field `active:` is changed to `false`, then this alert will still appear in the RSS feed, but the warning icon in the footer will not be displayed and this alert's listing on the Alerts page will be a standard non-prominent list item.

achow101:
I'll adjust the dates as we go. 

Should this have the banner displayed? I wasn't sure what to do about that. I think that the banner should be displayed initially for the first few days of the Alert then taken down until the Final alert is made and it can go back up briefly for that too.

> If we're RCing a new release ASAP, it would be a little unfortunate to cause the few remaining users on older versions to upgrade to something that is immediately outdated.

Considering that we are still probably a month out from the release itself, I think it would be fine to do this in the next couple of days.

harding:
The banner color can be toggled; we have the following presets but others can be created:

```yaml
## (Optional; default=alert) CSS class to set banner color
##   alert = red  |  warning = orange  |  success = green  | info = blue
bannerclass: info
```

I think an `info` banner would probably be good before the pre-final alert is sent as well as maybe for a week or two after it is sent.

More info about Bitcoin.org alerts options is here: https://github.com/bitcoin-dot-org/bitcoin.org#alerts

achow101:
I've set the banner to `info`.

harding:
For reference, here's a preview of the banner as set in 8039c66 : 



(This is from the main page; the banner is shown on every Bitcoin.org page.)

Cobra-Bitcoin:
Looks good.

Not sure if I should merge this now, it doesn't look like things have been 100% decided yet.

achow101:
@Cobra-Bitcoin It really depends on when @gmaxwell wants to send the alert. IMO we can merge this now and move forward with the alert for tomorrow. I can update the dates.

MarcoFalke:
> It really depends on when the 0.13.1 RC cycle begins 

There are still two bugs, so rc1 will come earliest next week. But this does not help here in any case because we don't want people to upgrade production systems to a rc.

MarcoFalke:
utACK 8039c66

achow101:
Pushed the dates back by one day and also added some additional links to related PRs.

laanwj:
ACK, thanks for writing this

achow101:
Bumped the dates. I think this can be merged

kanzure:
ACK 0ea7206d30fa0a561921c64a5fe34ea16266efbe

PR #1396, opened 2016-10-17
Push back prefinal alert date

achow101:
Pushing back the prefinal alert date because the RC process for 0.13.1 has already started.

See: https://botbot.me/freenode/bitcoin-core-dev/2016-10-17/?msg=74970645&page=5


MarcoFalke:
ACK, this makes sense. If everything goes well 0.13.1 might be out quite soon.

PR #1397, opened 2016-10-17
Revert prefinal alert notice to prevent people from upgrading

achow101:
@Cobra-Bitcoin This is urgent

We do not want people to upgrade yet because 0.13.1 is almost released. 
 
We can revert this after the release and just before the actual alert is sent.


Cobra-Bitcoin:
Done https://github.com/bitcoin-dot-org/bitcoin.org/commit/1d66c5fc8c4054e2ce21ec881e270df01aabceea.

MarcoFalke:
@achow101 I think this can be reverted again, now?

achow101:
Yep. Just need to update the dates. I'll do that in about an hour.

achow101:
See #1401

Issue #1413, opened 2016-11-19
Typo in block version header data type?


axvm:
Hello,

[Developer reference says](https://bitcoin.org/en/developer-reference#block-headers) version header data type is `uint32` but as @fanatid mention in bitcoinjs/bitcoinjs-lib#712 
[protocol documentation says](https://en.bitcoin.it/wiki/Protocol_documentation#Block_Headers) version header datatype is `signed int32`.

I'd like to know who is right. Thanks.

Source File: _includes/devdoc/ref_block_chain.md

axvm:
There is definitely a typo in developer reference because [nVersion declared as `signed int32` in official client.](https://github.com/bitcoin/bitcoin/blob/master/src/primitives/block.h#L24)

MarcoFalke:
@axvm Pull request welcome

axvm:
@MarcoFalke done. Let me know if I need to change something in pr.

axvm:
Made PR in correct repo.

PR #1414, opened 2016-11-20
Changed version data type to signed int32

axvm:
Fixed typo version data type in block headers. Based on issue #1413

### DIFF
--- a/_includes/devdoc/ref_block_chain.md
+++ b/_includes/devdoc/ref_block_chain.md
@@ -20,7 +20,7 @@ serialized header format part of the consensus rules.
 
 | Bytes | Name                | Data Type | Description
 |-------|---------------------|-----------|----------------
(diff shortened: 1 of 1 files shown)


jonasschnelli:
ACK

References:
* https://github.com/bitcoin/bitcoin/blob/master/src/primitives/block.h#L24
* https://github.com/libbtc/libbtc/blob/master/include/btc/block.h#L45

luke-jr:
NACK, it's a bitfield, not an integer. Implementation details are irrelevant...

jonasschnelli:
@luke-jr: It's a bitfield when deploying SF with BIP9, but "old" blocks are still using a unsigned integer as block version. I agree that the documentation should mention the new bitfield after BIP9. But the internal type can and should be a int32 instead of a uint32.

luke-jr:
The top bit is unusable even pre-BIP 9, so meh. :)

syaifudin857:
Join

PR #1419, opened 2016-12-02
resources: Remove Bitcoin Foundation

wbnns:
At present on our [Resources page](https://bitcoin.org/en/resources), the first link is to the [Bitcoin
Foundation](https://bitcoinfoundation.org/). That website has fallen into disrepair and is currently
broken.

Each day new users are discovering Bitcoin for the first time,
along with others who are learning more about it via Bitcoin.org. This
PR removes the link to the Bitcoin Foundation, for now, so we aren't
providing a top resource pointing to a broken web page:

image

### DIFF
--- a/_templates/resources.html
+++ b/_templates/resources.html
@@ -12,7 +12,6 @@ id: resources
 <div>
   <div>
     <h2 id="useful"><img src="/img/icons/ico_simple.svg" class="titleicon" alt="Icon">{% translate useful %}</h2>
-    <p><a href="https://bitcoinfoundation.org/">Bitcoin Foundation</a></p>
     <p>{% translate linkwiki %}</p>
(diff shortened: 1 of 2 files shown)


crwatkins:
ACK.  Site is broken.

MarcoFalke:
ACK. Not only the website seems broken.

Mirobit:
ACK. [Https](http://bitcoinfoundation.org/) seems to be the problem. Besides technical problems the last blog post is from June and it is not clear what they actually do.

Mirobit:
The donation part is still included in all other languages. What is the standard process here? Should this be fixed via Transifex or should I submit a commit here?

wbnns:
@Mirobit Hello, yes, that's correct the changes need to be made upstream on Transifex, so that they remain in effect on-site the next time the translations are updated.

This will be fixed in a subsequent series of commits. 👍

PR #1422, opened 2016-12-03
choose-your-wallet: Adding info about wallet properties.

jonathancross:
I did not find a clear page defining what the various properties like "checkfailtransparencynew" mean.  Is there something better we can link to?

### DIFF
--- a/_templates/choose-your-wallet.html
+++ b/_templates/choose-your-wallet.html
@@ -1,6 +1,9 @@
 ---
 # This file is licensed under the MIT License (MIT) available on
 # http://opensource.org/licenses/MIT.
+#
+# Definitions for the property checks below can be found here:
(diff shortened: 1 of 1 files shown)


crwatkins:
@jonathancross I believe you have indeed found the definitive self-describing definition at https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/_translations/en.yml

crwatkins:
@jonathancross I know this is really picky, but should the file reference in the comment be relative to our source tree instead of an absolute github URL?

I'm good with the addition either way.

jonathancross:
@crwatkins I think a full url is easier to copy / paste into a new tab for reference.
But happy to change if you'd prefer relative path.

crwatkins:
LGTM

wbnns:
Unless others object, this will be merged on Wednesday, December 14th.

PR #1428, opened 2016-12-08
GreenBits now supports Tor for both SPV and GreenAddress bc data/sign

greenaddress:
As of release [1.87](https://github.com/greenaddress/GreenBits/releases/tag/r1.87) of September 4th 2016 (now 1.93) GreenBits supports generic socks5 and explicit onion support with Orbot.

This PR updates the GreenBits wallet listing with the current released status.

### DIFF
--- a/_templates/choose-your-wallet.html
+++ b/_templates/choose-your-wallet.html
@@ -544,7 +544,7 @@ wallets:
         privacycheck:
           privacyaddressreuse: "checkpassprivacyaddressrotation"
           privacydisclosure: "checkfailprivacydisclosureaccount"
-          privacynetwork: "checkfailprivacynetworknosupporttor"
+          privacynetwork: "checkpassprivacynetworksupporttorproxy"
 - mycelium:
     title: "Mycelium"
     titleshort: "Mycelium"


crwatkins:
@greenaddress Could you point us to your user documentation that describes how to configure and use GreenBits with Tor?

greenaddress:
@crwatkins we don't currently have some explicit documentation. 

In any case the instructions are the following: Before entering the wallet, but once the app is open, there's an Android menu on the top right corner with Network options, once in that section it offers the ability to setup a socks 5 proxy and enable Onion Tor for talking with the multisig oracle (GreenAddress) - the latter only works if the socks 5 is Tor enabled (like Orbot or like a Tor daemon running on some other computer).

By default if one installs Orbot and starts it it will listen on 127.0.0.1 on port 9050. These details are shown to the user as an example and can be entered above. Additionally there's a Tor checkbox that forces the app to use the onion address for the server rather than going via the socks (Tor) but using the standard *.greenaddress.it url.

crwatkins:
@greenaddress I was launching GreenBits directly from a Ledger Nano, so I never saw that menu.

I already had Orbot running and "Connected to the Tor network", so I configured as you suggested.

When I plug in the Ledger Nano, GreenBits launches, asks for my PIN and one second later it closes (back to Android home screen).

When instead I ask to create a new wallet, on the passphrase page, there is a rectangular continue button.  When I press that button, it turns into a circle and then immediately back into a rectangular continue button, and repeats every time I press the button.  There are no error messages.

Do you have any suggestion or should pursue this through your support team and spare github?

In either case, I would highly recommend more detailed information in the app such as the information you provided above.  Also I'm not really sure whether that checkbox has anything to do with controlling the SPV traffic, or it will always use Tor if Tor is configured, or what.

greenaddress:
Thanks @crwatkins 

With Ledger or fingerprint auth indeed you don't get to see the option unless you start the app without Ledger first or unless you go cancel on the fingerprint auth, then you see the option and can then proceed to use Ledger or fingerprint. We plan to add/improve the documentation in this and other areas.

I repeated the steps and I had some issue establishing a connection at first, I had to wait a few minutes after Orbot was started before it worked but then it started working reliably each time. Can you try if it was just Orbot being a bit slow?

for any bug/issue/improvement probably best in https://github.com/greenaddress/greenbits/issues, meanwhile I created https://github.com/greenaddress/GreenBits/issues/192 to be more user friendly and handle disconnection better.

greenaddress:
@crwatkins to answer your last question either way all traffic goes via the socks (including resolving). The Tor option controls only whether the app will attempt to use the standard https domain or the onion one to talk to the backend. This is to allow the use of non Tor/Orbot Socks5 proxies.

For SPV, assuming SPV is enabled, it is as follows (again, the Tor flag has no impact):

If the proxy is not enabled the app attempts to do a DNS discovery.

If the proxy is enabled (regardless of Tor) the app won't do a DNS discovery but will attempt a HTTP discovery via Schildbach's HttpSeed (which seems to be giving 502: Bad Gateway at the moment).

In both cases if the user providers a full node address or more (onion or not) no DNS/HTTP discovery will be performed.

crwatkins:
Thanks @greenaddress, I've submitted details of Tor problems to GreenBits issues.  I've occasionally, but not consistently, been able to send and receive transactions with Tor seemingly enabled.  

However, during this time with SPV enabled, the Network Monitor shows "No Peers Connected".  If I disable Tor (remove the proxy from the network configuration) I see many peers listed in the Network Monitor.  Is that because of the HttpSeed issue that you mentioned above?  Should there be more feedback to the user here?

greenaddress:
yes related to current http seed issues. Perhaps there should be some better notification however the mode socks is mostly there for people setting their own nodes over Tor and they can always verify in the monitor if there are peers connected (and otherwise the synchronization notification only disappears/completed once spv caught up). the rest of the UI clearly marks transactions as unverified too.

from what I understand from reading bitcoinj updates and the current status of http seeds they don't have long life and over Tor we'd need some updates before we could have better discovery IMHO.

example config of today config:  https://www.reddit.com/r/Bitcoin/comments/5b9z9f/bitcoind_over_tor_a_miniguide_from_personal/

crwatkins:
I recommend the update to the listing in this PR.   While there seems to be issues with HttpSeed in SPV mode, we score based on the default mode of the wallet which is non-SPV.

@greenaddress I would highly recommend some enhanced feedback/diagnostics for users that might encounter the situations we discussed above.

wbnns:
Unless others object, this will be merged on Wednesday, December 28th.

@greenaddress @crwatkins Thank you both. 👍

Issue #1431, opened 2016-12-09
Soft Fork Definition


laudaa:
The soft fork definition, as found [here](https://bitcoin.org/en/glossary/soft-fork):

> A temporary fork in the block chain which commonly occurs when miners using non-upgraded nodes violate a new consensus rule their nodes don’t know about.

Seems quite different from what we can find on the [Wiki](https://en.bitcoin.it/wiki/Softfork):

> A softfork is a change to the bitcoin protocol wherein only previously valid blocks/transactions are made invalid. Since old nodes will recognise the new blocks as valid, a softfork is backward-compatible. This kind of fork requires only a majority of the miners upgrading to enforce the new rules.

Should be a trivial change in this [file](https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/_data/glossary/en/soft-fork.yaml).

theymos:
Yeah, that's totally wrong. The wiki text seems pretty perfect to me.

Mirobit:
Addressed in #1436

PR #1436, opened 2016-12-14
New softfork description

Mirobit:
As suggested in #1431 
Not happy with the description but better than the old one and options are quite limited with 255 chars.


luke-jr:
ACK after removing the last sentence. If only miners enforce a softfork, something is wrong.

Mirobit:
@luke-jr Agreed. While technical true, it is not desirable.

wbnns:
@Mirobit @luke-jr Thanks guys.

PR #1437, opened 2016-12-14
Show documentation link in header menu for all languages

Mirobit:
Not sure if this is useful but so far there is only a link from the _developers introduction page_ to the _Documentation_ for non English languages.
With this commit the _Documentation_ link is always display in the _Resources_ menu and _(EN)_ is added for non English languages.

devmenu


wbnns:
@Mirobit Thanks for thinking of ways to make the developer documentation more accessible to everyone. One issue with this change might be that if the site visitor decides to navigate through this new link, the rest of the site will change to English as well from that point on. Not sure if that might be confusing or overwhelming if the user is not an English speaker.

Also, if we can assume the the majority of people will know what `EN` means?

Let's see if anyone else would like to share thoughts on this.

Mirobit:
@wbnns You are right. I probably overestimate the amount of the developers. Most visitors are probably first time users.

wbnns:
@Mirobit Ok, cool. Still, great to bring up the point - finding ways we can connect more people with the developer material. More people building things that more people can use. 👍

PR #1445, opened 2016-12-19
Remove BFGMiner link because of malware warning on Chrome

i-rme:
Remove BFGMiner link because of malware warning on Chrome

### DIFF
--- a/_templates/development.html
+++ b/_templates/development.html
@@ -117,7 +117,6 @@ Want to contribute to a different project?
 
 <ul class="devprojectlist">
   <li><a href="https://github.com/etotheipi/BitcoinArmory">Armory</a> - A wallet with enhanced security features.</li>
-  <li><a href="http://bfgminer.com">BFGMiner</a> - A modular miner.</li>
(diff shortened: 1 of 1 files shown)


luke-jr:
NACK. That's a bug with Chrome, there is no actual malware.

See https://bugs.chromium.org/p/chromium/issues/detail?id=672984

wbnns:
@luke-jr Thanks for the note about Chrome. It looks like this is also happening in Safari and Firefox.

**Safari Version 10.0.1 (12602.2.14.0.7)**:
image

**Firefox 50.1.0**:
image

Is there a URL that this can be replaced with (instead of removing the link altogether as this PR suggests) until the issue is resolved?

i-rme:
@wbnns 
The webpage redirects to https://bitcointalk.org/?topic=877081

It might be a good replacement link although the download links still links to a website flagged as malware.

crwatkins:
NACK.  This is just a detection signature problem.  Extremely unfortunate, but there's nothing to be gained by removing our link.  Good luck @luke-jr getting them to fix it.

wbnns:
Unless others object, this will not be merged and the pull request will be closed on Sunday, December 25th.

@i-rme @luke-jr @crwatkins Thank you all for the feedback.

PR #1451, opened 2016-12-21
Remove CryptoCoinsNews from News Resources

kyletorpey:
I'm not sure why this site was added. It doesn't offer much of anything that you wouldn't find on the more established sites. There is little to no unique/original content, and there is apparently no fact checking before something is published. Latest example is an article claiming Phil Potter is a Bitcoin core dev while also getting his Bitfinex job title wrong (among other errors): https://www.cryptocoinsnews.com/bitfinex-cso-theres-chinese-political-effort-delay-block-segwit/


wbnns:
@kyletorpey Thanks for bringing this to attention. This was submitted as a pull request and merged after a waiting period for additional comments from the community, for which there was no dissent. That said, I agree with you. That's a pretty serious error with regard to the job title and is very misleading.

Unless others object, this will be merged on Sunday, December 25th.

PR #1462, opened 2016-12-28
Changes from Currency concept to Commodity concept.

theochino:
This change is due to the rulling of Florida (vs) Espinoza (Case # F14-2923)
and the Article 78 against New York Division of Financial Services (case to be heard 3/30/2017 in New York City)

For the good of (B|b)itcoin, it is important that the documentation reflect the true nature of what it is.
Currency is defined by the fact that a central authority creates and destroy a currency.
A commodity exist regardles of what the central authority decrete; so bitcoin is really the first "Virtual Commodity" ever created.

The documentation need to reflect the fact it is a commodity.
I kept the money part because, because the goal is that Money be backed by bitcoin and not gold or US/Euro.

### DIFF
--- a/_translations/en.yml
+++ b/_translations/en.yml
@@ -36,7 +36,7 @@ en:
     lowfee: "Choose your own fees"
     lowfeetext: >
       There is no fee to receive bitcoins, and many wallets let you
-      control how large a fee to pay when spending. Most wallets have
+      control how large a fee to trade when spending. Most wallets have
       reasonable default fees, and higher fees can encourage faster <a
       href="#you-need-to-know##instant">confirmation</a> of your
       transactions. Fees are unrelated to the amount transferred, so it's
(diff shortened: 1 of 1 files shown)


wbnns:
@theochino Hello, thank you for thinking of ways to improve the site and for spending the time to work through some of the content to propose this change.

I'm not sure we should be significantly modifying integral terms and definitions (and subsequent implied meanings) to make them potentially more complex and technical for everyone else in the world in order to conform with legal opinions in two of the United States.

Let's see if anyone else would like to share their thoughts.

davmarc:
I would think it's in the interest of any bitcoin-related organization to communicate (even advocate) bitcoin as a commodity rather than a currency to keep the public mindset and understanding that bitcoin isn't to be regulated to the same extent as a government issued currency which it clearly (for starters it isn't government issued). The extent of regulation and classification in some jurisdictions may vary, but I wouldn't jump to the full currency terminology based on that.

harding:
I oppose this change.  Bitcoin is money and has always been advertised as money---it has "coin" in its name and was first described as "electronic cash".

Moreover, I looked through the diff and the changes to phrasing make everything much less clear and much more confusing.  E.g. instead of "fee to **pay**" it becomes "fee to **trade**" which makes no sense---you don't trade fees, you pay them... with money.

IANAL, but it's my understanding that in the U.S. I can use gold coins to pay anyone who accepts gold coins as payment, and that we can in all respect talk about and treat those gold coins as money, at least up until the point when someone sues someone and legal tender laws come into play.  Lawyers are a clever lot, and if they can deal with gold coins being used as money then I think they can probably manage to deal with bitcoins being used as money too.

I do thank you for taking the time to create this PR.

mjdillon:
Perhaps the answer could be "both". Why not simply add some discussion about the legal recognition of bitcoin and cite how some jurisdictions are treating it as a commodity, perhaps citing that case as evidence?

theochino:
In today's Wired: https://www.wired.com/2017/01/bitcoin-will-never-currency-something-way-weirder

wbnns:
@theochino Hello, if you'd like to continue the discussion on this, please open an issue instead.

PR #1480, opened 2017-01-09
Update _events.yml

santy2001:
Adding an event


santy2001:
Sorry the event were already added

PR #1483, opened 2017-01-12
Example Transaction rpc output updated

Mirobit:


harding:
Untested ACK.  Thanks!

wbnns:
Unless others object, this will be merged on Sunday, January 15th.

@Mirobit @harding Thanks!

PR #1484, opened 2017-01-12
Bitcoin Core Releases, BIPs and protocol versions references update

Mirobit:


harding:
Untested LGTM.  Thanks!

wbnns:
Unless others object, this will be merged on Sunday, January 15th.

@Mirobit @harding Thanks again. :)

PR #1485, opened 2017-01-12
Correct an error in the rule for nTime; also wording clarification

daira:
nTime must be strictly greater than the median nTime of the previous 11 blocks, not just greater or equal:
https://github.com/bitcoin/bitcoin/blob/fac0f30482d56c7a38e9bdd7a4090b896965f030/src/validation.cpp#L2910


wbnns:
@daira Hello, thank you for submitting this.

Unless others object, this will be merged on Wednesday, January 25th.

daira:
Hmm, is there a delay before this is reflected in the site contents? I don't see the change at https://bitcoin.org/en/developer-reference#block-headers

wbnns:
@daira Hi, it's live - it just had to be deployed alongside some other changes first.

PR #1501, opened 2017-02-07
update info about TREZOR hardware wallet

prusnak:
TREZOR has been here for a few years already and is used by several thousands of people all around the world, so I think it deserves more than just a "New app" ranking.

I used `transparency: "checkgoodtransparencydeterministic"`, because our firmware can be built deterministically using Docker and one can verify that the fingerprint of the firmware distributed by us is the same as the fingerprint of the firmware built locally.

Also the product name is TREZOR, not Trezor.

### DIFF
--- a/_templates/choose-your-wallet.html
+++ b/_templates/choose-your-wallet.html
@@ -412,7 +412,7 @@ wallets:
         check:
           control: "checkgoodcontrolfull"
           validation: "checkneutralvalidationvariable"
-          transparency: "checkfailtransparencynew"
+          transparency: "checkgoodtransparencydeterministic"
           environment: "checkgoodenvironmenthardware"
           privacy: "checkneutralprivacyvariable"
 - ledgernano:


crwatkins:
@prusnak Thanks for submitting such a well deserved scoring update! I concur with the choice of ```checkgoodtransparencydeterministic```.

Bitcoin.org tends to use the Wikipedia [Manual of Style](https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Trademarks) for trademark representation as previously [discussed](https://github.com/bitcoin-dot-org/bitcoin.org/pull/921) specifically regarding Trezor.  If you can revert the spelling change, I will be very happy to recommend the scoring update.

Also note that the new https://trezor.io site does not meet the bitcoin.org HSTS criteria for listing.   I suspect you will want to get that updated shortly.

prusnak:
Reverted capitalization of the TREZOR.

HSTS for trezor.io is something we'll do soon, but we wanted things to settle down after the domain migratrions before we did something that is impossible to change afterwards.

crwatkins:
LGTM.  @prusnak Thanks again.

wbnns:
@crwatkins @prusnak Thanks!

PR #1503, opened 2017-02-09
Repository maintenance

wbnns:
This PR takes care of a few maintenance related items on the Bitcoin.org repository:

+ Adds a clickable link to "Watch" the Bitcoin.org repository on GitHub.
+ Adds a `CONTRIBUTING.md` for new and existing contributors to reference when working with the repository. This is the content that's currently in `README.md`, however, by adding this file a link to it will be displayed when people submit a pull request via GitHub, which is helpful.
+ Adds a `LICENSE` file (this is what is already in our [COPYING](https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/COPYING) file, along with the MIT license that it alludes to.
+ Updates my email address.
+ Adds a basic [Code of Conduct](http://contributor-covenant.org/) file to the repository for participants to reference.


achow101:
Contributing.md should really be for guidelines on contributing and less so how to contribute.

wbnns:
@achow101 Hello, thanks for the feedback - what parts of the document do you think aren't relevant for contributors? There are several sections that provide guidance along with contextual how-to information.

achow101:
While all of it is related to contributing, I think there are some sections that are more step-by-step howtos that I don't think belong in contributor guidelines. Specific "type this command" instructions don't really belong in the guidelines.

A table of contents would also be useful as the file is quite large.

These are things I think should be in their own files, probably in a docs folder. They are specific step-by-step howtos instead of guidelines for contributing. I don't think the full guides should be a part of the contributor guildelines, but just maybe a brief summary and a reference to the file.

* Instructions for building and previewing changes
* Translation instructions for the actual translations.
* Release notes instructions
* Alert instructions
* Wallet acceptance guidelines

From these two articles: https://github.com/blog/1184-contributing-guidelines and https://help.github.com/articles/setting-guidelines-for-repository-contributors/, the contributor guidelines are more for explaining how things should be formatted, what the code style is, etc.

wbnns:
@achow101 Thanks for the great ideas on this. 👍

harding:
I oppose the proposed code of conduct.  I think it's a boring document to read and, furthermore, that making rules about good behavior is---at best---useless and---at worst---tends to create a basis for boring arguments.  Like this one.

wbnns:
@harding Heya! I completely understand where you're coming from. Luckily stuff has been pretty docile here recently, however, in the past that hasn't always been the case (even outside of this repository, it's easy to find and witness flagrant abuse happening in the community right now).

With millions of people entering the space and potentially hundreds, if not thousands of new contributors on tap (with different backgrounds and ideals), I think it's helpful. Most major open source projects have some form of community guidelines, I don't see why we would be any different, or why it's a bad idea (if someone isn't interested, they just aren't going to read it - like you 😀):
+ https://www.mozilla.org/en-US/about/governance/policies/participation/
+ https://www.ubuntu.com/about/about-ubuntu/conduct
+ https://www.python.org/psf/codeofconduct/
+ http://rubyonrails.org/conduct/

In practice, yeah, it's silly that on this planet there even has to be stuff like this in the first place. At the same time, I don't see the harm for having something like this to reference for the people who are interested or for others who decide at some point in the future that they want to treat others without respect, etc. Working on open source projects, one thing people have often said is that some times they don't comment or contribute because they are worried that they might look dumb or could be personally attacked or ridiculed if others think their idea isn't a good one.

This, we can point to if needed at some point as something we work in harmony alongside, should we decide to warn about or moderate something (instead of just removing and getting labeled as authoritarians - which may happen anyways, but at least it's not coming out of left field and there's a reference on our end that's well adopted throughout other open source communities).

wbnns:
@achow101 Hey, this is just a quick note that I'm revising this PR based on your recommendations (thanks again for spending the time to organize some suggestions).

harding:
I should've noted in my earlier comment that I'd like my objection to be non-blocking.  I'm not as active here as I once was and I think the currently-active participants have every right to set the policies they think will make this an effective project.

wbnns:
Ok, this has been updated. 

Unless others object, this will be merged on Sunday, March 5th.

PR #1507, opened 2017-02-11
download: Note Ubuntu PPA as Unofficial

### LINKED ISSUE #1506: Conflict in installguide
The official download link leads to the ppa. However the ppa recommends to use official binaries.
 Which one should then be used and where downloaded from?

wbnns:
This fixes #1506 and provides additional clarity to the Ubuntu link by
adding a note that the PPA is Unofficial:

image

Unless others object, this will be merged on Wednesday, February 15th.

cc: @harding @pixdigit

### DIFF
--- a/_templates/download.html
+++ b/_templates/download.html
@@ -80,6 +80,7 @@ lin64: "x86_64-linux-gnu.tar.gz"
         <img src="/img/os/med_ubuntu.svg" alt="ubuntu">
         <span>
           <a href="https://launchpad.net/~bitcoin/+archive/ubuntu/bitcoin">Ubuntu (PPA)</a>
+            <span>Note: Unofficial</span>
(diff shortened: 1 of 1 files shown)


harding:
Perhaps `Note: unofficial` should be a translated string?  Otherwise LGTM.  Thanks!

luke-jr:
In what sense is the PPA unofficial?

luke-jr:
I consider the PPA official and recommended for Ubuntu users. It's technically less secure in a sense, but in the same way that using Ubuntu (or any OS) implies trusting the vendor maintaining it.

wbnns:
Let's put this on hold to see if anyone else wants to share feedback - also, FWIW, perhaps leaving it as-is, unchanged, could be a good option.

PR #1513, opened 2017-02-19
Update expiration date for laanwj keys

laanwj:
These keys have expired. Although the expiration date can be updated from the keyserver with --refresh-keys, it seems useful to update these here to avoid extra steps.

Edit: signed the commit


wbnns:
LGTM. 👍

Unless others object, this will be merged on Wednesday, February 22nd.

laanwj:
@andrea-chiavazza @wbnns Thanks for reporting, should be solved in #1525

PR #1514, opened 2017-02-22
update iTunes account for GreenAddress iOS app

greenaddress:
This PR changes the iTunes account we use for the GreenAddress app on iOS.

The app is otherwise unchanged (though we also added the testnet version now to the new iTunes account)


crwatkins:
LGTM

greenaddress:
thanks @crwatkins

PR #1534, opened 2017-03-08
alert: Drop current alert notification

wbnns:
Now that 0.14.0 has been released, this ends the notification about the
retirement of the alert system, which has been running and displaying on
the tops of all pages on Bitcoin.org for the last 4 months (to give
everyone advance notice).

Unless others object, this will be merged on Sunday, March 12th.

### DIFF
--- a/_alerts/2016-11-01-alert-retirement.md
+++ b/_alerts/2016-11-01-alert-retirement.md
@@ -4,7 +4,7 @@
 
 title: "Alert System Retirement"
 shorturl: "alert-retirement"
-active: true
+active: false
 banner: "Alert system is being retired (click here to read)"
 bannerclass: "info"
 ---


achow101:
The last step, releasing the alert key, has not happened yet. I think it
should remain up until then.

laanwj:
I think it's fine to remove it. 4 months is an eternity in bitcon. Everyone that cares knows about this, by now.  And I suppose the page will stay online on the site. We don't need to keep this as a critical alert alert all the time.

harding:
I did a quick search for "bitcoin alert" on DuckDuckGo and Google.  In the first case, the first two results both contained information about what was happening; in the second case, the first two results were price-monitoring software but the third result was a direct link to the retirement page.

Hopefully this means that anyone who still needs this information will be able to find it quickly even without the banner at the top of the page.

Mirobit:
LGTM. Agree with @harding. People easily find the page with any search engine.

wbnns:
Thanks everybody. 👍

PR #1555, opened 2017-03-29
fix incorrect translation key for btc.com wallet

rubensayshi:


crwatkins:
@rubensayshi For consistency, you should change they key in _templates/choose-your-wallet.html instead.

rubensayshi:
okay closing in favor of https://github.com/bitcoin-dot-org/bitcoin.org/pull/1557

PR #1561, opened 2017-04-10
Update security reporting instructions

laanwj:
Update security reporting instructions:
- The currently mentioned list (bitcoin-security@lists.sourceforge.net) was closed a while ago. Instead send people to the [contact page of bitcoincore.org](https://bitcoincore.org/en/contact/) to report security issues (which lists our current security alias,  security@bitcoincore.org).
- Remove the outdated list of core contributors.

We don't currently have GPG instructions on the contact page of bitcoincore.org for critical, private issues. This is something that needs to be addressed there.


wbnns:
@laanwj Thanks for fixing this, Wladimir. Merged. 👍

PR #1565, opened 2017-04-12
Added JP/KR exchanges

Mirobit:
I added a few Japanese and Korean exchanges.

I think for some exchanges the country classification needs to be changed. Paymium, Bitcoin.de and Rocked Trading are open to all SEPA supporting countries and have multilingual sites. Maybe a EU/Euro category would make sense.

While Bitwage is a great service I would not consider it an exchange or a way to buy/sell bitcoin.

### DIFF
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -110,6 +110,7 @@ id: exchanges
     <h3 id="japan"><img src="/img/flags/JP.png" alt="Japanese flag">Japan</h3>
     <p>
       <a href="https://bitflyer.jp/">bitFlyer</a><br>
+      <a href="https://www.btcbox.co.jp/">BtcBox</a><br>
       <a href="https://coincheck.com/">Coincheck</a>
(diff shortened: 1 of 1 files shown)


wbnns:
@Mirobit Heya! Thanks a bunch for adding these. 👍 Regarding Bitwage, I think it's okay, since people can use them to seamlessly exchange their wages paid in fiat for bitcoin.

Unless others object, this will be merged on Sunday, April 16th.

Mirobit:
@wbnns thanks for the fast reply. Any thoughts on the EU/Euro category proposal?

rosser2121:
Can you resend the proposal?

wbnns:
@Mirobit Hey, sorry, yes, a "Europe" classification sounds helpful. @rosser2121 He was saying that some exchanges are available beyond the countries they are currently listed under, and it might be help if we could classify those with more detail (i.e. Europe).

PR #1469, opened 2017-01-03
Add list of bitcoin exchanges by country to help people get bitcoin

wbnns:
This PR is a work in progress that creates an `Exchanges` page with a list of exchanges by country, starting with a list of exchanges in the world's 20 largest economies.

Preview:

bitcoin-exchanges

This page:
+ Links to the `Find an Exchange` button on the [Getting Started](https://bitcoin.org/en/getting-started) page and helps many of the new people entering the Bitcoin ecosystem via Bitcoin.org who are looking to adopt the technology, get bitcoin.
+ Improves SEO for Bitcoin.org, because the site can now potentially rank very well for "Bitcoin Exchange" as well as longer tail terms that include country names that are contained on the page (i.e. Bitcoin Exchange China, Bitcoin Exchange India, etc.) .
+ Has unique anchor links for each country that people can share on social media that will allow others to find the bitcoin exchanges that are available for their specific country.
+ Allows us to have an opportunity to focus the wallets page on products that allow people to maintain full control over their funds (Resolving #1109, because we could move custodial wallets that provide exchange services off of the wallets page to this page).
+ Will also be accessible via the `Resources` dropdown menu.

Unless others object, I will standby this week in case anyone would like to discuss, and plan to finish this, this weekend, and subsequently schedule this to be merged on Wednesday, January 11th.

Issue #1255, opened 2016-03-07
Adding "fee management" infos in the wallets section


jlegoupil:
Add “fee management” information in the bitcoin.org wallet section along with the standard infos : "control over your money", "validation", "transparency", "environment" and "privacy".

It would help users know how wallets manage fees so they can make the best choice and avoid having stuck transactions due to poor fee handling from wallets. Speaking for myself, "fee handling" was a decisive parameter when I chose my wallets. This is even more the case now.

We can also expect wallet innovation in the "fee management" space in the coming months due to the deployment of Opt-in RBF which makes it even more relevant, imo.

theymos:
This is a good idea IMO.

Cobra-Bitcoin:
Great idea!

jlegoupil:
A proposal for "fee management" categories :

- _Vulnerable to delays_ (orange) : the fee is hard coded (also includes wallets that enable different hard-coded fees such as low/high priority).

- _Basic protection against delays_ (green) : user can manually set the fee or the wallet estimates a smart fee from the current mempool of unconfirmed transactions.

- _Full protection against delays_ (bold green) : user can increase the fee after broadcasting the transaction to the network (RBF enabled wallets)

instagibbs:
@jlegoupil note that point #1 can also mean users are *regularly* over-paying in fees during slow times as well.

crwatkins:
The https://github.com/bitcoin-dot-org/bitcoin.org/pull/1568 PR has been merged.  I'll close this issue for now.  Please feel free open a new issue for any updates to the current fee scoring.  Thanks to @jlegoupil for the proposal.

Issue #1109, opened 2015-10-23
Bitcoin bank listings


crwatkins:
We currently list Bitcoin banks as wallets.  I know that there is some concern over the term "bank" implying some degree of trust or regulation which may not apply to the concept of a "Bitcoin bank" but I propose for this discussion that the term "Bitcoin Bank" simply implies third party control of funds.

I don't believe that our current wallet criteria https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/README.md#wallets provides (me at least) enough guidance to be able to recommend a Bitcoin bank alongside other Bitcoin wallets in the same categories.  I see this as similar to recommending home fire safes in the same category as national banks for the storage of funds; they're just not the same thing.  I would like to hear what others think about whether apps that are used for third party control should be recommended with wallets, separately, or not at all.  Do you think that our current criteria adequately covers Bitcoin banks?

In the meantime, I'll be having Bitcoin bank listing PRs tagged with a "Help Needed" label to indicate that we are looking for a volunteer, or volunteers to perform these reviews, at least until some other resolution can be found.

Current PRs which may fall into this category are #737, #832, #1056, and #1088.

Ninkip2p:
Yes, I think "bank" is the wrong word, but there should be a clear definition regarding custody of the funds and responsibilities.

I propose they should be separated into "Wallets" and "Custodial Services" with clear descriptions of the implications of each

Normally custodial services double-up as ways of buying Bitcoins so I would also flag if purchasing bitcoin is possible via the service, established services such as Coinbase are a way to introduce new people to Bitcoin with a familiar user experience and easy way to purchase, which is what a lot of noobs are looking for.

The Custodial Services section could subdivide into web, mobile etc. (similar to the wallets section)

*Any wallet purporting to be non-custodial without open-sourcing code should be ignored. These wallets are shirking responsibilities on both ends by not taking on custodial responsibility and at the same time refusing to offer proof that they are non-custodial (presumably to protect their IP)

crwatkins:
Since we haven't found anyone to review banks/custodial services nor have we had any refinements on our listing criteria, I have another proposal for consideration.  In agreement with @Ninkip2p I propose we create a separate listing category for banks (or whatever term we settle on).  There we would post listings with similar review and diligence as we now do for events (i.e. not much) as long as the listing meets some very basic criteria for legitimacy.  We would of course include a sufficient explanation of the listing criteria (a.k.a. "warning").

[...]

PR #1568, opened 2017-04-20
Wallets: add fee handling info to Choose Your Wallet page

harding:
Preview: http://dg0.dtrt.org/en/choose-your-wallet

This PR adds an extra entry to each wallet's listing describing how it handles fees.  For example, here's Bitcoin Core's updated entry:

This is really @crwatkins PR (although I support it); I just happened to have a Bitcoin.org build and hosting environment already setup for quick previewing and testing, so I'm opening the PR with his assent.

All fee-related text is at this point in the diff: https://github.com/harding/bitcoin.org/commit/61b28f582c2c308dc5f63a30a77006b6810476cd#diff-625d1a93b257504de31285773c0f35fd

Testing: I previewed each entry on Firefox and a few randomishly-selected entries on Chrome for Android.  More browser testing always welcome since this PR includes a couple CSS changes in a complicated layout.


theymos:
Very good idea!

I don't think that Bitcoin Core supports CPFP except through really hacky raw transactions stuff.

crwatkins:
@harding thanks so much for the build, preview, and CSS tweaks! 

Testing:  I previewed all entries on Safari and a few entries on Chrome on both an iOS phone and an iOS tablet.

@theymos I believe you are correct about CPFP and Core.  CPFP is only possible with many CLI commands with argument substitution.  However, Core supports RBF with a single simple CLI command (although no GUI) which allowed it to just squeak by with a "good" score on fees.

crwatkins:
Thanks to @jlegoupil for the original issue https://github.com/bitcoin-dot-org/bitcoin.org/issues/1255 that introduced this idea.  In response to @instagibbs [note](https://github.com/bitcoin-dot-org/bitcoin.org/issues/1255#issuecomment-198924011) I changed the topic of the scoring from just "delay" to the more inclusive "fees".

instagibbs:
awesome work (completely forgot my involvement)

jonasschnelli:
There is a QT fee bumper on the way for 0.15: https://github.com/bitcoin/bitcoin/pull/9697
Also, CPFP is kinda-possible via GUI if you enable `-spendzeroconfchange` and use coin-control 🤓

schildbach:
I think we should not rate "Full control" higher than the various forms of automatic. Full control also means complicated to use, not usable for non-technical people, etc. It's fine if there are power-user-wallets, but the others shouldn't be punished.

crwatkins:
@jonasschnelli I stand corrected on your Core GUI coin-control CPFP wizardry!

instagibbs:
@schildbach I don't see any wallets being punished for that, at least Coin.Space for example.

schildbach:
@instagibbs checkgood is rated higher than checkpass.

_templates/choose-your-wallet.html
```diff
@@ -105,6 +108,7 @@
           transparency: "checkpasstransparencyopensource"
           environment: "checkfailenvironmentdesktop"
           privacy: "checkgoodprivacyimproved"
+          fees: "checkpassfeecontroloverride"
```
achow101: Armory's next version (which will be released soon) will contain CPFP and RBF. Is that good enough to have it listed as full control now or should I wait and PR that after the release?

crwatkins:
@schildbach I certainly see your point, particularly for Core having to use a CLI option for RBF, however there are other wallets that make RBF/CPFP fairly simple, including your wallet!  Perhaps the "Full control" is poorly named?  It was named "Full" to be consistent with some of the other "top scores" in the other categories.  In this case, you'll see the complete text of "Full control" requires RBF or CPFP.  I can see that "Full control" might be taken as meaning the ability to set some arbitrary fee by the user (which is not what it means in this case).

achow101:
Looking at the prevous, it seems that Coinapault, Coinbase, and Xapo have no rating. Is that intentional?

crwatkins:
@achow101 Let us know as soon as it is released or just submit a PR then.   All of our scorings reflect released versions.  Thanks!

schildbach:
@crwatkins Yes indeed I thought full control means being able to type a fee amount. So I'm sorry for my misunderstanding.

crwatkins:
@schildbach Thanks!  You pointed out my poor choice of title.  I'll see if I can come up with something better.

Just as a note to meticulous reviewers:  There are two scores that are rated at the same "level" of "pass".  One involves providing dynamic fees and the other provides dynamic fees with the ability for the user to override them.

crwatkins:
@achow101 That is mostly due to my laziness; I just noticed that this morning also.  There has been some re-org recently related to custodial wallets, [Bitcoin banks](https://github.com/bitcoin-dot-org/bitcoin.org/issues/1109) and [exchanges](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1469) which may still be ongoing.  Until this morning, I thought [my recommendation](https://github.com/bitcoin-dot-org/bitcoin.org/issues/1109#issuecomment-268252161) to move them to the new category had already happened, but some of that still seems in progress.  In the meantime, I would be happy to see a PR to score them.

wbnns:
Unless others object, this will be merged on Sunday, April 23rd.

PR #1571, opened 2017-04-22
Bitcoin Core 0.14.1

laanwj:
Should be ready for merge if it passes travis


laanwj:
Travis complains that `bitcoin-0.14.1.torrent` cannot be found. 
Because one was not automatically generated (probably this script isn't active yet on the new server), I created my own torrent file and magnet link. I put this in the right place.
But somehow this Travis fails to find it. Does anyone know why? @wbnns @Cobra-Bitcoin

laanwj:
Oh, it didn't have the right file permissions. Should be accessible now.

PR #1592, opened 2017-05-09
Wallets: show up to 14 wallets

### LINKED ISSUE #1569: Coinomi wallet not visible under Android wallets
According to [the page](https://bitcoin.org/en/choose-your-wallet)'s source Coinomi should be visible under the Android section. Is there an issue with the web page?

Linking issue: https://github.com/Coinomi/coinomi-android/issues/146

harding:
Closes #1569

Preview: http://dg0.dtrt.org/en/choose-your-wallet

This uses @Mirobit's suggestion from the above issue to expand the display to show up to 14 wallets at a time, which is the maximum number of entries we have in any category except for the main page (which contais 18 wallets).

I've tested on Firefox 45 ESR for Linux, Chromium for Linux, and Chrome for Android.  Testing on other browsers appreciated.

In addition, wallet additions made after this PR is merged will cause a test to fail if any category begins to contain more than 14 wallets.  That way we shouldn't be surprised by wallets not being shown in the future.  We can discuss what to do about the problem then.

As part of the above, I made minor fixes to the entries for Electrum and Coin.Space that caused not all of their information to be displayed all of the time.  I added setting up JSON schema checker to my low-priority todo list so we can catch more of these sort of problems with automated testing in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,9 @@ pre-build-tests-fast: check-for-non-ascii-urls check-for-wrong-filename-assignme
     check-for-missing-copyright-licenses \
     check-bundle \
     check-for-english-in-en-dir \
-    check-for-consistent-bitcoin-core-titles
+    check-for-consistent-bitcoin-core-titles \
+    check-for-too-many-wallets-on-one-platform
+
 
 ## Post-build tests which, aggregated together, take less than 10 seconds to run on a typical PC
 post-build-tests-fast: check-for-build-errors ensure-each-svg-has-a-png check-for-liquid-errors \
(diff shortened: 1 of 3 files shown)


Mirobit:
@harding Thanks for picking that up. LGTM.

crwatkins:
Thanks @harding for this badly needed update.  Also thanks for the added sanity checks; I've been concerned about those interactions for a while now.

I've tested on Chrome and Safari browsers on iPhone, iPad, and Mac OS.  I noticed a missing wallet in certain UX sequences on iOS (and Android) that I need to research more and submit a separate issue since these same problems are reproducible on the currently deployed bitcoin.org and do not seem to be related to this update.

LGTM.

harding:
Looked into the issue @crwatkins mentioned (thanks!) and opened issue #1593 for it since it's not a bug introduced by this branch.  I'll try to look into more later today.

wbnns:
Thanks all. Unless others object, this will be merged on Sunday, May 14th (tomorrow).

PR #1493, opened 2017-01-27
added CoinCorner

brollymook:
UK Bitcoin Exchange

PR #1594, opened 2017-05-10
Wallets: fix accidental wallet hiding on category change

harding:
Closes #1493

Preview: http://dg1.dtrt.org/en/choose-your-wallet

Currently when you load the page in a mobile browser and click on a wallet, a copy of that that wallet's entry is copied using JS to a box later in the page where you can read about the wallet details.  When you change platform, all the currently displayed wallet icons are cleared and then refilled with the icons for the newly-selected platform, with a check running that ensure each wallet is only displayed once (which is important for combination categories such as "mobile" where we don't want to show the same wallet twice just because it has an Android and iOS version).

Since the wallet details box in the mobile display wasn't cleared when you changed platform, the check that prevents duplicate wallet displays thought that wallet had already been included in the listing and so prevented it from being displayed again.

This patch fixes the problem by giving the copy of the wallet entry that goes in the wallet details box its own id, which seems to have resolved the issue and has no side effects I've seen.

I've tested on Firefox 45 ESR for Linux and Chromium for Linux, both in mobile view, and on Chrome for Android on a device.  More testing always welcome.

Thanks to @crwatkins who first reported this issue.

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -389,6 +389,7 @@ function walletListener(e) {
       if (isMobile()) {
         var p = document.getElementById('walletsmobile');
         t = t.cloneNode(true);
+        t.id = 'inline_wallet_box';
         p.innerHTML = '';
         p.appendChild(t);
         scrollToNode(p);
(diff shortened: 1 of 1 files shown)


crwatkins:
ACK.   Tested on iOS in Chrome and Safari.

wbnns:
Thanks guys!

PR #1596, opened 2017-05-13
Full node guide: describe blocks-only mode

### LINKED ISSUE #1544: Add -blocksonly mode to reduce traffic section
In extreme bandwidth limited situations, -blockonly can reduce the bandwidth requirement to about 144MB per day, at the cost of not seeing transactions until they're confirmed.

https://bitcoin.org/en/full-node#reduce-traffic

This thread (https://bitcointalk.org/index.php?topic=1377345.0) talks about an 88% reduction but non-blocksonly will have improved because compact blocks has been created since then.

harding:
Closes #1544 (CC: @chris-belcher )

I tried checking my assertions about bandwidth usage by starting my node with `-blocksonly` and then using `bitcoin-cli getnettotals`, but the bandwidth reported used by that was 20% *less* than the combined size of the blocks I downloaded during the 90 minute test according to `bitcoin-cli getblock <hash> | jq .size`.

It's possible this is for some reason I don't know, but my guess is that it's because I had a saved mempool from before my restart so my node was able to make use of compact blocks.  However, the overhead for the upload bandwidth is consistent with the 1 MB I estimated and so I think we're fine saying 150 MB download bandwidth to account for changing peer connections and occasional stale blocks, plus the occasional fun times when someone running Bitcoin Unlimited produces and relays invalid blocks with sufficient PoW that we download them.

### DIFF
--- a/en/full-node.md
+++ b/en/full-node.md
@@ -1310,7 +1310,7 @@ blocks could be served to other nodes.
 
 ### Reduce Traffic
 
-Some node operators need to deal with bandwith caps imposed by their ISPs.
+Some node operators need to deal with bandwidth caps imposed by their ISPs.
 
 By default, bitcoin-core allows up to 125 connections to different peers, 8 of
 which are outbound. You can therefore, have at most 117 inbound connections.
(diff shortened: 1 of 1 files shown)


wbnns:
Thanks Mr. Dave.

Unless others object, this will be merged on Sunday, May 14th (tomorrow).

PR #1459, opened 2016-12-25
Adding ArcBit wallet

stequald:
Hi, I would like to add the bitcoin wallet ArcBit to Bitcoin.org.

PR #1600, opened 2017-05-16
Wallets: validate wallet YAML file using JSON Schema

harding:
As mentioned in #1592, I wanted to validate the 1,100-line wallet YAML to ensure consistency between entries.  I also wanted to better document the entry format since even I sometimes wonder which option does what.

This PR validates the wallet entries against a pretty strict schema.  It's not perfect---I've put in some TODOs where we could be more strict with changes to the current format---but I think it's better than nothing.

**Some examples:**

In #1592, I discovered Electrum declared an `android` platform without declaring a default `mobile` platform, so it wasn't being counted as a mobile wallet.  If that problem hadn't already been fixed, the schema would've caught it:

```text
user@devbco:~/bitcoin.org$ make check-validate-yaml 
The property '#/wallets/4/electrum/platform' has a property 'android' that depends on a missing property 'mobile' in schema https://bitcoin.org/quality-assurance/schemas/wallets.yaml
Makefile:299: recipe for target 'check-validate-yaml' failed
make: *** [check-validate-yaml] Error 1
```

In this PR, I've fixed two issues.  The first was that Copay declared "Linux" as a supported OS on its Mobile platform entry.  I checked #888 where we added Copay and that was never discussed, so I think it's just a copy/paste typo.  The schema reported that:

```text
user@devbco:~/bitcoin.org$ make check-validate-yaml 
The property '#/wallets/7/copay/platform/mobile/os/3' value "linux" did not match one of the following values: android, blackberry, ios, windowsphone in schema https://bitcoin.org/quality-assurance/schemas/wallets.yaml
Makefile:299: recipe for target 'check-validate-yaml' failed
make: *** [check-validate-yaml] Error 1
```

The other issue was that ArcBit was missing the `privacynetwork` field on just one of its supported platforms (Android).  Again, I think this was a typo because I don't see it discussed in the PR where ArcBit was added, #1459 

```text
user@devbco:~/bitcoin.org$ make check-validate-yaml 
The property '#/wallets/26/arcbit/platform/android/privacycheck' did not contain a required property of 'privacynetwork' in schema https://bitcoin.org/quality-assurance/schemas/wallets.yaml
Makefile:299: recipe for target 'check-validate-yaml' failed
make: *** [check-validate-yaml] Error 1
```

In addition to the validation, this PR also links to the schema from the documentation for adding a new wallet.

### DIFF
--- a/quality-assurance/schemas/wallets.yaml
+++ b/quality-assurance/schemas/wallets.yaml
@@ -20,6 +20,13 @@ properties:
     description: The page identifier to use for translations
     type: string
 
+  max_wallet_text_length:
+    description: >
+      The global maximum length for a wallet text description in any
+      langugage (counted in Unicode characters).  Ensures text doesn't
+      overfill its box.
+    type: integer
+
   ## The wallets array
   wallets:
     description: The array containing all wallets
@@ -102,7 +109,7 @@ properties:
                 - desktop
             properties:
               desktop:
-                description: Wallets than run on desktop operating systems
+                description: Wallets that run on desktop operating systems
                 type: object
                 required: &desktop-required
                   - text
--- a/quality-assurance/schemas/wallets.yaml
+++ b/quality-assurance/schemas/wallets.yaml
@@ -1,0 +1,311 @@
+---
+# This file is licensed under the MIT License (MIT) available on
+# http://opensource.org/licenses/MIT.
+id: https://bitcoin.org/quality-assurance/schemas/wallets.yaml
+description: The main container
+type: object
+required:
+  - layout
+  - id
+  - wallets
+additionalProperties: false
+properties:
+  layout:
+    description: The Jekyll page layout to use
+    type: string
+    enum:
+      - base
+
+  id:
+    description: The page identifier to use for translations
+    type: string
+
+  ## The wallets array
+  wallets:
+    description: The array containing all wallets
+    type: array
+    uniqueItems: true
+    items:
+      description: The wallets.  Each value is an arbitrary ID
+      type: object
+      additionalProperties:  ## matches anything since this is an arbitrary ID string
+        required:
+          - title
+          - titleshort
+          - compat
+          - level
+          - platform
+        additionalProperties: false
+        properties:
+          title:
+            description: The full wallet name; displayed when you view the full listing
+            type: string
+            minLength: 1
+            maxLength: 100
+
+          titleshort:
+            description: A short name displayed on the overview page
+            type: string
+            minLength: 1
+            maxLength: 20
+
+          ## TODO: change this into an array so we can practically test values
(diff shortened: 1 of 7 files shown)


quality-assurance/schemas/wallets.yaml
```diff
@@ -0,0 +1,311 @@
...
+                - mobile
+              windows:
+                - desktop
+              linux:
+                - desktop
+              mac:
+                - desktop
+            properties:
+              desktop:
+                description: Wallets than run on desktop operating systems
```
crwatkins: Typo:  than->that

crwatkins:
Concept ACK:  This is the best present a wallet maintainer could wish for.  I spend quite a bit of time reviewing the syntax of submissions and, more often than not, I have to have more than one offline cycle with submitters to fix the YAML.  In addition, we finally have a schema to refer submitters to so that they have a better chance of getting it right in the first place.  I'm excited to no longer have to check string lengths myself by hand.

Scoring change in choose-your-wallet.html:  ACK

wallets.yaml schema:  ACK (with slight typo comment above)

Thanks very much @harding 

PR: utACK

harding:
@crwatkins your comment made me realize that the initial commit doesn't limit the length of the main string, the wallet description, because that's pulled from the translation YAML (e.g. `en.yml`) rather than the wallet YAML and it has an arbitrary variable name for each wallet and so can't easily be checked by a JSON schema without structural changes to the translation YAML file

Instead, I've pushed a commit that causes the template compile to fail if that field is too long, as well as a setting on the wallets page that lets us adjust it.  Here's an example of the error produced:

```text
Liquid die tag called. Wallet text too long  -- Error creating output
  Liquid Exception: Liquid die tag called. Wallet text too long -- Error creating output in fr/choisir-votre-porte-monnaie.html
```

I set the max to 418 characters, which is the longest description we have in any translation at present (BitGo in French).  This is rather a bit longer than we've encouraged for English, I believe, but:

1. I have an idea to increase the space available for detailed wallet information in the future.  Hopefully it's within my meager JS skills/patience to implement.

2. The ideal way to test these values is not with character limitations but by checking the rendered variable-width text size.  For that we need something like Selenium, which is something I've done elsewhere and would like to do here eventually (but it's a big change to the testing framework).

3. The poor French have about 20% unpronounced characters in most of their words and a profusion of short words which require lots of spaces between them.  Who can blame them for long wallet descriptions? :-D

For testing this newest commit, I diffed the resultant HTML between this commit and the previous commit and verified there were only whitespace changes.

In the new commit, I also fixed the typo that @crwatkins identified (thanks!).

crwatkins:
utACK

wbnns:
Unless others object, this will be merged on Saturday, May 20th.

Thanks guys.

PR #1602, opened 2017-05-17
Fixed Missing word in guide_wallets.md

nikolaifischer:
"work" was missing in the first sentence describing the point() function.

### DIFF
--- a/_includes/devdoc/guide_wallets.md
+++ b/_includes/devdoc/guide_wallets.md
@@ -423,7 +423,7 @@ point (the public key):
 
 {% autocrossref %}
 
-Because of the way `point()` functions, it's possible to create a [child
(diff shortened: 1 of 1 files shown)


harding:
I think when I wrote that, I meant "functions" to be used as a verb rather than a noun.  If you'd like, I'd support replacing `functions` with `works` in the original (i.e. dropping `functions` in your version.).

Sorry for the confusion, and thanks for the PR.

nikolaifischer:
Ah ok, that makes much more sense! Yeah, maybe replacing it would make it clearer. 
Thanks :)

harding:
ACK 266315acca46cee4a48b276c93606c0e9183067b .  Thanks, @nikolaifischer!

PR #1603, opened 2017-05-17
Update simplified-payment-verification.yaml

r3ticuli:
fixed spv definition typo


harding:
I think the original is more correct.  Maybe we could change "were" to "are" to make it more clear?  (Thanks for the PR, though.)

harding:
ACK 165baa0ee1c9532dcd7bba01f538f883c5ae4f90 .  Thanks!

Issue #1604, opened 2017-05-18
Headers message may be empty.


evoskuil:
Source File: _includes/devdoc/ref_p2p_networking.md

This statement implies that empty headers messages are not sent, which is not the case:

> The headers message sends one or more block headers to a node which previously requested certain headers with a getheaders message.

It is typical for peers to send an empty headers message in response when the peers are in sync.

wbnns:
[Bounty](https://github.com/bitcoin-dot-org/bitcoin.org#earn-bitcoin-for-contributing): **5,000 bits / ~$10 USD**

evoskuil:
This issue may apply to the wording of the INV message as well, though I haven't verified the behavior.

Issue #1520, opened 2017-02-28
rescan an existing address


tal-beja:
Source File: _includes/devdoc/bitcoin-core/rpcs/rpcs/importaddress.md

> Notes: if the address or pubkey script is already in the wallet, the block database will not be rescanned even if this parameter is set

This is not really true, if you try to import the same address twice with (or without) the rescan flag set to true, it rescan twice, and not only for the first time as this docs suggests.

wbnns:
[Bounty](https://github.com/bitcoin-dot-org/bitcoin.org#earn-bitcoin-for-contributing): **5,000 bits / ~$10 USD**

PR #1611, opened 2017-05-22
Various corrections to developer reference

### LINKED ISSUE #1505: Reject nonstandard tx message may not have hash.
Location: https://bitcoin.org/en/developer-reference#reject

> The transaction will not be mined or relayed because the rejecting node considers it non-standard—a transaction type or version unknown by the server. Extra data is the rejected transaction’s TXID.

This does not appear to be consistently applied by Satoshi nodes. We periodically see no bytes in the supposed place of the tx hash. For example, this message decodes as follows: 

```
/Satoshi:0.12.1/
message payload: 0x0274784016746f6f2d6c6f6e672d6d656d706f6f6c2d636861696e
tx : nonstandard : too-long-mempool-chain : <empty>
```
I haven't dug through their sources to see if/when this is the case, but it's not reliably seen on the wire.

achow101:
Corrected Compact Size Unsigned Integer description to include the lower bounds.

Changed reject message code descriptions of the extra data field to be more accurate as the extra data is not always what was specified nor is the field required.

Clarified the empty message headers are allowed.

Removed incorrect note in importaddress which stated that addresses would not be rescanned if already in the wallet. They are rescanned.

Closes #1505, #1365, #1604, #1520 

If this is accepted, please pay the bounty to 1AjFwHanPewurVms4Z3ExtapHmKWF6aTVS

### DIFF
--- a/_includes/devdoc/bitcoin-core/rpcs/rpcs/importaddress.md
+++ b/_includes/devdoc/bitcoin-core/rpcs/rpcs/importaddress.md
@@ -41,7 +41,7 @@ The `importaddress` RPC {{summary_importAddress}}
 - n: "Rescan"
   t: "bool"
   p: "Optional<br>(0 or 1)"
-  d: "Set to `true` (the default) to rescan the entire local block database for transactions affecting any address or pubkey script in the wallet (including transaction affecting the newly-added address or pubkey script).  Set to `false` to not rescan the block database (rescanning can be performed at any time by restarting Bitcoin Core with the `-rescan` command-line argument).  Rescanning may take several minutes.  Notes: if the address or pubkey script is already in the wallet, the block database will not be rescanned even if this parameter is set"
+  d: "Set to `true` (the default) to rescan the entire local block database for transactions affecting any address or pubkey script in the wallet (including transaction affecting the newly-added address or pubkey script).  Set to `false` to not rescan the block database (rescanning can be performed at any time by restarting Bitcoin Core with the `-rescan` command-line argument).  Rescanning may take several minutes."
 
 {% enditemplate %}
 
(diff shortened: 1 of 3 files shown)


harding:
@achow101 in and of themselves, these changes LGTM.  I was wondering if you checked the implementation behavior to confirm all of those issues were accurate?  (I'm not saying you have to, I don't know what @wbnns's bounty rules are, but I usually like to confirm the problems before editing the docs.)

In particularly, I'm surprised by #1520 (rescan) as I don't think I would've added the extra sentence there unless I believed it was true based on my reading of the code or my testing (though it's possible the implementation has changed since I wrote it, or that I was wrong when I wrote it, or that I'm wrong now in guessing at my mental state 2.5 years ago when I wrote it).

Also #1365 (compactSize canonical encoding) has a link to a Bitcoin Core dev saying that the sizes weren't previously enforced, suggesting to me that there may be some subtleties in the behavior or maybe that this was even an unknown soft fork.

wbnns:
@achow101 Thanks for tackling this stuff! @harding Thanks also for the feedback.

Unless others object, this will be merged on Thursday, May 26th.

achow101:
@harding 

I did verify the behavior and checked in the code.

For the rescan functionality, it looks like the behavior did change with this commit https://github.com/bitcoin/bitcoin/commit/983d2d90af1b517bee51170d2ea059e68d09be35 so it no longer did nothing if the address/script already existed.

For the CompactSize uint encoding, that behavior changed with this commit: https://github.com/bitcoin/bitcoin/commit/8dc206a1e2715be83912e039465a049b708b94c1 to enforce canonical encoding.

harding:
@achow101 awesome!  Thanks!

wbnns:
@achow101 [Merged; Bounty paid together with #1610](https://live.blockcypher.com/btc/tx/b36645ba594b15d26de7fdf70e0bb53027b200e34dbb332198a25b3d8a80e05f/). 👍

PR #1620, opened 2017-06-03
Events plugin: stop requesting events from Meetup.com

### LINKED ISSUE #1617: Events page no longer displays Bitcoin meetups
I was tracing down another issue and I noticed the that Meetup API was failing for us.  It appears that they've changed to requiring an API key.  Here's the URL we currently use:

http://api.meetup.com/2/open_events?omit=description&status=upcoming&radius=25.0&topic=bitcoin&and_text=False&limited_events=False&desc=False&offset=0&format=json&page=500&time=0m%2C3m&sig_id=133622112&sig=cd874bc2c84f96d989f823880889bda2f5e4cdc5

And here's what it says when accessed:

    {"details":"API requests must be key-signed, oauth-signed, or accompanied by a key: https:\/\/www.meetup.com\/meetup_api\/docs\/#authentication","code":"not_authorized","problem":"You are not authorized to make that request"}

Fixing this would be kind of annoying because we wouldn't want to put the access key in the code itself (we'd probably use an environmental variable accessible on the server), but if we don't put the key in the code itself, we can't test with Travis CI or on any random contributor's machine.  Not that testing actually caught this bug whenever it started to fail.

I think the current [events page](https://bitcoin.org/en/events) is just fine without the meetups (and we do have a working link to Bitcoin meetups on the bottom of that page), so I suggest that we just drop that part of the [events plugin](https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/_plugins/events.rb).  (I suspect my other issue is in that part of the code also.)  I'm happy to make and test that change if there's agreement on that being a reasonable path forward.

harding:
Closes #1617 

In testing, the only rendered HTML changes to the site were three fewer lines of whitespace per rendered translation of the Events page.

I also checked the Events RSS feed and it appeared that we never included any Meetups information in it, so no changes were required there.

### DIFF
--- a/_plugins/events.rb
+++ b/_plugins/events.rb
@@ -1,8 +1,8 @@
 # This file is licensed under the MIT License (MIT) available on
 # http://opensource.org/licenses/MIT.
 
-#events.rb set site.conferences and site.meetups arrays based
-#on events in _events/ and meetups on bitcoin.meetups.com .
+#events.rb sets the site.conferences array based
+#on events in _events/
 #This is later used to populate the events map and display the
 #list in chronological order, in the RSS file and events pages.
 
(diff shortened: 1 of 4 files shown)


wbnns:
@harding Thanks

Unless others object, this will be merged on Wednesday, June 7th.

Mirobit:
@harding Submitted a pull request updating the docs regarding the removal of meetup.

harding:
@Mirobit thanks for catching that!  I added your commit a492959763938a8aa91b4954734017801262f024 to this branch, but GitHub isn't showing it here (and isn't showing it as merged on your PR), even though it does show on the [branch itself](https://github.com/harding/bitcoin.org/tree/events-drop-meetup).  I think this is a bug in GitHub, so I'll wait a few hours to see if it resolves itself before poking at it.

Thanks again!

harding:
Seems to be mostly working now.  @Mirobit's commit is showing on this page and in the diff view, but Travis is not checking it (not that it should---the internal site docs are not checked).

@wbnns I think this remains ready for Wednesday merge, baring objections, but please feel free to move back the planned merge if you want to give others a chance to review the additional commit.

PR #1624, opened 2017-06-08
Use shiny flags

Mirobit:
I think it gives the pages a more modern look.
Changes [Community](https://bitcoin.org/en/community#non-profit) and [Bitcoin Exchanges](https://bitcoin.org/en/exchanges)

**Old:**
old

 **New:**
new


harding:
Old combined size of all flags was 19,102 bytes; new combined size is 60,211 bytes.  A 40KB increase in just two pages seems perfectly fine to me, and I agree the new flags look nicer.  I tested with `optipng` per site policy to ensure flags were maximally compressed, and they were, so this LGTM.

Thanks, @Mirobit!

wbnns:
@Mirobit @harding LGTM. 👍

Unless others object, this will be merged on Sunday, June 11th.

Issue #1622, opened 2017-06-07
Remove Coinomi wallet from "Choose your Bitcoin wallet" page


Bitim:
This wallet is shady, and its "Basic transparency" is practically none.

The latest commit in their [git repo](https://github.com/Coinomi/coinomi-android/commits/master) made on February 6th, 2017, when they change their license from Apache to some custom license (the last actual change made on March 6th, **2016**).

On the other hand, the latest version on [Google Play Store](https://play.google.com/store/apps/details?id=com.coinomi.wallet&hl=en), released on June 3th, **2017**.

At least the "Basic transparency" should be change to **none**, if not removing this wallet from recommendations.

Thanks.

Mirobit:
I agree that transparency should be set to 'checkfailtransparencyclosedsource'. Last real code update on GitHub has been in April 2016.

crwatkins:
@erasmospunk Could you push your recent updates and tag the latest release?

GeorgeKimionis:
There is indeed a delay in publishing the code, however this decision was taken after being accused by hundreds of users that we stole their funds, as they funded their wallets and a bit later their funds were gone. 

Apparently scammers were cloning Coinomi wallet and then run campaigns, MLM schemes, they even organized proper presentations where they had the investors download the malicious clones and fund their wallets, and after a few weeks their funds were simply gone. They didn't even bother to change the Support link inside Coinomi's menu which was still forwarding all inquiries to our Support e-mail, which is basically why we are aware that so many users were scammed and lost their money and along with that their faith to crypto-currencies. Is this really the message we want to pass on to the new adopters? It's our duty to welcome new users to the cryptoland and protect them by all means.

On top of that, delaying the publishing of the source code also puts a hold to parasitic ventures like uberpay.io who basically built their revenue model on top of a clone of an older version of our wallet and  didn't even bother to run their own back-end servers, last time we checked they were still parasitizing on our back-end servers, which are paid by us, of course. And uberpay is just an example, there are many more like them who have been making a buck on our sweat. This kind of behavior only harms legit projects like Coinomi which has been serving hundreds of thousands of users for almost 4 years, and to achieve that we have also spent hundreds of thousands of USD in development and operational costs apart from the time and effort we put into this, so maybe we should focus on what is really harming our community instead of going those who have been helping cryptos evolve since day one.

[...]

PR #1627, opened 2017-06-13
Change Coinomi wallet link to solve issue #1622

erasmospunk:
As the policy of Bitcoin.org does not allow closed source
wallets that allow user private key access, we are changing
the link for the Coinomi Android wallet to point to the
Github release page, that contains the latest open source
version along with the necessary source code.

Additionally we revert the transparency to "checkpasstransparencyopensource"
as the source code is available from the same site.

### DIFF
--- a/_templates/choose-your-wallet.html
+++ b/_templates/choose-your-wallet.html
@@ -740,7 +740,7 @@ wallets:
     platform:
       mobile:
         text: "walletcoinomi"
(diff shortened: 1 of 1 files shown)


Mirobit:
I am not sure it solves the issue. People might just look for Coinomi in the app store. Very few people will actually install the app from Github.

PR #1637, opened 2017-06-17
Bitcoin Core 0.14.2

laanwj:


laanwj:
Not sure why Travis is failing: it's missing `/bin/bitcoin-core-0.14.2/bitcoin-0.14.2.torrent` but I'm sure I've put it there.

harding:
@laanwj can you check the file permissions?  I get an access denied error (looks like just that file):

laanwj:
Updated the file permissions and re-pushed, good catch!

PR #1647, opened 2017-06-27
Add BitMoney to Europe exchangers

pjendursik:


Mirobit:
NACK. Exchange is quite new and I couldn't find any feedback.

PR #1650, opened 2017-06-29
Exchanges fixes

Mirobit:
* Whitespace fix
* VE shiny flag
* Removed Germany and France since the exchanges are listed under Europe (SEPA)


wbnns:
Thanks, LGTM.

PR #1651, opened 2017-06-29
Make preview charset fix

Mirobit:
Since the incremental build (#1643) update "make preview" leads to the following error:
` Conversion error: Jekyll::Converters::Scss encountered an error while converting 'css/main.scss':
                    Invalid US-ASCII character "\xE2" on line 2427
jekyll 3.0.1 | Error:  Invalid US-ASCII character "\xE2" on line 2427
`
I don't know if this is the best solution but it fixes the error.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ default: clean build
 
 ## `make preview`: start the built-in Jekyll preview
 preview: clean
-	$S bundle exec jekyll serve --incremental
+	$S export LANG=C.UTF-8 ; bundle exec jekyll serve --incremental
 
 ## `make test`: don't build, but do run all tests
 test: pre-build-tests post-build-tests


harding:
Concept ACK and this should be safe for immediate merge since it's just changing the non-production preview target..  I'm testing now---I should have results in about 30 minutes.

Sorry, I didn't attempt to test incremental building outside my usual UTF-enabled environment.  I think this solution is ideal given that the regular `build` target does exactly the same thing, and that's what we use in production.

harding:
Tested ACK now.  This change renders the CSS files (and all other files) in the `preview` target as in the production `build` target.

wbnns:
@Mirobit Nice catch! @harding Thanks for double-checking.

Merged. 👍

PR #1665, opened 2017-07-05
exchanges: Add Kiwi-Coin (New Zealand)

vitalkadrug:
Hi, guys!

Please, accept my changes for the exchange.html page that now contains info about New Zealand.

Yours faithfully,
Vitaly.


Mirobit:
I couldn't find any positive reviews but also no negative. The NZ Bitcoin subreddit links to kiwi-coin and it is the only NZ exchange with somewhat of a volume.

@vitalkadrug You state that Kiwi-Coin.com was created in 2014, but the earliest record I could find is from 2016. Was kiwi-coin.com private till 2016?

Slight ACK if my question is answered.

Mirobit:
@vitalkadrug Thanks. ACK from my side if the flag is fixed.

wbnns:
Unless others object, this will be merged on Sunday, August 6th.

PR #1671, opened 2017-07-08
WIP: Add Docker Compose for local dev

conortm:
This proposed change adds support for running the site locally simply by running `docker-compose up`, eliminating the need to install anything locally (except [Docker](https://www.docker.com/)).

Considering this a "work in progress", as I have only tested on a Mac. It _should_ work everywhere, but I don't want to assume. Any assistance with verifying on Windows/Linux would be much appreciated!

Please let me know if there is interest in this as a feature, and if I can provide more details/context.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,5 @@
+FROM ruby:2.0.0-onbuild
+
+EXPOSE 4000
+
+ENTRYPOINT ["make"]
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,11 @@ default: clean build
 preview: clean
 	$S export LANG=C.UTF-8 ; bundle exec jekyll serve --incremental
 
+## `make preview-docker`: start the built-in Jekyll preview, with proper
+## host setting for running in Docker container
+preview-docker: clean
+	$S export LANG=C.UTF-8 ; bundle exec jekyll serve --incremental --watch -H 0.0.0.0
+
 ## `make test`: don't build, but do run all tests
 test: pre-build-tests post-build-tests
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,5 @@
-version: '3'
+version: "2"
+
 services:
   web:
     build: .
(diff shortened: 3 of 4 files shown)


wbnns:
@conortm This is awesome - nice work! I will test and let you know.

wbnns:
@conortm Hey, just following up. In Ubuntu 16.04 I get the following output when running `docker-compose up -d`:
```
ERROR: Version in "./docker-compose.yml" is unsupported. You might be seeing this error because you're using the wrong Compose file version. Either specify a version of "2" (or "2.0") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/```

conortm:
@wbnns Thanks for testing this out!

As for the version change, [upgrading docker](https://docs.docker.com/compose/compose-file/) would fix that. However, as we don't need any version 3 -specific functionality, I changed the `docker-compose.yml` file version from 3 to 2.

As for the second error, you def shouldn't need `sudo`. Did you try some of the solutions mentioned here: https://github.com/docker/compose/issues/4181? Perhaps restarting the daemon, or setting `DOCKER_HOST`, or reinstalling docker? I'll try spinning up a VM to try this out myself and report back.

Thanks again!

conortm:
Also, Docker Compose is kinda overkill, as we just have one service -- it really just provides the convenience of one command (`docker-compose up`) instead of two (`docker build ...` and `docker run ...`). I'd be open to dropping Compose altogether and just updating the README steps accordingly, but I'm curious what other folks' thoughts are. Thanks!

cjennyc28:
Done

wbnns:
@conortm Heya! If you're still interested in adding support for Docker, let's go ahead and drop Docker Compose and use the build and run commands instead, since we just have one service.

wbnns:
@conortm I'm closing for now; just let me know if you want to finish it up and we can reopen.

PR #1674, opened 2017-07-12
Alerts: add warning for potential 2017-08-01 BIP148 split

harding:
This PR adds a warning for the chainsplit that may occur on 1 August 2017.  It's focused on describing to non-expert users what they should do to secure their funds and does not advocate any particular position on the fork or attempt to describe the background of the situation.

The warning is displayed in orange above every regular page on Bitcoin.org; a comment in the file suggests changing the warning to red 72 hours before the beginning of the potential chain split.  I will submit a PR for that on the 28th.

I will do my best to submit PRs with changes to the page on and after August 1st, although of course anyone else can submit a PR to Bitcoin.org too.

Preview:

### DIFF
--- a/_alerts/2017-07-12-disruptive-forks.md
+++ b/_alerts/2017-07-12-disruptive-forks.md
@@ -5,17 +5,20 @@
 title: "Potential disruptive chain forks"
 shorturl: "disruptive-forks"
 active: true
-banner: "Warning: Bitcoin may be unsafe to use on August 1st (click here for details)"
+banner: "Warning: Bitcoin may be unsafe to use starting July 31st (click here for details)"
 
 ## 7/12 to 7/28: "warning" (orange); 7/29 until resolution: "alert" (red)
 bannerclass: "warning"
 ---
-*Last updated: 2017-07-12 12:00 UTC.  This page will be updated when new
-information becomes available.  See the [list of updates][].*
+{% assign start='<span class="date">2017/08/01 00:00 UTC</span>' %}
 
-On 1 August 2017, Bitcoin confirmation scores may become unreliable for
+*Last updated: <span class="date">2017/07/12 12:00 UTC</span>.  This
+page will be updated when new information becomes available.  See the
+[list of updates][].*
+
+At {{start}}, Bitcoin confirmation scores may become unreliable for
 an unknown length of time.  This means that any bitcoins you receive
-after August 1st may disappear from your wallet at a later time or be a
+after that time may later disappear from your wallet or be a
 type of bitcoin that other people will not accept as payment.
 
 Once the situation is resolved, confirmation scores will either
@@ -25,7 +28,7 @@ return to using Bitcoin normally; in the later case, you will need to
 take extra steps in order to begin safely receiving bitcoins again.
 
 This post currently describes what actions you can take to prepare for
-this situation.  Subsequent to August 1st, we will update this post as
+this situation.  Subsequent to {{start}}, we will update this post as
 best we can with relevant information, but you are also advised to
 monitor other Bitcoin [news sites][] and [community resources][] for
 updates and to cross-check all information, as someone may attempt to
@@ -35,15 +38,15 @@ Remember that you alone are responsible for the safety of your bitcoins,
 and that if you lose control of them for any reason, there is nothing
 the operators or contributors to this website can do to help you.
 
-## Before August 1st
+## Preparation
 
 1. If you accept bitcoins as payments, we recommend that you stop
-accepting Bitcoin payments at least 12 hours before 00:00 UTC on August
-1st (24 to 48 hours may be safer).  This will give time for all pending
-payments to confirm on the Bitcoin block chain before the event.
+accepting Bitcoin payments at least 12 hours before {{start}}, although
+24 to 48 hours earlier may be safer.  This will give time for all
+pending payments to confirm on the Bitcoin block chain before the event.
 
 1. If you send bitcoins as payments, note that many services may stop
-accepting bitcoins at 00:00 UTC on August 1st or earlier.
+accepting bitcoins at {{start}} or earlier.
 
 1. Be wary of storing your bitcoins on an exchange or other third-party
 wallet.  If they accept transactions during the event, they could lose
@@ -51,9 +54,9 @@ money and will likely spread those loses across all their users.  If
 there end up being two or more competing versions of Bitcoin, then they
 may refuse to give you your bitcoins on versions they don't like.
 
-## On August 1st
+## During the event
 
-1. Do not trust any payments you receive after August 1st until the situation
+1. Do not trust any payments you receive after {{start}} until the situation
 is resolved.  No matter how many confirmations the new payment says it
 has, it can disappear from your wallet at any point up until the
 situation is resolved.
@@ -66,7 +69,7 @@ recovery if you attempt this.
 by "splitting" your coins.  Some of these offers may be scams, and
 software claiming to split your coins can also steal them.
 
-## After August 1st
+## After the event
 
 No information available yet.  Please wait for multiple trustworthy
 sources to state that the event is resolved before returning to normal
@@ -78,9 +81,24 @@ A [full history][] of this document is available.  The following points
 summarize major changes, with the most recent changes being listed
 first.
 
-- 2017-07-12 12:00 UTC: initial version.
+- <span class="date">2017/07/12 12:00 UTC</span>: initial version.
 
 [full history]: https://github.com/bitcoin-dot-org/bitcoin.org/commits/master/_alerts/2017-07-12-disruptive-forks.md
 [list of updates]: #document-history
 [news sites]: /en/resources#news
 [community resources]: /en/community
+
+<script src="/js/jquery/jquery-1.11.2.min.js"></script>
+<script>
+// Localize dates
+$(".date").each(function() {
+  // Try to parse the string as a date
+  epoch = Date.parse($(this).text());
+  // Only convert the string to localtime if it's a number
+  if (isNaN(epoch) == false) {
+    var utcdate=new Date(epoch);
+    var localedate = utcdate.toString();
+    $(this).text(localedate);
+  }
+});
+</script>
--- a/_alerts/2017-07-12-disruptive-forks.md
+++ b/_alerts/2017-07-12-disruptive-forks.md
@@ -48,11 +48,12 @@ pending payments to confirm on the Bitcoin block chain before the event.
 1. If you send bitcoins as payments, note that many services may stop
 accepting bitcoins at {{start}} or earlier.
 
-1. Be wary of storing your bitcoins on an exchange or other third-party
-wallet.  If they accept transactions during the event, they could lose
-money and will likely spread those loses across all their users.  If
-there end up being two or more competing versions of Bitcoin, then they
-may refuse to give you your bitcoins on versions they don't like.
+1. Be wary of storing your bitcoins on an exchange or any service that
+doesn't allow you to make a local backup copy of your private keys.  If
+they accept transactions during the event, they could lose money and
+will likely spread those loses across all their users.  If there end up
+being two or more competing versions of Bitcoin, then they may refuse to
+give you your bitcoins on versions they don't like.
 
 ## During the event
 
--- a/_alerts/2017-07-12-disruptive-forks.md
+++ b/_alerts/2017-07-12-disruptive-forks.md
@@ -1,0 +1,86 @@
+---
+## This file is licensed under the MIT License (MIT) available on
+## http://opensource.org/licenses/MIT.
+
+title: "Potential disruptive chain forks"
+shorturl: "disruptive-forks"
+active: true
+banner: "Warning: Bitcoin may be unsafe to use on August 1st (click here for details)"
+
+## 7/12 to 7/28: "warning" (orange); 7/29 until resolution: "alert" (red)
+bannerclass: "warning"
+---
+*Last updated: 2017-07-12 12:00 UTC.  This page will be updated when new
+information becomes available.  See the [list of updates][].*
+
+On 1 August 2017, Bitcoin confirmation scores may become unreliable for
+an unknown length of time.  This means that any bitcoins you receive
+after August 1st may disappear from your wallet at a later time or be a
+type of bitcoin that other people will not accept as payment.
+
+Once the situation is resolved, confirmation scores will either
+automatically return to their normal reliability or there will be two
+(or more) competing versions of Bitcoin.  In the former case, you may
+return to using Bitcoin normally; in the later case, you will need to
+take extra steps in order to begin safely receiving bitcoins again.
+
+This post currently describes what actions you can take to prepare for
+this situation.  Subsequent to August 1st, we will update this post as
+best we can with relevant information, but you are also advised to
+monitor other Bitcoin [news sites][] and [community resources][] for
+updates and to cross-check all information, as someone may attempt to
+spread false news in order to exploit the situation.
+
+Remember that you alone are responsible for the safety of your bitcoins,
+and that if you lose control of them for any reason, there is nothing
+the operators or contributors to this website can do to help you.
+
+## Before August 1st
+
+1. If you accept bitcoins as payments, we recommend that you stop
+accepting Bitcoin payments at least 12 hours before 00:00 UTC on August
+1st (24 to 48 hours may be safer).  This will give time for all pending
+payments to confirm on the Bitcoin block chain before the event.
+
+1. If you send bitcoins as payments, note that many services may stop
+accepting bitcoins at 00:00 UTC on August 1st or earlier.
+
+1. Be wary of storing your bitcoins on an exchange or other third-party
+wallet.  If they accept transactions during the event, they could lose
+money and will likely spread those loses across all their users.  If
+there end up being two or more competing versions of Bitcoin, then they
+may refuse to give you your bitcoins on versions they don't like.
+
+## On August 1st
+
+1. Do not trust any payments you receive after August 1st until the situation
+is resolved.  No matter how many confirmations the new payment says it
+has, it can disappear from your wallet at any point up until the
+situation is resolved.
+
+1. Try not to send any payments.  Although you should not technically be
+able to lose money this way, wallets may become confused and require
+recovery if you attempt this.
+
+1. Be wary of offers to allow you to invest in the outcome of the event
+by "splitting" your coins.  Some of these offers may be scams, and
+software claiming to split your coins can also steal them.
+
+## After August 1st
+
+No information available yet.  Please wait for multiple trustworthy
+sources to state that the event is resolved before returning to normal
+Bitcoin use.
+
+## Document history
+
+A [full history][] of this document is available.  The following points
+summarize major changes, with the most recent changes being listed
+first.
+
+- 2017-07-12 12:00 UTC: initial version.
+
+[full history]: https://github.com/bitcoin-dot-org/bitcoin.org/commits/master/_alerts/2017-07-12-disruptive-forks.md
+[list of updates]: #document-history
+[news sites]: /en/resources#news
+[community resources]: /en/community
--- a/_alerts/2017-07-12-potential-split.md
+++ b/_alerts/2017-07-12-potential-split.md
@@ -2,8 +2,8 @@
 ## This file is licensed under the MIT License (MIT) available on
 ## http://opensource.org/licenses/MIT.
 
-title: "Potential disruptive chain forks"
-shorturl: "disruptive-forks"
+title: "Potential disruptive chain split"
+shorturl: "potential-split"
 active: true
 banner: "Warning: Bitcoin may be unsafe to use starting July 31st (click here for details)"
 
@@ -84,7 +84,7 @@ first.
 
 - <span class="date">2017/07/12 12:00 UTC</span>: initial version.
 
-[full history]: https://github.com/bitcoin-dot-org/bitcoin.org/commits/master/_alerts/2017-07-12-disruptive-forks.md
+[full history]: https://github.com/bitcoin-dot-org/bitcoin.org/commits/master/_alerts/2017-07-12-potential-split.md
 [list of updates]: #document-history
 [news sites]: /en/resources#news
 [community resources]: /en/community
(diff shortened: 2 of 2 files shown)


_alerts/2017-07-12-disruptive-forks.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Before August 1st
+
+1. If you accept bitcoins as payments, we recommend that you stop
+accepting Bitcoin payments at least 12 hours before 00:00 UTC on August
+1st (24 to 48 hours may be safer).  This will give time for all pending
+payments to confirm on the Bitcoin block chain before the event.
+
+1. If you send bitcoins as payments, note that many services may stop
+accepting bitcoins at 00:00 UTC on August 1st or earlier.
```
luke-jr: Due to replays, this is probably insufficient.

harding: The previous item recommends that people accepting payments stop doing so 24 to 48 hours early; I think that addresses your concern.

This line warns people that the consequences of receivers stopping the acceptance of new payments is that they won't be able to pay.  We can't say for sure when that will happen, as different services may develop different policies.

_alerts/2017-07-12-disruptive-forks.md
```diff
@@ -0,0 +1,86 @@
...
+spread false news in order to exploit the situation.
+
+Remember that you alone are responsible for the safety of your bitcoins,
+and that if you lose control of them for any reason, there is nothing
+the operators or contributors to this website can do to help you.
+
+## Before August 1st
+
+1. If you accept bitcoins as payments, we recommend that you stop
+accepting Bitcoin payments at least 12 hours before 00:00 UTC on August
```
achow101: It may be beneficial to link to a site which converts the UTC time to the local times for different timezones. People may not know when 00:00 UTC is for them.

harding: Good idea!  Actually, let me first see if I can find some freely-licensed javascript that will do the conversion to localtime in the browser.

wbnns:
@harding Thanks for working on this. I'm concerned in that Bitcoin.org has received just over half a million visits in the past week alone (~73,000 visits/day) and the TLDR on this alert (unless I am reading it wrong) is to not doing anything with Bitcoin after August 1, anywhere, and to wait for further instructions from multiple trustworthy sources or you could lose your bitcoin. Multiple trustworthy sources also sounds subjective and vague.

I'm not sure this is good UX, it seems very confusing from a site visitor's perspective and provides no long term reassurance or solutions. It sounds very cataclysmic and would show on every single page as the most prominent element.

I agree there are certainly risks with the upcoming dates, however.

harding:
@wbnns 

> the TLDR on this alert (unless I am reading it wrong) is to not doing anything with Bitcoin after August 1, anywhere, and to wait for further instructions from multiple trustworthy sources or you could lose your bitcoin.

That is exactly the intended TLDR.  Do you think I should summarize it at the top?

> Multiple trustworthy sources also sounds subjective and vague.

I can clarify to say "sources you personally trust".  Will that help?  In my attempt to be neutral here, I'm trying to encourage people to pick their own news sources, as surely there will be various spins on whatever events happen.  (Though I do link to our news site listing as a start for any real newbies.)

I'm also trying to make sure we don't set up Bitcoin.org as some sort final arbitrator of what to do.  People are responsible for their own bitcoins; we're just here to help when we can.

> it seems very confusing from a site visitor's perspective

In what regard?  Your TLDR grasps the point completely.

> and provides no long term reassurance or solutions.

There are none that I'm aware of.  This is a wait-and-see scenario.  Potential economic splits are scary as fuck, and I don't think we should dilute that message.

theymos:
I consider a significant BIP148 fork to be rather unlikely, but this is a prudent warning just in case. Better safe than sorry.

Maybe the page should warn people to move their BTC to a wallet that at least allows exporting private keys, since if you can't export private keys, then the service will be deciding how your coins are handled in a split for you.

harding:
@theymos the current text includes the following point:

> Be wary of storing your bitcoins on an exchange or other third-party
wallet.  If they accept transactions during the event, they could lose
money and will likely spread those loses across all their users.  If
there end up being two or more competing versions of Bitcoin, then they
may refuse to give you your bitcoins on versions they don't like.

Do you think I should also mention private keys?  Maybe say something about making a wallet backup, so it doesn't sound so technical?

> I've preferred to use the term "split" rather than "fork"

Good idea.  I'll make that change.  Thanks!

wbnns:
@harding If we are going to have criteria for this type of alert, we should probably also have something objective that we can agree on that will enable it to be removed.

A bright orange or red alert indefinitely on Bitcoin.org at the top of all pages is not a good first time experience. It is going to introduce users to many warnings and terms that they simply aren't going to be familiar with.

For example, one of our top pages is the Getting Started page (there have been 450k+ visits to that page in the last month). No doubt people are going to click on the warning before they read how to get started. The warning assumes they already have and will understand what you're suggesting they do.

_alerts/2017-07-12-disruptive-forks.md
```diff
@@ -0,0 +1,86 @@
+---
+## This file is licensed under the MIT License (MIT) available on
+## http://opensource.org/licenses/MIT.
+
+title: "Potential disruptive chain forks"
```
wbnns: As @theymos mentioned, let's change instances of `fork` to `split`.

theymos:
@harding Yeah, recommending making a wallet backup (or switching to a wallet that supports it) might be a good way to go about it. I think that a lot of people don't even realize that they're using a bank rather than a real wallet.

_alerts/2017-07-12-disruptive-forks.md
```diff
@@ -0,0 +1,86 @@
...
+shorturl: "disruptive-forks"
+active: true
+banner: "Warning: Bitcoin may be unsafe to use on August 1st (click here for details)"
```
wbnns: I think it would be better to say something like:

> Warning: A potential chain split may occur on August 1st.

Also, users don't need to "click here for details" if it's an active link.

_alerts/2017-07-12-disruptive-forks.md
```diff
@@ -0,0 +1,86 @@
...
+able to lose money this way, wallets may become confused and require
+recovery if you attempt this.
+
+1. Be wary of offers to allow you to invest in the outcome of the event
+by "splitting" your coins.  Some of these offers may be scams, and
+software claiming to split your coins can also steal them.
+
+## After August 1st
+
+No information available yet.  Please wait for multiple trustworthy
```
wbnns: Let's revise this to say that we'll update this section with more information after August 1st.

harding:
@wbnns Criteria for removal is a good idea.  It's difficult to commit in advance to hard criteria, but I suggest that we keep the warning up until all but one side of the chain split appears to be abandoned.  Alternatively, if more than a week passes and both chains still are being extended at a reasonable rate, we should consider providing instructions for splitting coins and ensuring normal transacting occurs on the user's preferred chain (or link to someone else's instructions for this); in that case, we'll probably have to keep the warning up for at least a month more so that we can try to stop users from making transactions without reply protection.

Regarding bad UI, I think the second sentence of the alert boils it down in plain English:

> This means that any bitcoins you receive
after August 1st may disappear from your wallet at a later time or be a
type of bitcoin that other people will not accept as payment.

I think the rest of the post is pretty non-technical too, since that was a primary goal, although I'm happy to receive any suggested phrasing changes you have.  I can also write a section or a separate document with a short explanation for non-Bitcoiners if you think that would be useful.  (I don't think it would be myself; I think the current document should make anyone hesitant about using Bitcoin at the moment, which is the correct response, even if they don't know all the terms.)

_alerts/2017-07-12-disruptive-forks.md
harding: I think the current title is better because it describes the consequence in plain English rather than the consequence in technical jargon.

For the parenthetical, we've done that with previous alerts because it isn't clear that the banner can be clicked.  E.g.:

```text
user@devbco:~/alerts/_alerts$ git grep 'click here'
2015-07-04-spv-mining.md:#banner: "WARNING: many wallets currently vulnerable to double-spending of confirmed transactions (click here to read)"
2015-10-12-upnp-vulnerability.md:#banner: "WARNING: serious vulnerability in UPnP library used by Bitcoin Core (click here to read)"
2016-11-01-alert-retirement.md:## banner: "Alert system is being retired (click here to read)"
2017-07-12-disruptive-forks.md:banner: "Warning: Bitcoin may be unsafe to use on August 1st (click here for details)"
```

I continue to think that it's a good idea.

wbnns:
@harding 

> I think the current document should make anyone hesitant about using Bitcoin at the moment, which is the correct response, even if they don't know all the terms.)

I do not recommend this. Bitcoin.org is one of the most highly trafficked bitcoin websites with millions of visits a month. Many people in the media and users alike around the globe often cite it each week - in TV, in print, on web pages around the world.

If this alert is too stark (in it's current version, as you mentioned, the TLDR you've intended to communicate is basically, don't use Bitcoin on or after August 1st), it risks spreading a lot of FUD and panic selling before August 1st that could wipe billions off Bitcoin's market cap and who knows how much out of people's pockets.

harding:
@wbnns 

> If this alert is too stark (in it's current version, as you mentioned, the TLDR you've intended to communicate is basically, don't use Bitcoin on or after August 1st), it risks spreading a lot of FUD and panic selling before August 1st that could wipe billions off Bitcoin's market cap and who knows how much out of people's pockets.

I refuse to evaluate my writing based on how it might affect Bitcoin's market cap.  I resolve only to tell the truth as best as I am able, and in this document I suggest to users exactly the concerns I hold for myself and exactly the same steps I plan to take myself---specifically that potential chain splits are scary and that I will not accept bitcoins from August 1st until I believe the situation is resolved.

wbnns:
@harding I understand, but disagree - there is a difference between what you are terming as a "correct response" (what you wrote) and speculation. You're speculating on a personal belief that Bitcoin shouldn't be used for an indefinite amount of time on or after August 1st and manifesting that as an official Bitcoin.org alert to millions of people:

image

I do agree there are risks associated with a chain split that we can inform users about, but you'll need to compromise and work on the wording so it doesn't read with so much FUD.

_alerts/2017-07-12-disruptive-forks.md
wbnns: Let's also change the title from:
`2017-07-12-disruptive-forks.md`

To:
`2017-07-12-chain-split.md`

Cobra-Bitcoin:
Looks good to me. We will just remove the alert if nothing happens, so I think it's good to be safe and warn users based on the worst case scenario. Many people don't even know there will be a split.

Let's try to merge this today, we can make small changes later, but I think an alert like this is long overdue. It's our duty to inform users and not sugar coat the harsh reality.

This situation sucks, but we have the the triumvirate of evil (Roger, Jihan, Craig) to blame for the current mess we found ourselves in...

wbnns:
@Cobra-Bitcoin We should update the wording before, though. As it stands now, it says not to use Bitcoin on or after August 1st with no clear definition of when it can ever be used again.

harding:
I'm just fixing some of the issues identified above now and will push the commits within the next 30 minutes.

harding:
Pushed some new commits and setup a [live preview](http://dg0.dtrt.org/en/alert/2017-07-12-potential-split) particularly for anyone who wants to help test the Javascript-based localized dates.

@achow101 I addressed the localtime issue using some Javascript.  If JS is enabled, the dates should be displayed in localtime (works here).  If JS is disabled or if the code is unable to parse the date, the original string is left unchanged (the base string is in this required format: `2017/08/01 00:00 UTC`).

I've only tested this on Firefox so far.

This change required some rewriting to not mention August 1st specifically in the text and also to deal with the full datetime string.  I noticed when testing this change that the fork starts for me on July 31st, which means it starts then for other people too, so I changed the banner message from "August 1st" to "July 31st".

@theymos I changed the filename and the text to use "split" instead of "fork".  I also mentioned wallet backups as a criteria for deciding whether or not you're using a third-party service.

@wbnns In the "After" section, I mentioned that we would updated; thank you for suggesting that clarification.

If you would like to suggest any other phrasing changes, even after this is merged, I'd be happy to consider them.

Thank you everyone for review.  Please feel free to suggest further changes and I'll either update before merge or open a new PR after merge.

harding:
@wbnns 

> As it stands now, it says not to use Bitcoin on or after August 1st with no clear definition of when it can ever be used again.

I think the definition is clear---you even cited it in your TLDR: for users, the situation is resolved when multiple news sources they trust say that the situation is resolved.  That's the best I think we can give non-experts.  Actually, it's probably the best we can give experts too, as this will be an interactive event with multiple people trying to affect the outcome---some of them in ways we may not be able to predict right now.

theymos:
LGTM.

_alerts/2017-07-12-potential-split.md
```diff
@@ -0,0 +1,105 @@
...
+{% assign start='<span class="date">2017/08/01 00:00 UTC</span>' %}
+
+*Last updated: <span class="date">2017/07/12 12:00 UTC</span>.  This
+page will be updated when new information becomes available.  See the
+[list of updates][].*
+
+At {{start}}, Bitcoin confirmation scores may become unreliable for
+an unknown length of time.  This means that any bitcoins you receive
+after that time may later disappear from your wallet or be a
+type of bitcoin that other people will not accept as payment.
```
wbnns: We should also add a sentence to the effect of:

`Leading up to, during and after this period, bitcoin may also experience significant price fluctuations in relation to other currencies.`

achow101:
ACK

_Sent from my Google Pixel using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

_alerts/2017-07-12-potential-split.md
```diff
@@ -0,0 +1,105 @@
...
+recovery if you attempt this.
+
+1. Be wary of offers to allow you to invest in the outcome of the event
+by "splitting" your coins.  Some of these offers may be scams, and
+software claiming to split your coins can also steal them.
+
+## After the event
+
+We will update this section with more information after {{start}}.
+Please wait until multiple news sources that you trust have stated that
```
wbnns: Let's change this to:

`Please wait until multiple news sources that you trust have stated that the event is resolved or this alert has been updated to indicate as such before returning to normal Bitcoin use.`

wbnns:
@harding 

> you even cited it in your TLDR

Apologies if I wasn't clear - I was referring in disagreement to what you said. We should just focus on what it is - a potential disruptive chain split - and what users can do about it (as described in the content of the alert), rather than inferring with the alert headline that Bitcoin shouldn't be used after July 31 for an indefinite period of time.

harding:
> rather than inferring with the alert headline that Bitcoin shouldn't be used after July 31 for an indefinite period of time.

Oh, sorry, I replied to that on your review comments so it got threaded rather than put into the main thread: "I think the current title is better because it describes the consequence in plain English rather than the cause in technical jargon."

Would adding "temporarily" in there be acceptable to you? 

> Warning: Bitcoin may be **temporarily** unsafe to use starting July 31st

 (emphasis just for illustration here; won't be in the headline)

_alerts/2017-07-12-potential-split.md
```diff
@@ -0,0 +1,105 @@
...
+## 7/12 to 7/28: "warning" (orange); 7/29 until resolution: "alert" (red)
+bannerclass: "warning"
+---
+{% assign start='<span class="date">2017/08/01 00:00 UTC</span>' %}
+
+*Last updated: <span class="date">2017/07/12 12:00 UTC</span>.  This
+page will be updated when new information becomes available.  See the
+[list of updates][].*
+
+At {{start}}, Bitcoin confirmation scores may become unreliable for
```
wbnns: We should also add a few sentences at the very beginning about what a chain split is and what bitcoin confirmation scores are. This will be really confusing to people who don't know.

wbnns:
@harding Regarding the title, if we focus on what it is and [provide a few sentences at the beginning of the alert](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1674#pullrequestreview-49631829) explaining, I think it will be more helpful to users:

> Warning: Potential disruptive chain split on July 31st, 2017

achow101:
If this alert is to be about potential chain forks happening soon in general, then we need to include information about BIP 91 (implemented in segwit2x). BIP 91 requires 269 of the last 269 blocks. Since BIP 91 has already started, there could be many mini-forks that happen between now and August 1st.

harding:
@wbnns I dropped the term "split" from the title, calling it just a network disruption, and added a link to our section about confirmation scores.

I think putting technobabble in the banner and the beginning of the article would just confuse non-technical people into thinking those things are important, when they aren't.  What's important is that Bitcoin will be unsafe to use starting 8/1 00:00 UTC until the outcome is clear.

harding:
@achow101 I was thinking that we didn't need an alert for that, given BIP91 requires about 80% of hashrate to signal onchain, more than 90% of hashrate has verbally agreed, and the only change any miner needs to make in the short term once BIP91 activates is setting bit 1 in their block headers---a change most of them can make without upgrading their underlying node.

Given that, I expected it wouldn't be much more turbulent than an old IsSuperMajority soft fork, with maximum fork lengths of fewer than 6 blocks with >99% probability.

And of course, that all assumes that BIP91 reaches 80% before August 1st.  Currently, no one is signaling it.[1]

Were you thinking differently?

[1] 

```text
$ height=$( bitcoin-cli getblockchaininfo | jq .blocks ) ; for i in $( seq $((height-100)) $height ) ; do bitcoin-cli getblock $( bitcoin-cli getblockhash $i ) | jq .versionHex ; done | sort | uniq -c
     51 "20000000"
     50 "20000002"
```

achow101:
@harding I was thinking that even though 80+% of the hash rate has verbally agreed to support BIP 91, given the low window and threshold, BIP 91 could activate with less than 80% hash rate support just due to luck and variation. My concern is that if BIP 91 did activate, even with 80% hash rate support, there is still 20% that are not supporting it, and that could cause problems. With the prevalence (still) of spy mining (aka SPV mining), we could have ourselves a repeat of the July 2015 fork which happened when less than 5% of the hash rate was not supporting BIP 66.

_alerts/2017-07-12-potential-split.md
```diff
@@ -0,0 +1,106 @@
+---
+## This file is licensed under the MIT License (MIT) available on
+## http://opensource.org/licenses/MIT.
+
+title: "Potential network disruption"
+shorturl: "potential-split"
+active: true
+banner: "Warning: Bitcoin may be unsafe to use starting July 31st (click here for details)"
```
wbnns: Let's update the banner to match the title (I left a couple of suggestions in previous comments) and remove the "click here" as mentioned in the previous review.

harding:
This post will require substantial rewriting to address @achow101's concerns.  I will work on that tomorrow and address some of @wbnns's points then.

achow101:
@harding Maybe you can just add a paragraph that mentions that BIP 91 could be a concern, but since no one is actually mining it now, it isn't all that important. We can update later once segwit2x actually releases software with all of their changes.

harding:
@achow101 

How about this at the end of the opening section:

> *Note:* there is a chance a milder level of disruption could start 
between now and {{start}}. If that is the case, this post will be
updated with details.

achow101:
@harding 

That looks good. Perhaps mention BIP 91 or segwit2x in that?

_alerts/2017-07-12-potential-split.md
harding: No.  It's important that users consult with multiple sources and not simply trust us.

harding:
@achow101 I tried to keep the post non-technical by not mentioning BIP148 or other details that are already known by the people who can understand them.  I think BIP91/segwit2x fits in with that.

What do you think?  I admit, I think maybe we could add a "see also" section at the bottom that links to some non-technical and medium-technical news stories, like the story stickied in /r/Bitcoin right now.

achow101:
@harding Ah, I see. Since none of the BIPs are actually mentioned, then that text is fine then.

wbnns:
@harding Thanks for the revisions. They remaining ones are:

1. Update/revise the banner text to more closely match the alert title (I mentioned some suggestions) and remove the "click here" (provided several linked citations why we shouldn't do that, also many of our users can't click since they a on mobile/tablets)
2. Add [note](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1674#discussion_r127071348) about volatility.
3. Revise wording in the `After the event` section (left [comment](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1674#discussion_r127074006) with suggested change)

luke-jr:
The reason the part about sending is insufficient, is that if there is a split, each chain will likely vary in value, and users probably only want to send coins on one chain and not the other, if they send at all.

harding:
@luke-jr excellent point, thank!  How about this as a rephrasing:

> Try not to send any payments.  During the event there may be two or
more different types of bitcoin and you may send all of the different 
types to a recipient who only expects one type.  This would benefit the
recipient at your expense.

harding:
Pushed three commits.

@wbnns I mentioned the volatility risk.  I also made the title change, even though I object to both parts of that request.  I did not make the change to the part about consulting multiple sources per [my previous reply to you](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1674#discussion_r127099030); you may suggest alternative phrasing, but I will in no case revise the document to suggest Bitcoin.org as a single authoritative source for whether or not the event is over.  I had only a spectators' view of the Ethereum hard fork, but there were a lot of people in that case blindly trusting the Ethereum Foundation and a handful of other insider sources when they said ETH-HF had won.  I think we should try to make Bitcoin better than that by encouraging people to check multiple sources.

I also added the send warning based on @luke-jr's useful criticism.

I will not be making any further edits for at least 12 hours.

achow101:
LGTM

wbnns:
@harding Thanks for the updates and also for spending time to help users be aware with the alert.

> I did not make the change to the part about consulting multiple sources per my previous reply to you; you may suggest alternative phrasing, but I will in no case revise the document to suggest Bitcoin.org as a single authoritative source for whether or not the event is over.

Regarding the above, no one said Bitcoin.org was a single authoritative source so don't worry about that, but if an alert is getting posted that millions of the site's own visitors are going to see (2,000,000+ people visited the site last month), then we also have a responsibility to let them know how to determine if it's over beyond telling them to go read the news.

@Cobra-Bitcoin mentioned getting this up today and we've done some revisions so I'm going to merge this.

I will submit a PR for the above change about referencing the alert for future updates and anyone else please feel free to open an issue or submit a PRs with additional changes to the wording, if they'd like to do so.

PR #1677, opened 2017-07-13
potential-split: Add note about monitoring alert

wbnns:
This adds one more small note to the end of the alert about continuing
to monitor the post for updates, in addition to news sources.

Unless others object, this will be merged on Friday, July 14th.

### DIFF
--- a/_alerts/2017-07-12-potential-split.md
+++ b/_alerts/2017-07-12-potential-split.md
@@ -81,9 +81,10 @@ software claiming to split your coins can also steal them.
 
 ## After the event
 
-We will update this section with more information after {{start}}.
-Please wait until multiple news sources that you trust have stated that
-the event is resolved before returning to normal Bitcoin use.
+We will update this section with more information after {{start}}. Please
+monitor this page accordingly and wait until multiple news sources that you
+trust have stated that the event is resolved before returning to normal Bitcoin
+use.
 
 ## Document history
 


jli225:
I agree with what @Phroneo said,

> It is FUD because they don't mention the Segwit2x dates at all which come before August 1st. It is almost certain that Segwit2x will activate beforehand but they pretend taht the NYA doesn't exist in this case. It's very strong fearmongering language that ignores the 85% intent to signal.

And as @benjyz said,

> the statement on bitcoin-org is meaningless since it only shows approvals, but not disprovals. It creates impression of consensus, when in reality that might not be the case.

harding:
ACK 41fc49d

@jli225 

> It is FUD because they don't mention the Segwit2x dates at all which come before August 1st. It is almost certain that Segwit2x will activate beforehand but they pretend taht the NYA doesn't exist in this case. It's very strong fearmongering language that ignores the 85% intent to signal.

We allude to it in the following paragraph: "Note: there is a chance a milder level of disruption could start between now and Mon Jul 31 2017 20:00:00 GMT-0400 (EDT). If that is the case, this post will be updated with details."

You can read our discussion about it starting [here](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1674#issuecomment-314913338).

We are monitoring BIP92/segwit2x and will update the alert accordingly as we learn more about the situation.

> the statement on bitcoin-org is meaningless since it only shows approvals, but not disprovals. It creates impression of consensus, when in reality that might not be the case.

I think you're referring to a document we published in 2015, the signatures supporting the capacity increases email by Gregory Maxwell.  I don't know what that has to do with the current document, except that they both have segwit in them.

wbnns:
I'm going to go ahead and merge this in an effort to keep the page as up to date as possible.

PR #1680, opened 2017-07-13
Add explanation describing chain splits

wbnns:
Since the [alert](https://bitcoin.org/en/alert/2017-07-12-potential-split) went live last night, I've read a lot of feedback from people who are confused about what a chain split is and are asking others to ELI5 (explain like I'm 5).

This PR adds a brief, high level overview of what a chain split is to the beginning of the post. Unless there is any strong objection that anything I've said is fundamentally incorrect in describing a chain split, I will merge this today at 18:00 UTC.  Also, if anyone would like to suggest any improvements or additional copy, please feel free to submit a PR to my branch before then (`wbinns-split-updates-0713171`), or open a new PR or Issue later today.

PR #1687, opened 2017-07-13
Fix typo: later (vs. latter)

sneurlax:
https://www.google.com/search?q=former+latter


wbnns:
@sneurlax Heya! Thanks for submitting this. It has been fixed in #1680. 👍

PR #1695, opened 2017-07-19
Updates instructions of setting up Transifex

kuzzmi:
It took me a while to understand those commands.
After I successfully did, I thought it makes sense to update these instructions with comments.


wbnns:
@kuzzmi Thanks for doing this. 👍

PR #1702, opened 2017-07-23
Correct bcc configuration reference

cdelargy:
Wiki page URL contained case error (bitcoin->Bitcoin)


wbnns:
@cdelargy Thanks for fixing!

PR #1706, opened 2017-07-25
Require access to private keys for wallets 

crwatkins:
This PR codifies what I believe to be our de facto criteria for the last two years.  There has been a fair amount of [discussion](https://github.com/bitcoin-dot-org/bitcoin.org/issues/1109) et al. with very little support for listing wallets that do not allow access to private keys.  For over two years we have not listed any new wallets that do not allow access to private keys.  Instead, we have listed such submissions elsewhere on the site under other categories.

### DIFF
--- a/docs/managing-wallets.md
+++ b/docs/managing-wallets.md
@@ -30,16 +30,12 @@ in the future
 - Avoid address reuse by displaying a new receiving address for each transaction
   in the wallet UI
 - Avoid address reuse by using a new change address for each transaction
+- User has access to private keys
 - If private keys or encryption keys are stored online:
   - Refuses weak passwords (short passwords and/or common passwords) used to
     secure access to any funds, or provides an aggressive account lock-out
 feature in response to failed login attempts along with a strict account
 recovery process.
-- If user has no access over its private keys:
-  - Provides 2FA authentication feature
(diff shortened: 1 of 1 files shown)


schildbach:
Just checking, this would essentially remove Coinapult, Xapo and Coinbase from the web wallets section? All others grant you full or shared control over your money.

crwatkins:
@schildbach I left that out of this PR for now so that we could discuss the criteria and let it stand on its own, but after we all settle on any new criteria I would submit a new PR to update the listings.  If the wording remains the same, the answer to your question is [yes](https://github.com/bitcoin-dot-org/bitcoin.org/issues/1109#issuecomment-268252161) those web wallets would most likely be removed.  Note that in [this PR](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1469) we already moved these wallets, but have not yet removed them from the wallet listings.  

Note that Coinbase has a non-default wallet mode which I believe allows for access to private keys.  Even though our normal criteria are applied against default operating modes, I would be willing to consider this for an exception if Coinbase or another party would be interested in proposing and supporting the listing.

wbnns:
Unless others object, this will be merged on Wednesday, August 2nd.

PR #1713, opened 2017-07-29
Adds English language as a source in Transifex config

kuzzmi:
This PR adds English language as a source in Transifex config, what is required when working with Transifex CLI to manually upload files.


wbnns:
@kuzzmi Thanks!

PR #1715, opened 2017-07-31
Upgrades Ruby version in Gemfile and Travis config

kuzzmi:
This PR upgrades Ruby version as well as some dependencies (json library) to improve stability and minimize a risk of future upgrades. The library upgrade is necessary due to incompatibility with Ruby version > 2.0.0.

Apart from that I noticed slightly reduced build time using a newer version.

As stated in the instructions I need to notify site maintainers, in this case I assume that's @wbnns, that `rvm install 2.4.1` also must be run on the build server before doing a merge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: ruby
 rvm:
-  - "2.0.0"
+  - "2.4.1"
 sudo: false
 cache: bundler
 env:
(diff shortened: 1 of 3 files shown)


kuzzmi:
Actually on Travis, it's even more visible:

13 min 12 sec with 2.4.1
32 min 51 sec with 2.0.0

cjennyc28:
K

PR #1718, opened 2017-07-31
Remaining multibit mentions removed

Mirobit:
I am still confused about the translation process. I removed the _walletmultibithd_ string from the en.yml file. Do I need to push this on Transifex too or is this done automatically or periodically by the maintainer?


kuzzmi:
This is still a manual process and done by maintainers, otherwise you cannot force push to Transifex.

wbnns:
This will get updated on Transifex the next time the English strings are updated there.

PR #1731, opened 2017-08-05
Improves the donation handling

### LINKED ISSUE #1728: Improve the donations handling
[Bounty](https://github.com/bitcoin-dot-org/bitcoin.org#earn-bitcoin-for-contributing): **264,000 bits / ~$750 USD**

Currently the donation address is placed in the footer of the site, where very few people will see it. It would be good to also have an additional donation appeal somewhere on the top of the site (this might boost the amount of donations we receive significantly).

This donation appeal at the top should look similar to the one used by Wikipedia (http://i.imgur.com/YIlZ6zH.png), though obviously be changed and tweaked a little to better match bitcoin.org's theme. It should feature a small one sentence explanation of why we're asking for donations ("Bitcoin.org is a community funded project, donations are appreciated and used to improve the website.") and feature a green "Donate" button which when pressed will launch a modal with the donation address and the QR code. This donation appeal at the top should **only** be visible to non-mobile users.

The functionality is not that hard to implement, but the key thing about this bounty is that the design and user experience should be excellent. We need the right choice of colours and it's also important that the modal actually looks like it belongs on bitcoin.org and not something that's come out of Twitter Bootstrap.

kuzzmi:
This PR improves donations UI and closes #1728.

What was done in this PR:
- [x] Adds Donations configuration to `_config.yml`
- [x] Adds collapsible Donations bar to the top of the page
- [x] Displays Donations bar only on desktop
- [x] Adds modal window with QR code
- [x] Adds buttons for suggested donations
- [x] Adds custom donation inputs with auto-conversion USD <-> BTC based on Blockchain.info ticker
- [x] Adds field for an optional message
- [x] Auto-generates QR code based on the amount and optional message
- [x] Adds support for future translations of a Donations bar and modal window

### Screenshots

image

image

image

Any feedback is highly appreciated.

### ! Warning !

This PR modifies JavaScript and CSS files, which are cached with Nginx with no control over the changes (we've seen that after wallets pages were refactored). This might result in bugs if users get old `main.js` and `main.css`.

---

Bounty can be sent to:
14kafbQ3WBmvMhAA3y6gwohUuYQ4zLZrLq


crwatkins:
This comment may be out of the scope of this particular effort (particularly because we currently don't have any dynamic content capabilities deployed) but it would be good to eventually have a unique payment address (perhaps BIP32-based) for each user for privacy reasons.

kuzzmi:
@crwatkins It makes sense, however, the transparency is equally important in my opinion. Anyway, with current configuration it won't be difficult to tweak the front-end code to support this.

_includes/layout/base/html-head.html
```diff
@@ -16,6 +16,8 @@
 {% if page.lang == 'bg' or page.lang == 'el' or page.lang == 'ko' or page.lang == 'hi' or page.lang == 'pl' or page.lang == 'sl' or page.lang == 'ro' or page.lang == 'ru' or page.lang == 'tr' or page.lang == 'uk' or page.lang == 'zh_CN' or page.lang == 'zh_TW' %}<link rel="stylesheet" href="/css/sans.css">{% endif %}
 <script type="text/javascript" src="/js/base.js"></script>
 {% if page.id != 'download' %}<script type="text/javascript" src="/js/main.js"></script>{% endif %}
+<script src="/js/jquery/jquery-1.11.2.min.js"></script>
+<script src="/js/jquery.qrcode.min.js"></script>
```
Francisc: Why not generate QR code at build time rather than every time the modal is opened?
Would spare an extra dependency too.

_config.yml
```diff
@@ -582,6 +582,14 @@ collections:
   platforms:
     output: false
 
+donation_banner:
+  address: 1GwV7fPX97hmavc6iNrUZUogmjpLPrPFoE
+  display: true
```
Francisc: This should also be used for conditionally loading other files (e.g. in index.html).

_includes/layout/base/pagetop-donation.html
```diff
@@ -0,0 +1,71 @@
...
+This file is licensed under the MIT License (MIT) available on
+http://opensource.org/licenses/MIT.
+{% endcomment %}
+
+<div class="donation-container">
+  <div class="donation-text">
+    <p>
+      {% translate donation-banner-text layout %}
+    </p>
+    <button class="donation-btn" onclick="openDonationModal()">
```
kuzzmi: `type="button"` is needed in the forms, otherwise it will default to a submit type and submit a form. In the modal `<form>` is not needed, thus not used, therefore `type="button"` not needed either.

_includes/layout/base/pagetop-donation.html
```diff
@@ -0,0 +1,71 @@
...
+      {% translate donation-banner-donate-button layout %}
+    </button>
```
kuzzmi: Yes, it's used in a handful of pages and in this scenario is used as an entry point for loading BTC/USD rates and setting up event handlers.

_includes/layout/base/pagetop-donation.html
```diff
@@ -0,0 +1,71 @@
...
+  </div>
+  <div class="donation-visibility-toggle" onclick="toggleDonationBanner()">
+    {% translate donation-banner-toggle-button layout %}
+  </div>
```
kuzzmi: Agree, will update. Don't know why I didn't use it in a first place.

_includes/layout/base/pagetop-donation.html
```diff
@@ -0,0 +1,71 @@
...
+        class="donation-btc-address"
+        href="bitcoin:{{ site.donation_banner.address }}"
+        target="_blank">
+        {{ site.donation_banner.address }}
+      </a>
+    </div>
+    <div class="modal-body bg-white" style="overflow-y:hidden;text-align: center;">
+      <div>
+        {% for amount in site.donation_banner.amounts_in_usd %}
+        <button class="donation-amount-btn" data-amount-usd="{{ amount }}">
```
kuzzmi: See comment above regarding `type="button"`.

_includes/layout/base/html-head.html
kuzzmi: QR code is regenerated anytime a user selects a sum, and/or enters a message. This can be removed, however there will be no need for quick donation buttons, and for neither custom amount input, nor optional message input. In the end the donation window will contain only static QR code and address.

If this is what we want to show people - a boring modal window, then jQuery and QR code generation can be removed and appropriate code.

_includes/layout/base/html-head.html
```diff
@@ -16,6 +16,8 @@
 {% if page.lang == 'bg' or page.lang == 'el' or page.lang == 'ko' or page.lang == 'hi' or page.lang == 'pl' or page.lang == 'sl' or page.lang == 'ro' or page.lang == 'ru' or page.lang == 'tr' or page.lang == 'uk' or page.lang == 'zh_CN' or page.lang == 'zh_TW' %}<link rel="stylesheet" href="/css/sans.css">{% endif %}
 <script type="text/javascript" src="/js/base.js"></script>
 {% if page.id != 'download' %}<script type="text/javascript" src="/js/main.js"></script>{% endif %}
+<script src="/js/jquery/jquery-1.11.2.min.js"></script>
```
kuzzmi: It's used for QR code generation. Currently nginx uses very long-living cache and Gzip, so I don't see a big problem in the long run, as in the end it will result in additional 32kb file to be loaded.

If this is everybody's concern, I'll rework this without jQuery.

js/main.js
```diff
@@ -527,3 +527,139 @@ var xint = setInterval(function() {
...
 }, 200);
+
+function generateDonationUrl(address, amountBtc, message) {
+    var result = [
+        address
+    ];
+
+    amountBtc = parseFloat(amountBtc);
+
+    if (!isNaN(amountBtc)) {
```
kuzzmi: > Check that amountBtc is a number first.
> For example, isNaN('') returns false because it first tries to convert to number, than checks NaN.

A check for being a number is already done using parseFloat:
```
amountBtc = parseFloat('') // NaN
isNaN(amountBtc) // true
```

`> 0` check is not really necessary as this transaction doesn't make sense.  As I will make a few changes, I'll add it.

js/main.js
```diff
@@ -527,3 +527,139 @@ var xint = setInterval(function() {
...
+            return amountUsd.toFixed(2);
+        }
+
+        $('#donation-input-amount-usd').on('keyup', function() {
+            var amount = $(this).val();
+            $('#donation-input-amount-btc').val(usdToBtc(amount));
+            generateDonationQrCode();
+        });
+
+        $('#donation-input-amount-btc').on('keyup', function() {
```
kuzzmi: Makes sense, thank you!

_config.yml
kuzzmi: Agree, will update a template to use/ignore additional files.

_includes/layout/base/html-head.html
Cobra-Bitcoin: I think the interactive way with the QR code being regenerated each time is much better and more likely to produce better results. No need to change anything.

_includes/layout/base/html-head.html
Cobra-Bitcoin: jQuery is fine, and keep in mind that this should only be shown to non-mobile users, so having some extra dependencies isn't the worst thing.

Cobra-Bitcoin:
Looks really good so far, should be ready to be merged soon. Test should pass once you force JSHint to ignore the jQuery files, similar to what was done [here](https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/js/analytics.js).

kuzzmi:
Thank you. Moving a jQuery QR generator to `/js/jquery` helped with passing JSHint.

_includes/layout/base/pagetop-donation.html
Francisc: Actually, type submit is required in forms since they are the only element for which a submit action makes sense.
Outside of forms, buttons should be just buttons.
Internet Explorer (10- or 9-) will trigger the first submit button found in DOM when pressing enter in an input that is outside of a form.

js/main.js
Francisc: I missed the `parseFloat()` statement, sorry about that.
The greater than `0` checks is useful in protecting users from themselves. :)

_includes/layout/base/html-head.html
Francisc: Good to hear caching and compression are enabled.

However, it's not just download size that matters.
While the file may be cached (never a guarantee), it's parsed and executed by the JS engine every time and takes up memory afterwards.

I realize bitcoin.org isn't a web application where performance is crucial, but why load anything that isn't required? Generally, not just jQuery. Even if you have super-fast internet, last-gen processors and huge SSD drive, I'd still try to deliver the smallest payload possible.

Also, keep in mind this is version 1.11.2 which dates back to December 2014. "Old" is an understatement.

TL;DR:
- I'd use a QR generator that's has no dependencies. If not complicated, I'd do this at build time rather than each time a users opens bitcoin.org.
- I'd use CSS animations and tricks like forcing repaints or `requestAnimationFrame` / `setInterval`. It's just a div sliding down after all.
- For XHR, I'd use the lightest possible wrapper for `XMLHttpRequest` object I could find, unless bitcoin.org already has one (which isn't jQuery).
- DOM "interrogation" hasn't required jQuery since IE8, just use native methods.

However, if you guys still want to keep jQuery, at the very least update it to latest.

_includes/layout/base/html-head.html
Francisc: Fair point, I thought the QR code was static.

wbnns:
Here's a live preview for anyone who wants to check it out:
https://bitcoin.cryptopelago.com/

_includes/layout/base/html-head.html
kuzzmi: > Also, keep in mind this is version 1.11.2 which dates back to December 2014. "Old" is an understatement.

The "old" jQuery is sitting in the repository for two years. It makes absolute sense to upgrade it, however it's out of scope of this PR.

> I'd still try to deliver the smallest payload possible.

That's great! Good for you and I'm all in. Furthermore I totally agree with each point that you raised (except `requestAnimationFrame` and `setInterval` for the animations, they are just simple CSS transitions).

TL;DR:
* No doubts there are a lot of things that can be optimized and you always can open a separate PR aside/later to address those issues and do what.
* The work done here is basically just working with what is already present in the repository. 

Agree, QR generator could be used without jQuery and DOM modifications could be done with native document and node methods.

If anybody else has the same concerns about performance and minimal footprint of the website, it won't be hard to get rid of jQuery.

_includes/layout/base/pagetop-donation.html
kuzzmi: Agree, this is addressed in the additional commit.

Francisc:
Checked out the live preview and it looks nice.
Good job, @kuzzmi.

_sass/_donation-banner.scss
```diff
@@ -0,0 +1,174 @@
...
+        font-weight: bold;
+        transition: margin-top .2s ease-in;
+        height: 90px;
+        margin-top: -110px;
+        padding: 10px;
+
+        &.expanded {
+            margin-top: 0;
+            border-bottom: 4px solid #ef9e5b;
+            box-shadow: 0 5px 10px rgba(0, 0, 0, 0.22);
```
Francisc: I'd offset the shadow 2px lower, in the live preview you can (barely) see the shadow of the pulldown over the expanded banner.

Mirobit:
@kuzzmi Looks great. I have suggestions.

The description for the textarea should be changed. The message doesn't reach bitcoin.org. It is only for the user's wallet.

Do we need the collapsible bar? Why not just open the pop up directly. There could be a container with the same text on the top part of the donation pop up.

Not sure about this, but we could have a "Donated" checkbox that users can check after they donated. This creats a cookie that prevents the notice from beeing shown to the user.

wbnns:
Let's avoid pop ups. I think it's OK to have it collapsed in the current state. 

People just won't click to open/reopen it if they've already donated (or aren't interested).

kuzzmi:
@Mirobit:
> The description for the textarea should be changed. The message doesn't reach bitcoin.org. It is only for the user's wallet.

Totally agree, addressed this in a separate commit.

I also included Ukrainian and Russian translations for the bar and modal.

_sass/_donation-banner.scss
kuzzmi: Good catch! Done.

Cobra-Bitcoin:
Unless others object, this will be merged Monday 7th August.

wbnns:
@kuzzmi Thanks for all of the work on this. Here's the [transaction confirmation](https://live.blockcypher.com/btc/tx/376788ebc0026cd2ae132b1a402b3f0c85af5c6a653da521aa10f9aeb936975c/) for the bounty.

kuzzmi:
@wbnns Thank you!

PR #1732, opened 2017-08-07
update -regtest generate time

eklitzke:
When I ran the listed command, it took 100 ms on my laptop:

```
real	0m0.100s
user	0m0.001s
sys	0m0.003s
```

I don't know how my laptop compares to a "generic PC", but I don't think it's 10x faster, much less 300x faster.


Mirobit:
ACK. Takes 1 second for me on my windows pc.

wbnns:
@eklitzke @Mirobit Thanks!

PR #1755, opened 2017-08-21
IRC/Slack update

Mirobit:
- Adds Bitcoin Core Slack channel to dev communities
- Removes the freenode webchat iframe (doesn't work for me (Firefox, Chrome))
- Replaces #bitcoin-dev with bitcoin-core-dev. We link to the Core project on Github, so we should link to the specific core channel. The channel is also a lot more active and doesn't require a freenode account. I am okay with listing both, if anyone wants to keep the old one.


wbnns:
@Mirobit Thanks!

PR #1757, opened 2017-08-22
Add itBit and BitBay; Update Peru to shiny flag

Mirobit:
I don't think exchanges that [gamble](https://qz.com/1059179/huobi-and-okcoin-chinas-two-biggest-bitcoin-btc-exchanges-were-themselves-to-150-million-in-idle-client-funds/) with their customers money should be listed. That's why I removed Huobi and OkCoin.

I added itBit to us exchanges. Has been around for a while and is trusted.


wbnns:
@Mirobit Hey, thanks for adding itBIt. Regarding removing the Chinese exchanges, I'm not sure we should remove them because:
1. AFAICS there has been no loss of customer funds.
2. Customers can withdrawal to any bitcoin address after using the exchange to convert fiat.
3. We have a bold disclaimer at the top of the page instructing users to choose a wallet where they will keep their bitcoin. 
4. In the linked article, OKCoin says they do not use customer funds for their own gain and that the claim is false.

Mirobit:
Okcoin and huobi have shown questionable behavior in the past and lied about it (e.g wash trading). We will only know if funds were lost when they collapse. I am afraid of an mtgox 2.0.

But I am fine with keeping this on hold till more evidence shows up.

wbnns:
@Mirobit Ok, let's see. If you want to rebase and submit a separate PR to add itBit, that SGTM!

Mirobit:
@wbnns I will open a new PR if the OkCoin/Huobi situation develops.

I also added the largest polish exchange.

PR #1758, opened 2017-08-22
Fixed BitGive url

Mirobit:
- Fixed BitGive url
- Add Core project as press contact
- Removed unused string (domain doesn't exist any more)


wbnns:
👍

PR #1769, opened 2017-08-24
add verbosity option:2

dcousens:
Reference https://github.com/bitcoin/bitcoin/blob/master/src/rpc/blockchain.cpp#L725


wbnns:
@dcousens Thank you!

PR #1784, opened 2017-09-06
Minor grammatical improvements

hannemennah:


wbnns:
@hannemennah Hello, thank you for taking time to review these. We're going to wait on incorporating these for now.

Since the site is currently available in 25+ languages and the meanings of the modified excerpts haven't changed, the net result would be very minute changes in English, without any material change in other languages except for a bunch of work that needs to be redone by the translators to keep in sync with your suggested changes.

I hope this makes sense!

PR #1785, opened 2017-09-06
Add Quadriga CX canadian bitcoin exchange

findkiko:
Quadriga has ~80% market share in Canada, seems a dis-service to newcomers to omit it from the list.


wbnns:
@findkiko Thanks for taking care of adding this!

PR #1811, opened 2017-09-19
Bitcoin Core 0.15.0.1

laanwj:
Add Bitcoin Core 0.15.0.1.

There's only one bugfix in this release, that affects only GUI users. It was annoying enough to do a new build, though:

> After upgrade to 0.15.0, some clients would crash at startup because a custom
> fee setting was configured that no longer exists in the GUI. This is a minimal
> patch to avoid this issue from occuring.

I've kept the release notes the same as 0.15.0, updating the download link and adding mention of the fix.

### DIFF
--- a/_releases/v0.15.0.1.md
+++ b/_releases/v0.15.0.1.md
@@ -1,0 +1,925 @@
+---
+# This file is licensed under the MIT License (MIT) available on
+# http://opensource.org/licenses/MIT.
+# Text originally from Bitcoin Core project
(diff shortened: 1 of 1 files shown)


wbnns:
@laanwj Ok, thanks - I'll merge once tests pass.

wbnns:
@laanwj Hey, apologies, just noticed that it looks like commit `45bdba4` isn't signed - would you mind updating?

wbnns:
@laanwj Thanks, merging once tests pass. 👍

PR #1818, opened 2017-09-21
Remove repeated word "compute" by mistake

tenghuanhe:
It seems that the word "computed" is repeated twice by mistake.


wbnns:
@tenghuanhe Thanks! Fixed. :)

PR #1824, opened 2017-09-28
Add Segwit2x Safety Alert

Cobra-Bitcoin:
This adds a new alert warning users about storing any BTC in Coinbase, Bitpay and Xapo. These companies continue ahead with their plan to hard fork away from consensus, and so this alert is necessary to make sure users are informed enough about the incoming threat.

Thoughts? Feedback? I've tried to keep it fairly short, and I don't want this to grow into a huge complex document full of a million different things. I think the key part is to just tell people to get their coins off these dangerous services, and I think this alert accomplishes that pretty well.

I've made this into a red alert because I think this incoming hard fork is possibly the greatest threat to Bitcoin, so a red warning is necessary to catch people's attention. I want as many as possible to know what these insidious companies are planning so we can prevent them quietly pushing it through. Raising awareness is key. I'm going to be suspicious of anyone changing this document in such a way as to make the issue seem like less of a danger and trying to "tone it down" without clear well thought out reasons.


achow101:
I think instead of warning specifically about Coinbase, Bitpay, and Xapo, this should be a more generic warning for any service that has stated support segwit2x. We can name companies in the post itself. I don't think it is appropriate to have them in the title, especially if they were to change their position later.

_alerts/2017-09-28-segwit2x-safety.md
```diff
@@ -0,0 +1,51 @@
+---
+title: "Beware of Coinbase, Xapo, and Bitpay!"
+shorturl: "segwit2x-safety"
+active: true
+banner: "Beware of Coinbase, Xapo, and Bitpay!"
```
achow101: Instead of naming specific companies in the title, I think the title should be more generic warning about companies and services that support segwit2x in general.

_alerts/2017-09-28-segwit2x-safety.md
```diff
@@ -0,0 +1,51 @@
...
+date: 2017-09-28
+---
+
+## Summary
+
+A minority of Bitcoin businesses plan on switching their users to a new incompatible altcoin
+which they plan on wrongly naming Bitcoin through a controversial and contentious hard fork.
+We believe that this new altcoin has no technical legitimacy and will only exist to make it
+easier for corporations to influence the direction of the Bitcoin project. We urge all
+Bitcoin users to immediately remove all their coins from Xapo, Bitpay and Coinbase in particular before November.
```
achow101: Perhaps instead of listing specific companies here, we should be generic and say "remove all their coisn from such services before November." Then at the bottom of the post we can have a list of companies that are currently known to support segwit2x and something pointing to a statement where they say they will. This way we can remove companies if they change their position.

I think this should be done here and elsewhere "Xapo, Bitpay, and Coinbase" is mentioned in this post.

_alerts/2017-09-28-segwit2x-safety.md
```diff
@@ -0,0 +1,51 @@
...
+The best course of action is to cease using these services, and to instead migrate all your
+coins to a full node wallet such as Bitcoin Core, as this is the only way to ensure you are
+sovereign over your own money during the controversial hard fork.
+
+## After the fork
+
+As the very first Bitcoin site, and original home of the Bitcoin project launched by Satoshi Nakamoto, we will
+never support Segwit2x, and will remove any wallets or services that present Segwit2x as Bitcoin and not as an altcoin.
+Segwit2x is an attack on the Bitcoin originally envisioned by our site's founder, it essentially plans to turn Bitcoin
+into something similar to PayPal and sets a precedent that rather than the users making their own choices, choices are made
```
achow101: I disagree with comparing segwit2x to Paypal. I think we should just leave this at "companies deciding what bitcoin is rather than the users".

_alerts/2017-09-28-segwit2x-safety.md
```diff
@@ -0,0 +1,51 @@
...
+them to abuse their position and decide for you what is and isn't a Bitcoin, and we believe
+this abuse is likely to cause significant economic damage to you and the wider Bitcoin economy.
+The best course of action is to cease using these services, and to instead migrate all your
+coins to a full node wallet such as Bitcoin Core, as this is the only way to ensure you are
+sovereign over your own money during the controversial hard fork.
+
+## After the fork
+
+As the very first Bitcoin site, and original home of the Bitcoin project launched by Satoshi Nakamoto, we will
+never support Segwit2x, and will remove any wallets or services that present Segwit2x as Bitcoin and not as an altcoin.
```
achow101: If you plan on leaving this statement as is, then we need to delist all of the services that support segwit2x before or at the same time this post goes live.

Cobra-Bitcoin:
> I don't think it is appropriate to have them in the title

@achow101 If you cut off the head the body will die. The largest companies should be the ones getting the most pressure applied on them prominently. Regular users don't keep track of who is and isn't supporting Segwit2x, and if I made it general, the largest companies and ones causing the most damage will get drowned out by a list of random companies nobody has ever heard of.

_alerts/2017-09-28-segwit2x-safety.md
Cobra-Bitcoin: No need for this right now, we can remove them nearer to the fork once we know whether they plan to present Segwit2x as Bitcoin or not. There's a chance some wallets add protection to stay on the true Bitcoin chain.

achow101:
> Regular users don't keep track of who is and isn't supporting Segwit2x, 

Which is why I suggested that a list be added to the post with all of the companies who are supporting 2x.

> and if I made it general, the largest companies and ones causing the most damage will get drowned out by a list of random companies nobody has ever heard of.

You can order the list by biggest/most recognizable companies first. You can bold them and emphasize them in some way so that people who read it see and understand that those companies are big and can cause the most damage.

Having those three companies in the title and constantly repeating them throughout the post makes it feel more like an attack on those three companies specifically rather than a warning about segwit2x which includes much more than just those three companies.

Cobra-Bitcoin:
@achow101 This alert is created to focus exclusively on these three companies.

achow101:
> This alert is created to focus exclusively on these three companies

Well the PR title and alert url endpoint suggests otherwise.

if that is the point, can you still expand it a little bit to include a warning about other companies? I fear that people may go "I'm safe, I don't use bitpay, coinbase, or xapo, I use bitgo, I must be safe." but they wouldn't really be safe since bitgo and others also support 2x.

theymos:
I agree with the general idea, but the message needs work. (I'll have more time to look into it next week.)

wbnns:
As the site maintainer, I’m completely against this PR. It’s opinion, not fact, and also misleading in speculating about  what may or may not happen in the future - companies shouldn’t be singled out in this way.

Aside from the above, it’s also defamatory and could have legal ramifications.

Cobra-Bitcoin:
@wbnns What is misleading about it? How is it defamatory? These companies put out a statement that they're hard forking to a different incompatible network, we're not defaming them here, we're literally telling users what could happen if they continue to store their coins in companies that are publicly stating that they will leave the system. There are clear risks. For example, the stuff I mentioned about waiting months to access your coins (as Bitcoin Cash users are doing on Coinbase) and issues around replay attacks (also experienced on Coinbase with the Eth split) are things that have actually happened.

> companies shouldn’t be singled out in this way

Yes they should. Is this the part where I'm supposed to feel sorry for singling out these poor helpless multi-million dollar companies?  These companies are attacking my money by not strongly advocating for replay protection, risking damaging Bitcoin's brand and reputation, and are undermining the basis of decentralization and user choice that the system was built on. They are the ones that should be worried about legal ramifications from the millions of Bitcoin users who they are attacking, not us for simply warning users to stay away from centralized services that plan to exit the system.

Maybe the message needs a little bit of work, but the overall tone and structure should stay the same. I have no intention of playing nice with a bunch of suits that plan to screw up my money. Neither should any of you. When these businesses are plotting and scheming to take over our network, why are we sitting here worried about how it's mean/unfair to single them out in an alert?

If they want to leave the New York agreement, they have a little bit of time, but I haven't seen any hints of this at all. Not even them backtracking a little. They're committed to doing this, so it's fair game for us to broadcast a warning telling people that we believe these companies are dangerous to store BTC on.

ekerstein:
I'm against this as well. The strength of Bitcoin.org is that it is a neutral, non-partisan, trustworthy source of information. You're suggesting that this site wade into a divisive political issue. That's the wrong move.

Cobra-Bitcoin:
> You're suggesting that this site wade into a divisive political issue. That's the wrong move.

Unbelievable. How is it the wrong move to warn our users that some of the biggest Bitcoin services are shortly planning to present a completely different altcoin as "Bitcoin".

Bitcoin.org exists to serve Bitcoin users and the true Bitcoin network. This new altcoin that will be created by @jgarzik is not Bitcoin, and won't ever be Bitcoin, and companies that adopt it and present it as Bitcoin will be thoroughly and completely removed from our website because Bitcoin.org exists to serve Bitcoin and not what some corporate elites decide for us is Bitcoin.

@ekerstein I can't expect you to understand any of this though, considering on your Twitter you're [rushing to draw attention to this PR](http://archive.is/kYQ1e) to the exact same corporate elites I'm talking about, presumably so a wave of sockpuppets, trolls and FUD can attack this PR. How is the pay at Bank of America by the way? 😄

ekerstein:
@Cobra-Bitcoin The fact that you say things like "unbelievable" and the "true Bitcoin" and "altcoins"  and "corporate elites" just goes to show how emotionally and politically involved you are in this. That's not the right perspective to be suggesting changes from.

Yes, my first instinct was to **involve** people who would be impacted by this and may have a differing opinion. Not sneak things through without public commentary. That's the different between emotions versus reason being your guidance.

Cobra-Bitcoin:
Merge scheduled for October 9th to give @theymos and others enough time for further review.

wbnns:
@Cobra-Bitcoin No one other than yourself has said this alert should be added in its current format. You're acting autonomously, making a contentious change that no one has approved, and being the sole arbiter of determining who you want to single out/lambast, along with what should and shouldn't go in it.

You haven't yielded to any recommendations or followed the consensus of others. You're making demands and setting hard deadlines. Like someone forking Core and making a contentious change to the Bitcoin network, you're attempting to do the same thing with this PR to Bitcoin.org.

***

Bitcoin is what it is today, because so many of us have worked hard together, despite our individual differences to make something great and awesome, that can hopefully create positive change in the world and usher in a new era of fiscal independence to liberate us as sovereign individuals from the confines of broken systems. 

If you're the only one who is calling the shots, that's not Bitcoin - it's really disappointing.

Mirobit:
I agree with @wbnns. 
I know this is your site but it is also a community based site. With actions like this one, the site loses it's community feeling. I personaly don't want to spend time working on a site that just expresses the owners opinion. Bitcoin.org should not become like the awfuel Bitcoin.com.

Cobra-Bitcoin:
@wbnns @Mirobit The purpose of the week long review process is to get as many people to review the change and suggest changes so we can come to a consensus.

nvk:
I can't stand this 2X BS as much as most of you, but I think this may not be the way to go. I think this will only legitimize their fork and give them exposure.

sunnankar:
NACK.

I do agree that the 2X does pose a threat and users (primarily at custodians like Coinbase and Xapo  but not Bitpay because they do not really custody coins) could lose significant amounts of funds. There is already a post on this type of topic (https://bitcoin.org/en/alert/2017-07-12-potential-split) but I do understand why another update would be warranted to help prevent user confusion.

Unfortunately, this PR as written is a complete mess and undermines the creditability and authority of bitcoin.org and is not in keeping with the usual presentation on the site.

More specifically, I think if anything is written on the 2X topic then it should be written in an objective manner and neutral tone with credible and verifiable citations for assertions instead of a subjective manner and biased tone without foundation.

alp-bitcoin:
NACK as it stands now, but will be providing feedback for changes to make the same type of message.

chris-belcher:
I completely agree with the aim of this message. It's very important that something like this goes out, but with edits.

Calling out specific companies by name is a very good idea because its *concrete*. We know one of the ingredients for the [best way to explain ideas to people](https://en.wikipedia.org/wiki/Made_to_Stick#Overview) is to make them concrete, and that means naming specific companies rather than using abstract terminology. Users should know straight away that Coinbase etc will screw them over unless the user takes action.

All that talk about "corporate elites" makes you sound like a communist. Blockstream, Bitrefill, F2Pool, Vaultoro and [plenty of others](http://nob2x.org/) are also corporations yet they fully support bitcoin. Corporations are not inherently bad or good.

With reference to [communicating better](https://en.wikipedia.org/wiki/Made_to_Stick#Overview), the emotion-laden words make the text seem less credible. It would be far better to tell the user about concrete dangers, not lofty principles, so that means tell them how the only way to be safe is to withdraw coins from custodians and check which cryptocurrency is supported by any entity they transact with, and tell them what happens if they don't do this. (I know the text already contains these, I'm comparing the two types of arguments). It's totally possible be neutral-sounding and also bias to your own view of protecting bitcoin.

The stuff about appealing to the legacy of Satoshi Nakamoto isn't that great either. He is still a point of centralization which is probably why he disappeared. If he appeared tomorrow and said we should adopt 2X, would you listen to him? Of course not! Also again many users don't care about high lofty principles like Satoshi, they just want to protect their own money so appeal to that.

People talking about the "political neutrality of bitcoin.org" have gotten it completely wrong. The strength of bitcoin.org comes from its good and well-written content, and that requires warning users that their coins may be in danger. Bitcoin cannot be neutral against people who wish to harm it.

"This PR will only give the bad guys exposure" isn't a good argument because the bad guys already have all the exposure they need. They have convinced the CEOs of a couple of massive bitcoin companies. The ship has already sailed in terms of exposure.

Cobra-Bitcoin:
@nvk Unfortunately we have to give some exposure to this fork and the main companies behind it so that users know what's about to happen. 

@sunnankar I'll get rid of phrases like "corporate elites" and try to make it seem a bit less biased.

@alp-bitcoin Thank you! Looking forward to seeing your recommended changes.

> Calling out specific companies by name is a very good idea because its concrete.

@chris-belcher Agreed. One thing I really wanted to avoid is to make it general, because making it general weakens the power of the statement significantly. It's much more effective to point the finger at the main companies pushing this thing. I'll work to make the text seem less emotional.

achow101:
@Cobra-Bitcoin Can you please at least add somewhere in the post a list of other companies to avoid and mention that there are other companies involved in Segwit2x other than Xapo, Coinbase, and Bitpay? A warning about Segwit2x safety should include and call out the other companies involved in Segwit2x even if they aren't the big ones.

If you want to keep mentioning them throughout the text - fine. But in the title, can you make it a bit more generic? I fear that with the current title we are primarily going to get people who use Coinbase, Bitpay, and Xapo who read the article and thus fail to warn other people who are using other services that support segwit2x. With a more generic title I think that we can reach a wider audience and get more people to act against all companies that are behind segwit2x, not just these three.

sunnankar:
@Cobra-Bitcoin Making it a 'bit less biased' is a good start. But it really needs a complete rewrite in an objective form in contrast to its current subjective form. This will make it much more powerful in its statement because the positions will be much more defendable.

http://www.diffen.com/difference/Objective_vs_Subjective

Cobra-Bitcoin:
@achow101 I can change the title to "Beware of Coinbase, Xapo, Bitpay & others" and then append a list of these "others" at the bottom of the alert.

@sunnankar I'll make it as objective as I can, but some parts will remain in subjective form, because we need to warn users about what we believe _could_ happen.

nvk:
@Cobra-Bitcoin since you feel very strongly about this solution. Your version starts too vague, and users have a tendency to be lazy and not expand/read. May I suggest starting the alert title with "Beware of Bitcoin's incompatibility with..."

Cobra-Bitcoin:
I've updated the statement and completely rewritten it. Let me know what you think.

achow101:
I like this new version much better.

harding:
Untested ACK. 273df5544c5f8b5c1169e471ab7f52944ddf4af3

(At a guess, the Travis CI failure is caused by some missing element from disabling the donation banner.)

_alerts/2017-10-09-segwit2x-safety.md
```diff
@@ -0,0 +1,121 @@
...
+bannerclass: "alert"
+date: 2017-10-09
+---
+
+## Summary
+
+A subsection of the Bitcoin economy including Coinbase have signed an agreement to adopt
+and implement a contentious hard fork of Bitcoin [sometime](http://bashco.github.io/2x_Countdown/) in November. A hard fork is a
+backwards incompatible change to the Bitcoin network. This hard fork is not supported by
+the majority of the Bitcoin users and developers and is therefore a contentious hard fork.
```
alp-bitcoin: A majority would not be sufficient to avoid being a contentious hard fork.  This should be reworded to make sure even a majority cannot hard fork bitcoin unilaterally.

nvk:
Much better.

chris-belcher:
Looks good, although IMO Xapo and Bitpay should be in the title too.

Xapo does a lot of bitcoin volume https://www.walletexplorer.com/wallet/Xapo.com

sunnankar:
@Cobra-Bitcoin It is getting better but still needs significant work and refinement.

For general outline and better organization perhaps break up into these sections, (1) intro about potential hard fork, (2) issue about what Bitcoin is, (3) define the rule of how to define Bitcoin doing full-node transaction validation, (4) raise the issue about SPV wallets not doing validation and (5) explain why and how that poses a security risk, (6) provide the extremely safe and conservative advice about how to (a) perform your own validation and (b) possess your own private keys and then a conclusion of the article.

I also think it would be good to blunt some of the sharpness. Perhaps distinguish between companies that want to adopt 2X as BTC and those companies that will support both BTC and 2X.

I may have some time at the end of the week or next week to do some work on it.

_alerts/2017-10-09-segwit2x-safety.md
Cobra-Bitcoin: What about "This hard fork is not supported by an overwhelming majority..."?

Cobra-Bitcoin:
@chris-belcher I didn't want to make the title too long.

@sunnankar I don't see how it needs "significant" work. It sounds like your issue is more to do with the document organization and structure, and less about the message. One of the reasons this is merge scheduled is to avoid creating a complex and large document which is the product of a lot of people reorganizing, restructuring, adding sections, etc. The current structure and message is extremely minimal and simple.

theymos:
After thinking about it more, I don't think that this message is quite what is needed right now. I propose instead:

1. Remove NYA-signers from bitcoin.org now.
2. Post a blog post saying that NYA signers must clarify their position within 1 week. They need to say, "We will under no circumstances fraudulently label S2X as BTC/'Bitcoin'".
3. After the week is up, any NYA-signer that fails to clarify or clarifies in the wrong direction can be warned about in an alert, *and* the alert should give people everything they need to know about Nov 1 preparations. The first version of the alert will be the one that the most people see, so it should be as comprehensive as possible. The current PR mostly focuses on the companies, which isn't everything.

I do agree with calling out specific companies.

Cobra-Bitcoin:
@theymos Can you do a quick pull request implementing that proposal? If there's consensus to go with that approach, we can merge that in, and hold off on this statement.

I like that your proposal encourages companies to clarify their stance, however there's the chance that they just remain silent, and we don't make use of time we could have spent informing users about the incoming fork through an eye-catching alert.

What does everyone else think?

harding:
I like the idea of a blog post before a warning, although I'm not sure we need to give the companies a week.  Maybe three full business days.  Here's some quick draft text in case it's useful:

---

# Bitcoin.org to denounce Segwit2x

On FIXME:date/time (UTC), Bitcoin.org is planning to publish a banner on every page of the site warning users about the risks of using services that will default to the so-called Segwit2x (S2X) contentious hard fork.  S2X companies will be called out by name.  To ensure that we only warn users against companies that will actually put user deposits at risk, we urge all companies to publicly clarify their stance before the above date, either by a highly-visible public statement or by commenting on Bitcoin.org issue #FIXME:link (or by doing both).

If there are mitigating factors, such as your company planning to continue supporting Bitcoin by default and only supporting S2X insofar as required to allow users to sell or withdraw their S2X coins, please inform us of these details on the above issue so we can remove your entry if appropriate.

By default, we will be using the following list of companies known to support S2X in our warning:

FIXME:include_list

theymos:
@Cobra-Bitcoin, @harding: I made a PR at #1836.

I still think that this PR should be expanded with more info before going live.

kbenzle:
We need to alert community that community is under attack by 95% of the community!!!

whackashoe:
NACK. This PR as it stands is deceiving. 

> companies continue ahead with their plan to hard fork away from consensus, and so this alert is necessary to make sure users are informed enough about the incoming threat.

>> The proof-of-work also solves the problem of determining representation in majority decision
making. If the majority were based on one-IP-address-one-vote, it could be subverted by anyone
able to allocate many IPs. Proof-of-work is essentially one-CPU-one-vote. The majority
decision is represented by the longest chain, which has the greatest proof-of-work effort invested
in it. 

Consensus isn't an attack, it is how bitcoin is designed to function. Beyond the fundamental misunderstanding, the writing is self-indulgent and has a sense of viciousness to it. 

---

Collaborating with @jgarzik to write something more neutral would likely be a good course of action if the goal is to end up with something informative rather than inflammatory. Education is a great goal, and it does seem to be at least part of the motivation behind this pr- but the way it is presented seems like something a ministry of propaganda would produce.

CosmicHemorroid:
@whackashoe @jgarzik tend to dictate rather than collaborate.

Cobra-Bitcoin:
@theymos Are you working on expanding this with more info? We have to merge this on the 11th, so if you don't have time to do any work, just let me know anything you feel is missing, and I'll make the changes.

theymos:
@Cobra-Bitcoin I'm not sure if I'll have time to modify it. The changes I'd make are:

* The list of S2X services should be categorized (eg. "wallets", "exchanges", "miners", "other"), since a few of them are important, but most of them are not.
* Alternatives to Core should be given, since realistically it's just too difficult for most people to run Core. Both Ledger and Trezor have been pretty good about handling this stuff. For software wallets, Electrum and GreenAddress are both imperfect in some ways, though they might be the best options. (Electrum has the capability of choosing between forks, though I'm not sure what it'll do by default.) People should maybe be explicitly warned against using blockchain.info.
* General advice similar to the Aug 1 alert and [my reddit guide](https://www.reddit.com/r/Bitcoin/wiki/faq_handling_splits) should be given, especially including a warning against anyone transacting near the fork time. Even users of Bitcoin Core should not really be advised to transact around fork time, since this might cause them to throw away their B2X that they might've wanted to sell.
* I'd emphasize that due to the lack of proper replay protection and the possible miner support, this is much different than the Bcash split.
* Coinbase has softened their stance a bit, so I'm not sure that they should be called out in the *title*. Maybe the title should instead be something like "Beware of Bitcoin's possible incompatibility with some major services".

Also, is it OK that the Travis CI build is failing? Has someone checked that the final build actually works?

Cobra-Bitcoin:
I've updated the statement. Made the following changes:

- Categorise the list of S2X services. Some services are both wallets and exchanges, and so they are listed under both categories.
- Recommend GreenAddress and Electrum for _storing_ coins.
- Mention to not use blockchain.info.
- Change the title to "Beware of Bitcoin's possible incompatibility with some major services".
- Rework the warning around SPV wallets to instead warn against "mobile" wallets. Most average users don't understand what is and isn't an SPV wallet. However it's easy for the average user to know if they're using a mobile wallet or not.

The document is getting quite long. We have to merge this tomorrow by noon (UTC). Last minute review is really appreciated.

chalbersma:
Given the size, number and economic clout of the companies on this list; is it wise (assuming Core opposes Segwit2x) to post a banner leading to a post showing that the economic majority supports Segwit2x?

Mirobit:
[Yours](https://www.yours.org/about/) uses Bitcoin Cash, so I would not include them in the list.
[Bitoasis](https://www.reddit.com/r/Bitcoin/comments/75au0p/nya_signer_uae_company_bitoasis_wont_support_s2x/) will not support segwit2x trading. Seems like they will stick with Bitcoin.

Cobra-Bitcoin:
@chalbersma It's not an economic majority.

@Mirobit I'll remove Yours and Bitoasis. Thanks!

If you know any CEO's in any of the companies listed, please urge them to quickly make a public statement about their stance on how they'll handle 2X.

_alerts/2017-10-09-segwit2x-safety.md
```diff
@@ -0,0 +1,134 @@
...
+## Wallets
+
++ Abra (United States)
++ Bitcoin.com (St. Kitts & Nevis)
++ **BitPay (United States)**
++ BitPesa (Kenya)
++ BitOasis (United Arab Emirates)
++ **Blockchain.info (UK)**
++ **BTC.com (China)**
++ Circle (United States)
```
achow101: I don't think circle should be listed considering that they are only supporting Bitcoin as a grandfathered-in thing. See https://support.circle.com/hc/en-us/articles/217972003-New-Ventures-of-Old-Bitcoin

_alerts/2017-10-09-segwit2x-safety.md
```diff
@@ -0,0 +1,134 @@
...
+However users should only use these wallets to store their coins and never perform transactions until well after the hard fork. Any transactions you
+make on the Bitcoin blockchain can also be replicated and "replayed" on the altcoin chain. If the coins on the contentious
+hard fork have any value, there will be methods you can use to "split" your coins and have access to them. Pay extra attention
+to major Bitcoin communication channels and media shortly after the fork so you stay informed.
+
+## Mobile wallets
+
+Miners in Bitcoin are responsible for transaction ordering through putting transactions into blocks, and collectively
+creating a chain of these blocks. Most mobile wallets depend on blindly trusting the longest chain produced by the miners in order to safely send and receive transactions.
+There is considerable mining power supporting the contentious hard fork. A supermajority of miners have pledged to support the contentious
```
achow101: I don't think it can be considered a supermajority of miners anymore. It's just a majority now.

_alerts/2017-10-09-segwit2x-safety.md
```diff
@@ -0,0 +1,134 @@
...
+hard fork, therefore the longest chain as seen by most mobile wallets may not be the true block chain. As a
+result of this, spending money from a mobile wallet during a hard fork is dangerous. You will be
+vulnerable to many attacks as your wallet is not aware that the miners are breaking the consensus rules on the longest chain.
+
+If the contentious hard fork has failed and the supermajority of miners have broken their pledge
```
achow101: Same here about supermajority.

_alerts/2017-10-09-segwit2x-safety.md
```diff
@@ -0,0 +1,134 @@
...
+
+The following companies and services have pledged to adopt the contentious hard fork:
+
+## Wallets
+
++ Abra (United States)
++ Bitcoin.com (St. Kitts & Nevis)
++ **BitPay (United States)**
++ BitPesa (Kenya)
++ BitOasis (United Arab Emirates)
```
achow101: BitOasis has possibly withdrawn from the NYA: https://www.reddit.com/r/Bitcoin/comments/75au0p/nya_signer_uae_company_bitoasis_wont_support_s2x/. Perhaps followup with them may be needed to confirm this.

_alerts/2017-10-09-segwit2x-safety.md
```diff
@@ -0,0 +1,134 @@
...
++ BTCC (China)
++ BTER.com (China)
++ **Coinbase (United States)**
++ Coins.ph (Phillipines)
++ CryptoFacilities (UK)
++ Korbit (South Korea)
++ Safello (Sweden)
++ SFOX (United States)
++ **ShapeShift (Switzerland)**
++ SurBTC (Chile)
```
achow101: SurBTC has withdrawn: https://blog.surbtc.com/our-stance-on-the-segwit2x-hard-fork-9fd04323667b

theymos:
Looks good to me.

SurBTC just made a [statement](https://blog.surbtc.com/our-stance-on-the-segwit2x-hard-fork-9fd04323667b) which seems sufficient to me, so they should probably be removed.

_alerts/2017-10-09-segwit2x-safety.md
```diff
@@ -0,0 +1,134 @@
...
+
++ ANX (Hong Kong)
++ Bitex (Argentina)
++ bitFlyer (Japan)
++ Bitso (Mexico)
++ BTCC (China)
++ BTER.com (China)
++ **Coinbase (United States)**
++ Coins.ph (Phillipines)
++ CryptoFacilities (UK)
```
achow101: CryptoFacilities has possibly withdrawn: https://www.reddit.com/r/Bitcoin/comments/740son/nya_signer_uk_company_crypto_facilities_pulls/. Followup to confirm this may be needed.

chalbersma:
@Cobra-Bitcoin It does appear to be so based on this list. Possibly listing companies that support Core's vision for Bitcoin would make it more clear that it's not the economic majority that supports Segwit2x.

Cobra-Bitcoin:
Merged! Thanks to everyone who reviewed or suggested changes.

achow101:
Post merge ACK

mjamin:
@wbnns what's your position on the final version of this alert?

lichtamberg:
Post merge ACK

evilrobotted:
I am baffled by your complete lack of ethics @Cobra-Bitcoin @theymos. You are literally the worst thing to happen to Bitcoin since Mt. Gox.

Feri22:
ACK

Pheromon:
Let's rewrite it as "we don't want all those companies having a say in Bitcoin, but just the one paying us: Blockstream".

That way the message is much clearer and everybody agrees.

OCHtoe:
This conversation reads like a JV political strategy meet-up on how to strong-arm and force people against their will.  The desperation is palatable.  

As you clench your fists locking down dissent your waning power is exposed.  The harder you try to hold on the more your grip on the bitcoin community weakens.   

The season of Cobra and Theymos is coming to an end.

PR #1828, opened 2017-10-03
Update rpcs/intro.md to explain use of named arguments

esotericnonsense:
Added in Bitcoin Core 0.14.0.

I was made aware of this during a discussion on IRC just now and had no idea these even existed (they're in the release nodes for 0.14.0).


wbnns:
@esotericnonsense Thanks!

Unless others object, this will be merged on Monday, October 9th.

PR #1834, opened 2017-10-04
Add Privacy category to developer resources with ZeroLink

nopara73:
I was long wondering if such addition is welcomed.  
I probably also made some coding mistake and it will not work as I expect it to work, so for concept ACK, please don't merge blindly.


wbnns:
@nopara73 Hey, a privacy category sounds helpful, however, we would need a couple more resources to go along with it so that there is more than one entry in the category. Also, there's one other problem related to this, which is that we currently have six categories. If we add a seventh, it will break the symmetrical layout of the resources page (since seven doesn't divide equally).

Maybe the [Protect your privacy](https://bitcoin.org/en/protect-your-privacy) page could potentially be a location for some additional content and a link, related to this?

nopara73:
> Maybe the Protect your privacy page could potentially be a location for some additional content and a link, related to this?  

That's definitely the place. I haven't actually seen that before.

PR #1838, opened 2017-10-06
Update 2017-10-05-denounce-segwit2x.md

PiscisSwimeatus:
The list of companies opposing 2X is shorter and hence easier to maintain.


JaredR26:
ACK

PR #1848, opened 2017-10-11
Recommend TREZOR

Mirobit:
A lot of people are wondering if their coins are safe on a trezor wallet. Pavol and Marek handled the Bcash fork very well and have been strong opponents of Segwit2X. So I think it is safe to say that Trezor can be trusted regarding the fork.


Cobra-Bitcoin:
"htts://trezor.io" needs fixing.

PR #1864, opened 2017-10-17
Add CoinMall on 'Spending Bitcoin' page

beyourseff:
This Pull Request adds CoinMall to the Spending Bitcoin page found here: https://bitcoin.org/en/spend-bitcoin

CoinMall is a marketplace on which users can buy and sell digital goods for Bitcoin. Our company fundamentals align with those of the rational Bitcoin community. We do not and will not support any coins created from forks such as Bitcoin Cash or Segwit 2X.

Feedback is more than welcome.

### DIFF
--- a/_translations/en.yml
+++ b/_translations/en.yml
@@ -767,7 +767,7 @@ en:
     pagetitle: "Spending Bitcoin"
     summary: "There are thousands of businesses, across the globe, that accept Bitcoin."
     products-online: "Find products for sale online"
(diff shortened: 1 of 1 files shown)


Cobra-Bitcoin:
You need to fix the link to CoinMall, it links to Spendabit instead.

beyourseff:
Whoops, fixed now 👍

wbnns:
@beyourseff Congrats on your first contribution and thanks for submitting this. [CoinMall](https://www.coinmall.io/) seems like a cool idea for a website (the design also looks nice). It would be great if people had more places like this to be able to buy things with their bitcoin.

The Spending Bitcoin page gets a significant amount of traffic (just under 70,000 visits this month alone), and there aren't that many products on the site just yet. At the moment, there's less than 50 products and several of them are free. We should wait until there's a bit more inventory before we send so many people in this direction. 

Happy to reconsider if others disagree. We could also revisit this once more products are added, so that visitors from Bitcoin.org would have more options.

PR #1876, opened 2017-10-23
Reinstate coinbase

theymos:
Coinbase released quite a [good statement](https://support.coinbase.com/customer/portal/articles/2892985-segwit-2x-faq) on B2X which may warrant their complete reinstatement. That's what this PR does.

There is a sticking point in that they say, "If the Segwit2x change is accepted by most users, we may choose to rename these blockchains at a later date." That may contradict the requirements in the blog post; however, I tend to lean toward this being sufficiently forward-looking and constrained that it's OK. Discussion is required. If we decide not to completely reinstate them, then they should at the very least get a special mention in the alert as having a much better policy than eg. Xapo.


Mirobit:
ACK.

Cobra-Bitcoin:
ACK.

wbnns:
ACK.

achow101:
Coinbase released a [clarification](https://blog.coinbase.com/clarification-on-the-upcoming-segwit2x-fork-d3c0f545c3e0) on their post where they say

> We are going to call the chain with the most accumulated difficulty Bitcoin.

I'm not sure if we should reinstate them entirely. Maybe still mention them in the post with a caveat.

theymos:
@achow101 Very disappointing. Maybe it's slightly better than eg. Xapo, but since that statement completely contradicts their earlier statement, it's impossible to really say what they're going to do.

This PR shouldn't be merged, and IMO the alert is OK as-is.

PR #1877, opened 2017-10-24
Add Coinwale

punit2502:


Mirobit:
NACK:
- Only crypto to crypto
- What is the difference to nexchange.io? Nexchange.io seems to be the original.
- No feedback
- Anonymous owners

PR #1890, opened 2017-10-30
Update fr.yml

maseox:
Some fixed/new translations.


wbnns:
@maseox Hello, thanks! Is this from Transifex? If not, could you please first make the updates there, upstream? This will ensure that these aren't overwritten.

Also, it looks like you have some tests, failing. Please let me know if you need help fixing.

maseox:
I must say I don't get what's the problem, but it's for sure because I have been editing the file without using Transifex. I'll do it one more time this way and request a new pull.

Sorry for the inconvenience !

wbnns:
@maseox No problem and no apologies necessary! Thanks for trying to help with this. Here's [documentation](https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/docs/assisting-with-translations.md) on what you need to do in order to correctly update the French translation.

If you have any questions, please let me know. We can also connect directly outside of the repository, via email or chat, if you'd like, and I can walk you through the process.

PR #1891, opened 2017-10-30
Fix incorrect `GetTxOut` description

nopara73:
See details: https://bitcoin.stackexchange.com/questions/48504/how-to-check-if-an-output-has-been-spent/48505?noredirect=1#comment70553_48505
>  It gives information about unspent outputs, and won't return anything for spent outputs.


wbnns:
@nopara73 Thanks!

PR #1903, opened 2017-11-07
Add Karsha Exchange

khashmeshab:
Automatic exchange. Multinational, but I personally live in the UAE. Strongly No2X (listed here: http://nob2x.org/). Thank you.


khashmeshab:
Would you please just add us to the list? We want to get out of some competitor altcoin's list (guess who!), but we currently don't have a reliable source of traffic. Thank you.
P.S. you reviewed us once before, the problem was only with our position in the list, which is now resolved.
P.S.2 I think you must rename the "International" section to "Multinational", to make it less misleading.

wbnns:
@khashmeshab Thanks for your patience and also for the feedback, you've just been added.

Issue #1906, opened 2017-11-09
The Bitcoin Warning should be rewritten.


JeffCSlentz:
Source File: _alerts/2017-10-09-segwit2x-safety.md

The warning should be modified immediately now that Segwit2x will no longer be happening. It should instead speak in general about contentious hard forks.

wbnns:
Hello, thanks for mentioning. There's currently a pending pull request to drop the alert (#1910).

PR #1907, opened 2017-11-09
Update 2017-10-09-segwit2x-safety.md

JeffCSlentz:
Changed most present tense to past tense in this document. The title/banner text could use some work I think.

### DIFF
--- a/_alerts/2017-10-09-segwit2x-safety.md
+++ b/_alerts/2017-10-09-segwit2x-safety.md
@@ -1,73 +1,66 @@
 ---
-title: "Beware of Bitcoin's possible incompatibility with some major services"
+title: "Hard forks: Segwit2x and beyond"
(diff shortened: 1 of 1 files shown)


JeffCSlentz:
#1906

wbnns:
@JeffCSlentz Hello, thanks for spending the time to change the conjugation on this. #1910 drops the alert. Forks can be threatening and catastrophic without consensus, so the alerts themselves should be saved for when they appear to be imminently bearing down upon us.

Leaving it up, and changing all of the wording to the past tense, would be similar to leaving a Typhoon or Hurricane Warning up on everyone's televisions, well after a storm is gone. People will become desensitized and stop paying attention to it. It will then have less effect when it's needed most, the next time around.

JeffCSlentz:
Perfect, works for me. Thanks Will

PR #1908, opened 2017-11-09
Update ref_block_chain.md

ender2101:
High time to update that part.


wbnns:
@ender2101 Good catch, thanks!

PR #1909, opened 2017-11-09
Update sv translations

buren:
I found some time to add all the missing Swedish translations.

I downloaded the YAML-file from Transifex.

ℹ️ The GitHub URLs have been changed from `github.com/bitcoin-dot-org/` => `github.com/bitcoin`, the links still work (GitHub redirects..). But it seems like using `bitcoin-dot-org` would be "more" correct, since it doesn't have to redirect.

Let me know if there is something I should fix :)


kuzzmi:
Can not check Transifex project at the moment, but please make sure that these changes are reflected there as well. Otherwise the chances are changes you made might be overwritten when syncing GitHub and Transifex.

wbnns:
@buren Hello, thanks for the update! Could you please confirm that the changes were first made on Transifex, and reopen this once you've done so? As @kuzzmi mentioned, this is to ensure that your changes don't get overwritten at a later date.

buren:
Yes I have updated Transifex.

PR #1914, opened 2017-11-13
Update signrawtransaction.md

rhermanoff:
signrawtransaction only accepts an array of private keys and not a base58 string.  Providing a base58 private key string produces "error: Error parsing JSON: myPrivkey".


wbnns:
@rhermanoff Thanks for updating this.

PR #1940, opened 2017-11-23
Removed broken link

dimitris-t:
Removed broken link in https://bitcoin.org/en/development

This text points to http://whatcanidoforbitcoin.xyz that no longer
exists. The text behind the ‘morechoose’ translation string can be
found at
https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/_translations
/en.yml and currently reads "You can choose a project to contribute to
by answering a few <a
href=\"http://whatcanidoforbitcoin.xyz/\">questions about your
skills</a>."

The 'morechoose' translation string would become obsolete. If it is appropriate I can amend this PR to remove the string from https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/_translations
/en.yml, unless this task needs to be performed on Transifex.


dimitris-t:
Let me know if I should squash these 2 commits into 1.

wbnns:
@dimitris-t Nice catch, thanks for fixing.

PR #1965, opened 2017-12-08
Download latest DE translations from Transifex

buren:


buren:
Closing this, since there is a translated URL that contain "special" characters. Will fix it and then open a new PR.

PR #1969, opened 2017-12-10
Fix typo in full-node.md

alvarogzp:
`allot` to `allow`


laanwj:
I don't think "allot" is a typo, it is a correct use of the word.

alvarogzp:
Oh, it's true, I didn't know the meaning of that word. Sorry, I am not a native English speaker.
So, may I close the PR?

PR #1975, opened 2017-12-11
Fix typo in translations at ja.yml

khrtz:
`WIndows` to `Windows`


wbnns:
@khrtz 

Thanks for fixing this. I corrected this upstream on Transifex so that it will get updated the next time the strings are updated.

PR #1357, opened 2016-09-05
Adding BTC.com wallet

AlejandroDeLaTorre:
Hello,

We'd like to add BTC.com's wallet to Bitcoin.org. Our former name was Blocktrail, but we've rebranded to BTC.com. Please let me know if I need to correct or add something more. 

Thank you

PR #1979, opened 2017-12-12
Add BTC.com wallet

AlejandroDeLaTorre:


crwatkins:
Based on [my previous review](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1357#issuecomment-282388546), I concur with the scoring and I support this PR.

utACK

Cobra-Bitcoin:
I don't think we should ever add NYA companies back on to bitcoin.org.

crwatkins:
I believe that this wallet meets our wallet listing criteria, and thus we should list it. If we want to change our listing criteria, that would be a matter for another PR.

Cobra-Bitcoin:
@crwatkins Agreed that it meets the listing criteria. Would have no issue if this was merged.

But since they only recently just did something crazy like supporting a contentious hard fork, I just don't think it makes sense to point users to this service. Likely they won't ever do anything like that again, but it looks bad on us if we are so quick to forgive reckless behavior.

crwatkins:
As I have [said](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1337#issuecomment-241895424) in the past, I personally view the bitcoin.org wallet listings as a service to folks looking for a wallet to use.  Thus, my major focus is providing a list of wallets that are safe to use.  Many of our criteria require wallets to function transparently in ways that minimize the potential harm to users regardless of future decisions of the developers.  I know I'm being a bit naive, but I don't have the resources or the aptitude to analyze the potential future decisions of all developers that want to list wallets, so I rely on minimizing the impact of unknown future decisions.

As such, this wallet complies with the criteria and I believe it would be in the interest of our users to merge this.

PiscisSwimeatus:
ACK

AlejandroDeLaTorre:
Happy new years everyone, could we get an update please on this PR? Thanks

AlejandroDeLaTorre:
Are there any issues with this PR that we need to work on?

TommyHorber:
Tommi_23

AlejandroDeLaTorre:
hey all, is there anything we are missing?

crwatkins:
I believe we should either move this merge forward, or propose/discuss/merge a policy which explains why not.  I don't believe it is in the interest of our open community to simply ignore PRs.

That said, I currently support merging.

Cobra-Bitcoin:
This is scheduled to be merged on Wednesday 12th of September unless others disagree.

PR #2046, opened 2018-01-02
Added news source in Spanish

hcarpach:
CriptoNoticias is the most read digital newspaper about Bitcoin in Spanish since 2015. We have over 1 million views per month with exclusive news about Latinamerica and Spain and other significant topics related to cryptocurrencies.


Cobra-Bitcoin:
ACK.

PR #2076, opened 2018-01-15
Update tr.yml

gulsum8119:
added more deductive turkish word for the better translation.


wbnns:
@gulsum8119 Hello, thanks, I'll update this on Transifex.

PR #2124, opened 2018-02-03
Add to the list of exchanges.

godexdev:
Hello!
Can you please add Godex to your exchanges list.
Thanks in advance!


wbnns:
@godexdev Hello, thanks for your submission. It looks like your order page accepts invalid address strings, which would make it difficult (or potentially impossible depending on volume) to accurately match and settle customer orders:

image

Please let us know when this is fixed and we will resume the review.

godexdev:
Good day! 
Thank you for paying attention to this defect. 
Now it's fixed, you can check it out.

PR #2180, opened 2018-02-25
The time field was missing in response

boris-kolar:


wbnns:
@boris-kolar Thanks for the fix!

PR #2221, opened 2018-03-19
bitcoin.org redesign

alexcherman:


wbnns:
@alexcherman 

Thanks, could you please update this PR each day with new commits so we can review the daily progress?

cc: @natiwa  @marekkirejczyk

alexcherman:
Ok, Will

Cobra-Bitcoin:
There's something wrong with the way the way container for the content separation and scrolling works, for example there's a bar showing which shouldn't be the case: https://i.imgur.com/pt6qACa.png. Also the containers should always be collapsed by default, but right now you need to click to open them, https://i.imgur.com/TRoGLzq.png.

It should behave exactly the same as the current site, where you can just scroll endlessly to the end without having to click anything and the tracker just follows you along. Right now it feels clunky and strange, I hope this is just a temporary thing, and not the intended final design.

Cobra-Bitcoin:
Also let's keep updating this spreadsheet: https://docs.google.com/spreadsheets/d/17ncFqPXjiFlyA3i2rk3mq-kHuxpl_DqafhqjZPjHDY8/edit#gid=0.  I can see that some of the things marked as "To do" have been done now. Really important to update that so we have a clear view of exactly what is left to be done.

wbnns:
@alexcherman Hey, since we're getting closer to going live, please open PRs for any subsequent changes against the [redesign branch](https://github.com/bitcoin-dot-org/bitcoin.org/tree/redesign).

PR #2224, opened 2018-03-21
Update hard-fork.yaml

kshilov:
Just tried to create great guide which will show the whole picture to the reader.


wbnns:
@kshilov Thanks for the suggestion, however, this is a bit too broad given the context of the post.

PR #2261, opened 2018-04-05
Update getinfo.md

alexandersjn:
Just misspelling


MarcoFalke:
You might as well remove it as a whole, since it no longer exists in the latest release of Bitcoin Core.

See https://github.com/bitcoin/bitcoin/blob/0.16/src/rpc/misc.cpp#L448-L458

paulvi:
in #2318 just added that it was removed in 0.16.0

@MarcoFalke 
Some developers may be looking for it, if removed completely, then it would be feeling of broken/incomplete docs.

PR #2327, opened 2018-05-03
Adding links to fxempire.com/crypto/exchanges

iritru:
Adding link to fxempire.com


wbnns:
@iritru Hello, this website has multiple pop-up notifications and modals. Also, there appears to be no rhyme or reason as to how the exchanges are ordered on the ranked exchanges list - all of them have 0 reviews and ratings of 0.0. Lastly, the right sidebar is repetitively keyword stuffed with "buy [insert_cryptocurrency]". 

We wouldn't be able to add this as a helpful addition to the resources page at this time.

PR #2354, opened 2018-05-10
subsequent redesign changes

alexcherman:


wbnns:
@alexcherman Thanks! Since this makes the entire box clickable, what happens to the icons that appear inside of the box, and what is the default wallet page that opens when a wallet is selected?

wbnns:
@alexcherman Thanks.

PR #2381, opened 2018-05-21
Add new Exchange

saintclair:
We are the **Groupon** of Cryptocurrencies!

We are experiencing a dematerialization of the FIAT currency, since cryptograms can not be classified as a financial investment, nor as an active asset, they are digital assets. Therefore, as we discuss how, by using blockchain we allow a stock exchange or asset through a process that resembles a *Digital Barter*, this model allows us to be *Well more than a Cryptocurrencies*

### DIFF
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -63,6 +63,7 @@ id: exchanges
     <div>
       <h3 id="brazil"><img src="/img/flags/BR.png" alt="Brazilian flag">Brazil</h3>
       <p>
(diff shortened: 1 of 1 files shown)


saintclair:
@wbnns sorry for the anxiety, yesterday when we did the PR the platform was not still in the air, can evaluate again, the label this red I imagined it had been rejected, an important observation, we have nothing to do with the 3xbit dot com, was a misfortune in in the choice.

PR #2384, opened 2018-05-21
Remove coinapult wallet

crwatkins:
PR #1706 requires that listed wallets provide access to private keys, and thus wallets scored as

```
control: "checkfailcontrolthirdparty"
```
should not be listed.  Coinapult is the only wallet left in the listings meeting this criteria and this PR removes that listing.  Coinapult may qualify for listings elsewhere on the site.


wbnns:
@crwatkins Thanks.

Unless others object, this will be merged on Wednesday, May 23rd.

PR #2387, opened 2018-05-22
add value to offline signing

frankegoesdown:
value is required field for offline signing


wbnns:
@frankegoesdown Thanks and congrats on your first contribution to the repository.

PR #2388, opened 2018-05-22
Corrected a following to subsequent

aureliogrb:
Changing a transaction has no impact on previous blocks, it only impacts the block on which the transaction occurs and the subsequent/following blocks.


wbnns:
@aureliogrb You're right, thanks for fixing. This will be merged once tests pass.

PR #2392, opened 2018-05-24
German translation

99coins:
Added German translation to the Bitcoin whitepaper


wbnns:
@99coins Hello, thank for doing this. We'll carry it over from the whitepaper repo where all of the translations originate.

Issue #1633, opened 2017-06-15
Mention IBD


paveljanik:
[Bounty](https://github.com/bitcoin-dot-org/bitcoin.org#earn-bitcoin-for-contributing): **13,500 bits / ~$110 USD**

The page https://bitcoin.org/en/full-node#what-is-a-full-node should mention IBD more extensively (https://github.com/bitcoin-core/bitcoincore.org/issues/349#issuecomment-285590983).

It would be handy to have a FAQ/IBD menu entry on the web so we can redirect people when asking why Bitcoin Core GUI is not doing anything on the first run, why it is consuming 100% CPU and IO at the first run etc.

nickthetait:
Is this page sufficient? https://bitcoin.org/en/developer-guide#initial-block-download
If not, what information is missing? Or does it need to be simplified for someone new to Bitcoin?

paveljanik:
The page is for developers. The description of the first run, high CPU, network traffic and storage is missing. Together with -prune description. For "joe" user.

PR #2451, opened 2018-06-16
Add IBD section in the full node guide

rod-lin:
Hi! I made a (possible) fix for the issue #1633. Not really confident with my wording and style though...

### DIFF
--- a/en/full-node.md
+++ b/en/full-node.md
@@ -222,7 +222,7 @@ have an easy-to-use node.
 
 ## Initial Block Download(IBD)
 
-[Initial block download](https://bitcoin.org/en/developer-guide#initial-block-download)
+[Initial block download](/en/developer-guide#initial-block-download)
 refers to the process where nodes synchronize themselves
 to the network by downloading blocks that are new to them.
 This will happen whenever a new or offline node starts up.
@@ -234,8 +234,7 @@ During the download, there could be a high usage for the network and CPU
 (since the node has to verify the blocks downloaded), and the client will take up an
 increasing amount of storage space. Before the node finishes IBD, you cannot send/receive Bitcoins.
 
-If you are using Bitcoin Core GUI, you can monitor the progress of IBD at the
-bottom left corner in the software.
+If you are using Bitcoin Core GUI, you can monitor the progress of IBD in the status bar(left bottom corner).
 
 ![Bitcoin-Qt Initial Block Download](/img/full-node/en-bitcoin-qt-ibd.png?{{site.time | date: '%s'}})
 
--- a/en/full-node.md
+++ b/en/full-node.md
@@ -230,7 +230,7 @@ In the process of IBD, a node does not accept incoming transactions nor request 
 
 If you are trying to set up a new node following the instructions below, you will go
 through the IBD process at the first run, and it may take a considerable amount of time since a new
-node has to download the entire blockchain (which is roughly {{site.text.chain_gb}} now).
+node has to download the entire block chain (which is roughly {{site.text.chain_gb}} now).
 During the download, there could be a high usage for the network and CPU
 (since the node has to verify the blocks downloaded), and the client will take up an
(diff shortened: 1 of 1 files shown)


en/full-node.md
```diff
@@ -220,6 +220,24 @@ have an easy-to-use node.
 
 <div class="toccontent-block boxexpand expanded" markdown="1">
 
+## Initial Block Download(IBD)
+
+[Initial block download](/en/developer-guide#initial-block-download)
+refers to the process where nodes synchronize themselves
+to the network by downloading blocks that are new to them.
+This will happen whenever a new or offline node starts up.
```
harding: IBD is a specific mode of operation for BItcoin Core that does various things to prioritize download of new blocks.  For example, it doesn't accept incoming connections, doesn't request unconfirmed transactions it hears about from other nodes, and uses the memory usually dedicated to the mempool to extend the UTXO cache.

Because those are significant changes to the operating mode, Bitcoin Core only enters IBD mode if it detects that it is 100 or more blocks behind the tip of the chain---not just whenever an offline node starts up.

en/full-node.md
```diff
@@ -220,6 +220,24 @@ have an easy-to-use node.
...
+
+If you are trying to set up a new node following the instructions below, you will go
+through the IBD process at the first run, and it may take a considerable amount of time since a new
+node has to download the entire blockchain(which is more than 100 gigabytes by the time this paragraph is written).
```
harding: 1. There should be a space between "blockchain" and "(which"; same for the parenthesis is the subsequent paragraph.

2. For the size of the block chain, you can use the variable `{{site.text.chain_gb}}`, which is used elsewhere in this document if you want to see an example.  This allows someone to update the variable definition in `_config.yml` periodically to reflect the current chain size.  If you want to do it in this PR, this variable can probably be updated now as it says 145 GB but the current size is,

```text
$ du -sh .bitcoin/blocks/
183G    .bitcoin/blocks/
```

en/full-node.md
```diff
@@ -220,6 +220,24 @@ have an easy-to-use node.
...
+During the download, there could be a high usage for the network and CPU
+(since the node has to verify the blocks downloaded), and the client will take up an
+increasing amount of storage space. Before the node finishes IBD, you cannot send/receive Bitcoins.
```
harding: You can generate an address and receive bitcoins on the network---you just won't see them until you've caught up to the block that contains the transaction.

harding: Also, for the "increasing amount of storage space" part, it may be worth mentioning the pruning section described later in the doc that allows the user to restrict the disk space to a moderate constant amount.  See https://bitcoin.org/en/full-node#reduce-storage

rod-lin:
@harding Thanks for the editing, David. I've fixed the issues.

en/full-node.md
```diff
@@ -225,14 +225,19 @@ have an easy-to-use node.
...
 refers to the process where nodes synchronize themselves
 to the network by downloading blocks that are new to them.
-This will happen whenever a new or offline node starts up.
+This will happen when a node is far behind the tip of the [best block chain](/en/glossary/block-chain).
+In the process of IBD, a node does not accept incoming transactions nor request mempool transactions.
 
 If you are trying to set up a new node following the instructions below, you will go
 through the IBD process at the first run, and it may take a considerable amount of time since a new
-node has to download the entire blockchain(which is more than 100 gigabytes by the time this paragraph is written).
+node has to download the entire blockchain (which is roughly {{site.text.chain_gb}} now).
```
wbnns: This instance of the word "blockchain" needs to be split into two words as well, for conformity.

rod-lin:
@wbnns Ok, fixed.

wbnns:
@rod-lin Thanks.

This is scheduled to be merged on Wednesday, July 4th.

rod-lin:
@wbnns OK. Thank you!

PR #2462, opened 2018-06-20
Upgrade wallet privacy disclosure scoring

crwatkins:
This request makes two "housekeeping" changes:
```
checkpassprivacydisclosurefullnode -> checkgoodprivacydisclosurefullnode
checkfailprivacydisclosurespv -> checkpassprivacydisclosurespv
```
By "housekeeping" I mean that this will change internal scoring symbol names that currently have no affect on what users see on bitcoin.org.

Currently the privacy disclosure scores are as follows:
```
good: (none)
pass: fullnode
fail: spv, centralized, & account
```
I try not to get too caught up in the definitions of the words "good", "pass", and "fail" because these are just internal symbol names that end users of the site never see.  I sometimes think of them as "top", "middle", and "bottom".

Currently SPV wallets are lumped in the bottom group with centralized servers and servers that require an account.  This request elevates fullnode wallets to good status and elevates SPV wallets to pass status to distinguish them from the others:
```
good: fullnode
pass: spv
fail: centralized & account
```
The reason for this internal change is to prepare for some potential future efforts which would be done to help users compare wallets.   While most of the current scoring categories have about 25% of the wallets in the bottom group, the privacy disclosure category has about 75% of the wallets in the bottom group.

In summary, I propose this change for two reasons:

1.  For privacy disclosure scoring, I don't believe that SPV wallets should be lumped into the same group as wallets which use centralized servers and user accounts.

2. This scoring category is currently very bottom heavy and provides too little differentiation with no entries in the top group and 75% of the entries in the bottom group.

To reiterate, this is currently only an internal change.  No bitcoin.org pages currently display the good/pass/fail attribute of the privacy disclosure score.


harding:
I have no objections to non-visible changes desired by the person doing the work of wallet review, however I do think it's worth commenting on the reasons for current scoring.

Wallet privacy can generally be divided into two classes, receiving privacy and sending privacy.

- Full nodes by default have information theoretical perfect receiving privacy because they simply download every transaction on the block chain.  That means anyone who observes your TCP/IP traffic can tell that you're a Bitcoin user (or at least Bitcoin curious) but they can't see what transactions you're interested in because you query the block chain locally.

    For sending privacy, full nodes by default aren't as good, although Bitcoin Core and Knots are better than they were when @saivann first chose these scores.  Anyone can can observe *everyones'* TCP/IP traffic could determine which IP address originated a transaction.  Maybe the NSA does this already.  However, we know for sure that there are companies such as Chainalysis and Blockchain.info that operate numerous spy nodes that connect to individual's nodes to record which node is first seen relaying a transaction.  In earlier versions of Bitcoin Core, the first node observed relaying a transaction was likely the node that originate it.  More recently, Bitcoin Core has introduced some deliberate delays into transaction relay for different sets of peers in order to make first-seen-relay less correlated with transaction-origin.  More advanced techniques such as the [Dandelion Protocol](https://arxiv.org/abs/1701.04439) are under active development.  (Full nodes can also use Tor for further increased sending privacy, but Bitcoin.org's wallet scoring has always focused on the default state of the wallet for typical users, which I would not consider to include Tor.)

- P2P SPV clients, those that connect to random nodes, have abdysmal privacy for receiving.  BIP37 bloom filters were theorized to provide privacy during their implementation, but in practice P2P SPV clients have set their false positive rate so low (in order to reduce bandwidth) and regenerated the filters so often that they don't provide any practical privacy against spy nodes such as those run by Chainalysis.  This is not the assertion of some neckbeard privacy geek, but the statement of record by one of the developers who proposed and implemented BIP37, Mike Hearn: https://groups.google.com/forum/#!msg/bitcoinj/Ys13qkTwcNg/9qxnhwnkeoIJ  BIP157/BIP158 offer a potential solution to receiving privacy for SPV clients---not information theoretical private like full nodes, but much better than the current case; however, only one client I know of (Neutrino) uses it and the specification is still not finalized and implemented in Bitcoin Core, so it'll probably be a year or so before other P2P SPV clients start using it.

    For sending privacy, the P2P SPV client story is just as bad.  Unlike full nodes, clients don't relay other people's transactions, so every transaction they send belongs to their wallet.  That means any transaction they send to one of the spy nodes out there is almost certainly recorded in a database with the client's IP address.  When Dandelion is implemented, P2P SPV clients will be able to take advantage of it, which may increase their privacy, although again they don't have the plausible-deniable relay of full nodes, so it's still not the same degree of privacy.

- For client SPV wallets such as those relying upon a single centralized service (like BitGo) or decentralized servers (like Electrum), wallet user privacy is fully dependent upon the server host.  If the server keeps logs (and maybe sells them), then this is as bad as current P2P SPV.  If the server doesn't keep logs or sell them, then the user has fantastic privacy.  However, with current server implementations, there's no way to know.  (There are some ideas for how to prevent servers from collecting private information, such as Private Information Retrieval (PIR) protocols, but AFAIK these are not on any implementation's roadmap.)

So, in summary, my preference for scoring would be:

    good: none or full nodes; definitely full nodes after Dandelion has been deployed for several months and seems to be working
    pass: full node (now), P2P SPV clients that use BIP157/158 and that send transactions using Dandelion if that seems to be working
    fail: P2P SPV, client SPV, and account

wbnns:
@crwatkins @harding Thanks!

This is scheduled to be merged on Monday, June 25th to provide additional time in case anyone else would like to contribute feedback.

crwatkins:
@wbnns I would like to hold off on merging.  There's no hurry at all and there's more to think about and discuss.  I agree with all of @harding's points and I think @saivann had incredible foresight in choosing these scorings.  Perhaps there is a better option to provide the differentiation that I'm looking for.

wbnns:
@crwatkins Ok, no problem.

crwatkins:
[Above](https://github.com/bitcoin-dot-org/bitcoin.org/pull/2462#issue-196263209) I said

>I try not to get too caught up in the definitions of the words "good", "pass", and "fail" because these are just internal symbol names that end users of the site never see. 

I think I've changed my mind on that and I believe that by merging this PR I would be co-opting the original definitions of these labels which I don't think is the right thing to do.  If someone would want to redefine these ratings (say to high medium and low) then all of the labels should be changed to that.  Just changing one category and making that inconsistent with the naming is not a good idea, so I'm closing this PR.

I still think that the category could use more differentiation, but this isn't the right way.

PR #2465, opened 2018-06-22
Alerts: Bitcoin Network Alert Retirement: Fix broken table

harding:
On https://bitcoin.org/en/alert/2016-11-01-alert-retirement#the-retirement-plan , the markdown-table is broken by the div added to the end of it:

This should fix it.  Note, I only tested with a quick preview not a full site build.

This is related to the links being added in https://github.com/bitcoin-core/bitcoincore.org/pull/567


wbnns:
@harding 

Thanks for catching and fixing:

image

PR #2498, opened 2018-07-05
exchanges: Add Dasset (New Zealand)

sm-dasset:
Hello,

Please accept the changes made to exchange.html that has included Dasset which is an exchange based in New Zealand that accepts New Zealand Dollars in exchange for BTC amongst other cryptocurrencies.

Best Regards,
Dasset Team

### DIFF
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -248,6 +248,8 @@ id: exchanges
 <div class="toccontent-block boxexpand expanded">
   <h2 id="new-zealand" class="exchanges-tab-title exchanges-newzealand-title">New Zealand</h2>
   <p>
+    <a class="marketplace-link" href="https://www.dasset.co.nz/">Dasset</a>
+    <br>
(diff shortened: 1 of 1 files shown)


wbnns:
@sm-dasset 

Hello, I tried to create an account on the exchange, however, despite verifying my email address, it says invalid username or password.

Also, upon sign up, the site asks for an email address to register, along with a password to associate with the account. Once a verification email is received, the login prompt on the website asks for a username.

There is no interface where a user provides a username to be created, so I'm assuming this is the same as the email address.

Please let us know when these issues are resolved so we can review the exchange in more detail.

sm-dasset:
@wbnns 

Hi, your username would be the email you signed up with.

Customers used to sign up with a separate username from their email address, however we removed the username function so customers do not need to remember a separate username to login. We changed this about 2 weeks ago, so most customers still have a username. However, this has not affected new customers signing in and logging into their accounts with their email address.

PR #2519, opened 2018-07-14
update to replaced Desktop app, deprecate web

greenaddress:
-  greenaddress web use was already discouraged, now deprecated
-  google deprecated chrome apps, greenaddress chrome app replaced by desktop app

Also updates from checkfailtransparencyremote to checkpasstransparencyopensource - the argument was that people wouldn't use the chrome app with web, but both web and chrome app are deprecated and the app is not loaded remotely.

### DIFF
--- a/_wallets/greenaddress.md
+++ b/_wallets/greenaddress.md
@@ -5,7 +5,7 @@
 id: greenaddress
 title: "GreenAddress"
 titleshort: "Green Address"
-compat: "mobile desktop web android ios windows mac linux"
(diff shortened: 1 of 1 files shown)


greenaddress:
Not clear to me why the CI build is failing .. any idea?

greenaddress:
@crwatkins thank you for the feedback. I removed web as suggested (I missed it!) and squashed the fixup in the relevant commit.

crwatkins:
LGTM

wbnns:
@greenaddress @crwatkins Thanks.

PR #2523, opened 2018-07-16
Update German Translation

wbnns:
This updates the site's German translation with all of the latest strings. Thanks @Komodorpudel, for all of your help! 

This is scheduled to be merged on Wednesday, July 18th.


_config.yml
```diff
@@ -106,6 +106,12 @@ redirects:
   /tr/yeni-baslamak: /tr/baslarken
   /en/bitcoin-for-press: /en/press
   /de/ueber: /de/faq
+  /de/bitcoin-paper: /de/bitcoin-abhandlung
+  /de/developer-examples: /de/enwickler-beispiele
```
Komodorpudel: Hey Will, made a small mistake there. Should be "entwickler-beispiele" and not "enwickler-beispiele". Already corrected it on Transifex

wbnns: @Komodorpudel Thanks! Will update here as well.

PR #2529, opened 2018-07-19
External Link Maintenance (Drop/Update)

wbnns:
There were several external links the website was pointing to that were either dead or needed to be updated. This PR drops them and/or updates them so we don't send users to websites or web pages that no longer exist.

This is scheduled to be merged on Saturday, July 21st.

Thanks to @ap4y for helping set up a compatible regex for HTML Proofer.


docs/contributing-to-developer-documentation.md
```diff
@@ -396,8 +396,7 @@ one of the following topics.
 
 * [ ] Multisig wallets like GreenAddress.it
 * [x] Hardware wallets like Trezor
-* [ ] Assurance contracts such as those created by [Lighthouse](http://blog.vinumeris.com/2014/05/17/lighthouse/)
-    * [ ] Mention that the recipients of assurance contracts can pay to their own contract to guarantee that they meet their goal and receive the funds.
+* [ ] Mention that the recipients of assurance contracts can pay to their own contract to guarantee that they meet their goal and receive the funds.
```
harding: I think this line that was a sub-point should also be removed, as it doesn't make sense to describe a vulnerability without describing the protocol that contains that vulnerability.

PR #2530, opened 2018-07-19
bither: Remove Bither for iOS

wbnns:
It seems that Bither for iOS [may no longer be available](https://itunes.apple.com/us/app/bither/id899478936) in the App Store. This drops the iOS option so users don't get misdirected to it.

Reached out to Bither Support to confirm (or perhaps others are able to load the page and it is some sort of localized or regional problem with the App Store). Will wait before merging just in case.

### DIFF
--- a/_wallets/bither.md
+++ b/_wallets/bither.md
@@ -5,28 +5,12 @@
 id: bither
 title: "Bither"
 titleshort: "Bither"
-compat: "mobile ios android desktop windows mac linux"
+compat: "mobile android desktop windows mac linux"
 level: 2
 platform:
   - mobile:
     name: mobile
     os:
-      - name: ios
-        text: "walletbither"
-        link: "https://itunes.apple.com/us/app/bither/id899478936"
-        source: "https://github.com/bither/bither-ios"
-        screenshot: "bithermobile.png?1528322191"
-        check:
-          control: "checkgoodcontrolfull"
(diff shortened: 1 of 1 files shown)


songchenwen:
Hi @wbnns , I'm the CTO of Bither team. Thank you for contacting us.

Apple removed Bither for iOS from App Store due to our Chinese company name does not reflect the app's name. It seems that Apple suddenly decided to take Bitcoin wallets as financial institutes so that more restricted review guidelines are applied to Bither. We don't know what makes Apple do this.

Recently we have built a new Canada company with the name of BITHER Limited and recruited a new team. Bither iOS app's ownership was transferred to our new company. According to the review guidelines, our newly submitted version should be approved. But because Bither was previously removed, Apple needs more time to review our new submission. This review process can take another one or two weeks according to Apple and cannot be expedited.

In the last few months, the new company and team things have took us quite some time. We have done a lot of communication with Apple and even more communication through the Cananda company with Apple. Now all the issues have been resolved, and our company name reflects the app name. The only thing left is to wait for Apple's review. After the first App Store review Bither iOS should be back as normal.

image

wbnns:
@songchenwen Hello, thanks for letting us know. Sorry to hear about the troubles. Please keep us updated on how it goes with the review. In the interim, we'll remove the link so that users don't get sent to an app that is no longer available.

Issue #2532, opened 2018-07-20
QR Code not appearing on certain pages


Cobra-Bitcoin:
[Bounty](https://github.com/bitcoin-dot-org/bitcoin.org#earn-bitcoin-for-contributing): **40,000 bits / ~$300 USD**

When clicking on the "Donate" button on the footer in certain pages, the QR code doesn't display in the modal. This happens on both Firefox and Chrome. However in some pages, the QR code shows up fine. This is despite the presence of all the necessary DOM elements and JS libraries in the source code. The error shown in the console is pasted below:

Uncaught TypeError: $(...).qrcode is not a function
    at generateDonationQrCode (main.js?1531922758:463)
    at openDonationModal (main.js?1531922758:542)
    at HTMLAnchorElement.onclick (events:528)

The JS code causing the error: `$('#donation-qr-code').qrcode({ width: 150, height: 150, text: text });`. The code can be found on line 463 in main.js: https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/js/main.js#L463.

Examples of pages where the QR code shows up fine:
- https://bitcoin.org/en/
- https://bitcoin.org/en/choose-your-wallet
- https://bitcoin.org/en/community

**Examples of pages where the bug occurs**:
- https://bitcoin.org/en/events
- https://bitcoin.org/en/vocabulary
- https://bitcoin.org/en/faq

I suspect it's an issue with the JS code behaving in an awkward way and corrupting something, since a lot of the pages with the bug have more interactive elements and behaviours. This bounty will go to the first person who fixes this issue in a pull request (which gets merged).

RobinDeBaets:
You're loading jQuery again after loading the QR code library here: https://github.com/bitcoin-dot-org/bitcoin.org/blob/71862011a76477822ce706c31659ddcf9df95f34/_templates/events.html#L68

johneakin:
Isn't the payment of this bounties a perfect thing to make a smart contract out of? ;)

PR #2533, opened 2018-07-20
Fixed qrcode JavaScript error

RobinDeBaets:
jQuery was sometimes loading twice, which caused the old version (injected with the qr version), to be void.
I fixed this by injecting the qr code script again and by removing some duplicate jQuery imports.

Fixes: #2532 
Address: 3BPucBK5Y6ybvouBGw7husQfwk86BRkSQE

### DIFF
--- a/_layouts/about-sidebar.html
+++ b/_layouts/about-sidebar.html
@@ -3,7 +3,6 @@
 # http://opensource.org/licenses/MIT. 
 layout: base
 end_of_page: |
-  <script src="/js/jquery/jquery-1.11.2.min.js"></script>
   <script>updateToc();</script>
   <script>accordion();</script>
   <script>onScrollButton();</script>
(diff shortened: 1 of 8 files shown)


hesido:
Congrads, it was obvious that the extended jquery object was getting re-written somewhere, re-executing the qrcode script made it work. Too bad I had to fix my kids some dinner.

wbnns:
@RobinDeBaets Congrats on your first contribution to Bitcoin.org and thanks for the help!

Here's the [transaction ID](https://live.blockcypher.com/btc/tx/835d4d7f846e6d6ffa3193a860ab4c048acf83a8b763075fb8bbd64193201495/) for the bounty you took down.

RobinDeBaets:
Thanks a lot for the bounty, glad I could help.

PR #2534, opened 2018-07-20
Move deterministic nonce wallet criteria from optional to required

crwatkins:
I propose moving the deterministic nonce wallet criteria from optional to required.

It's just a good idea.  The only downside I've ever heard expressed is a 10% performance hit.  In the past three years all new wallets listed on bitcoin.org have signed with deterministic nonces.


harding:
ACK.

wbnns:
@crwatkins @harding Thanks.

This is scheduled to be merged on Wednesday, July 25th.

PR #2539, opened 2018-07-23
Update exchanges

godexdev:
Add Godex to the list of exchanges.


wbnns:
@godexdev Hello, in reviewing the page, we were not able to progress beyond the home page - nothing happened when the "Exchange" button was pressed. We also noted some strange red text that was overflowing onto the input field. Here is the observed behavior as well as the console output:

screen recording 2018-07-26 at 16 56

PR #2543, opened 2018-07-24
Exchanges: Including brazilian exchange PagCripto

Raizen:
For more information, see www.pagcripto.com.br

### DIFF
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -280,6 +280,8 @@ id: exchanges
           <br>
           <a class="marketplace-link" href="https://www.omnitrade.io/">OmniTrade</a>
           <br>
+          <a class="marketplace-link" href="https://www.pagcripto.com.br/">PagCripto</a>
+          <br>
           <a class="marketplace-link" href="https://walltime.info/">Walltime</a>
         </p>
       </div>


wbnns:
@Raizen Hello, it looks like the current exchange rate for one bitcoin here is almost $900 higher than the [global average bitcoin price](https://bitcoinaverage.com/en/bitcoin-price/btc-to-usd).

Raizen:
@wbnns That's why the last big execution was greater than our book availability. Our normal exchange rate are equal the others Brazilian exchanges. Can you review it, please?

wbnns:
@Raizen Hi, understood. I reviewed all of the other Brazilian exchanges yesterday and did not see an instance where their advertised BRL/BTC exchange rates were equal to yours - they were all less.

Raizen:
@wbnns That's because like I said, we have received a big buy order who clean our sell book until it. It still not normal, actually we are 1% higher than the others because that. That not mean that we are scam or similar.

wbnns:
@Raizen Thanks for the clarification. It is over 12 hours later, however, and the current exchange rate still seems to be the equivalent of $700 higher than the global average spot price.

Just trying to help people get a fair price, especially those who are new to bitcoin.

Raizen:
@wbnns If you look today, we are in the market price for BTC in Brazil. If you check the brazilian history, we have almost everyday +4% in comparision with binance or bitfinex prices. $700 higher than the global are normal for the Brazilian market...

PR #2551, opened 2018-07-26
Added new resource

dannmat:
- Added new link to https://www.bitcoinregulation.world in the resources section

### DIFF
--- a/_templates/resources.html
+++ b/_templates/resources.html
@@ -42,6 +42,9 @@ id: resources
         <p>
           <a href="http://www.iamsatoshi.com/category/video-archive/">IamSatoshi</a>
         </p>
+        <p>
+          <a href="https://www.bitcoinregulation.world">Bitcoin Regulations</a>
+        </p>
       </div>
 
       <div class="card resources-card">


wbnns:
@dannmat Hello, thanks for sharing the link and letting us know about the site. Many visitors come to Bitcoin.org, and I'm not sure this resource in its current form would give them a comprehensive-enough perspective. For example, many visitors are from the United States, Latin America and Asia - most of the entries seem to be in the EU. Also, for a lot of the EU entries, they seem incomplete:

image

Many entries just say "n/a" in several of the data fields.

Perhaps when the site is more complete we can reconsider, or alternatively, if other active contributors disagree and think this should be added as-is.

dannmat:
Hi @wbnns , we're currently in the process of adding the United States, as the laws are mainly decided on a state-by-state basis, this is taking us a while, however by next week we should have a new build live with several of these states. We also plan to add the regulations for every country, which obviously takes a while.  In terms of the feedback about using "N/A", I'll take this into account :)

I will contact you again when we have more countries 👍 

Thanks,
Matt

PR #2558, opened 2018-07-27
Exchange Page Maintenance

wbnns:
I went through the exchanges on the Exchanges page and recommend we remove some listings:

#### Cambodia
- [Bitcoin Cambodia](https://bitcoincambodia.com/) - This exchange is no longer accepting new user registrations.

#### Canada
- [Bitaccess](https://bitaccess.ca/) - Their Buy/Sell button redirects to [another site](https://buybitcoincanada.com/) which was showing the current exchange rate of BTC/CAD as $9164.75 ($6993.01 USD) - this was off by almost a thousand dollars. Revisiting this page approximately 12 hours later, the price was unchanged - it was exactly the same. It seems they are providing an unreliable exchange rate.

#### China
- [OKCoin.cn](https://okcoin.cn/) - While it's possible to sign up, it does not seem possible to trade BTC/CNY. There are also no other exchange rates listed. It seems to be more-so of a wallet at this point, where people can withdraw. It does not appear that there is a way to deposit. Therefore, this seems like it would be confusing for new Chinese users who are looking to get started with bitcoin. Also, it seems there is more of an international focus at the company, now, on a completely different website - [OKEx](https://www.okex.com/).
- [Huobi](https://www.huobi.com/) - Huobi has also shifted focus. Their website is in English by default, they brand themselves as a global exchange. And it seems to be purely crypto-to-crypto. It doesn't seem possible to exchange Yuan. This may be very confusing for new users in China who want to get started with bitcoin.

#### Japan
- [Coincheck](https://coincheck.com/) - This exchange has suspended its operations and no longer accepts new user registrations.

#### India
- [Coinsecure](https://coinsecure.in/) - This exchange has suspended its operations and no longer accepts new user registrations.
- [Coindelta](https://www.coindelta.com/) - The Bank of India has issued a decree preventing INR deposits and withdrawals, therefore, people in India would not be able to deposit (or withdrawal) their local concurrency in order to exchange bitcoin.
- [Throughbit](https://www.throughbit.com/) - Throughbit is subject to the Bank of India ruling.
- [Zebpay](https://www.zebpay.com/) - Zebpay is also subject to the Bank of India ruling.

***

Also made a few additional cleanup-related fixes:
- Renamed SurBTC to Buda (they changed their name).
- Grouped continents/countries together (e.g. North America w/ South America, Australia w/ New Zealand).
- Cleaned up some of the code and spacing in exchanges.html.

***

This is scheduled to be merged on Monday, July 30th.

### DIFF
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -107,8 +107,6 @@ id: exchanges
         <h3 id="china" class="no_toc">China</h3>
         <p>
           <a class="marketplace-link" href="https://www.huobi.com/">Huobi</a>
-          <br>
-          <a class="marketplace-link" href="https://www.okcoin.cn/">OKCoin</a>
         </p>
       </div>
     </div>
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -232,34 +232,6 @@ id: exchanges
 </div>
 
 <div class="toccontent-block boxexpand expanded">
-  <h2 class="exchanges-tab-title exchanges-australia-title" id="australia">Australia</h2>
-  <p>
-    <a class="marketplace-link" href="https://bitcoin.com.au/">Bitcoin Australia</a>
-    <br>
-    <a class="marketplace-link" href="https://www.coinjar.com/">CoinJar</a>
-    <br>
-    <a class="marketplace-link" href="https://www.coinloft.com.au/">CoinLoft</a>
-    <br>
-    <a class="marketplace-link" href="https://www.coinspot.com.au/">CoinSpot</a>
-    <br>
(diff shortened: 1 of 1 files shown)


Cobra-Bitcoin:
LGTM.

UpalChakraborty:
It is not only Coindelta which has suspended INR deposits. Every Indian exchange, I know of, has done the same. For example, here is the same announcement from ZebPay - https://twitter.com/zebpay/status/1014431241226113024

wbnns:
@UpalChakraborty Thanks for mentioning this.

PR #2568, opened 2018-08-03
Added Bitex.la argentinian exchange

norchow:


_templates/exchanges.html
```diff
@@ -286,6 +286,8 @@ <h3 id="argentina" class="no_toc">Argentina</h3>
           <a class="marketplace-link" href="https://www.buenbit.com/">Buenbit</a>
           <br>
           <a class="marketplace-link" href="https://www.satoshitango.com/">SatoshiTango</a>
+          <br>
+          <a class="marketplace-link" href="https://bitex.la/">Bitex</a>
```
wbnns: Exchanges are listed alphabetically under each category/country. Please move this to ensure that is the case with this addition.

norchow:
@wbnns Thanks for the review. It's fixed now

wbnns:
@norchow Thanks. This is scheduled to be merged on Friday, August 17th.

norchow:
@wbnns That's great news! Thank you.

I wonder if we can apply to be listed also as an International Exchange, given that we accept fiat transfer from all over the globe. Is that possible? Do you need any more information?

Issue #2585, opened 2018-08-13
Wallet listings redundantly displays currently selected wallet


Cobra-Bitcoin:
If the user is viewing a wallet, for example: https://bitcoin.org/en/wallets/desktop/windows/bitcoincore/. The bottom wallet listings repeats the wallet the user is already viewing, which is redundant.

On another note, maybe we should bring back the top black bar navigation we discussed in the redesign? The wallet listings being on the bottom makes it easy to not see.

PR #2586, opened 2018-08-14
Hide currently selected wallet

Cobra-Bitcoin:
This fixes https://github.com/bitcoin-dot-org/bitcoin.org/issues/2585 and hides the currently selected wallet because it's redundant to display it.


odarboe:
ACK. I don't know how much concerns we have for indentation here. The indentation of the first change is different from the rest. ;)

crwatkins:
I could go either way on this one.  I don't mind having a complete list presented to the user each time.

odarboe:
ACK.

PR #2598, opened 2018-08-19
change "block to valid" to "block to be valid"

RandyMcMillan:
simple edit.


RandyMcMillan:
Contributing where I can.

PR #2603, opened 2018-08-21
Add Bitex exchange in Chile, Paraguay and Uruguay

norchow:


wbnns:
@norchow Hello, we did not see any specific exchange rates or deposit instructions for users in these additional locales. Please let us know where they are and we can continue the review.

PR #2626, opened 2018-09-05
Fix donation modal and footer bugs

Cobra-Bitcoin:
This fixes a number of bugs with the donation modal and footer and also improves the use of JS across the site to make it more consistent:

- The donation modal now displays consistently across the entire site (previously it didn't display properly on certain pages, like on the Core subsite, Download page, 'Running a full node' page, etc)
- Makes the footer identical across the site (footer on aforementioned pages is different to one on other pages), and keeps links all the same
- All jQuery libraries are now imported in the HTML head, as they should be, and not spread across a bunch of pages, as the previous situation was making it easy to duplicate load JS libraries

Please test and review, since it's possible I introduced some bugs, and I want to be really careful before we merge this. Thanks.


wbnns:
@Cobra-Bitcoin Thanks, I'm not seeing any new issues with the changes when testing locally.

This is scheduled to be merged on September 17th.

PR #2631, opened 2018-09-05
Correct list bullet overflow

jdabs:
Added padding to the ul so that the bullets aren't half-hidden behind the container. Not the most eloquent and would like feedback if there's better.
bullets-cut-off


wbnns:
@jdabs Could you please let me know what browser version and OS you're running so we can try to replicate the issue?

jdabs:
@wbnns you're fast 👍 I am seeing this on Chrome 68 on Win7

wbnns:
@jdabs Ok, thanks - we'll check it out.

wbnns:
@jdabs It looks like this change, as-is, breaks the alignment in Firefox. You may be able to unset the padding and overflow and set `list-style-position: inside;` on `.devprojectlist` to fix.

If you have any questions, please let us know.

jdabs:
Ok, thanks. I am setting up a more complete local clone environment and will send the changes in an update.

PR #2634, opened 2018-09-06
Add minor spelling fixes

jdabs:
Adverbs, which end in -ly, do not need hyphens. I am continuing onward with my documentation review and testing.


wbnns:
@jdabs Thanks.

PR #2645, opened 2018-09-12
Textual improvements

merland:


en/full-node.md
```diff
@@ -1042,14 +1042,6 @@ For example, to safely stop your node, run the following command:
 
 {{complete_list_of_commands_and_ibd}}
 
-
-When Bitcoin Core daemon first starts, it will begin to download the block
```
wbnns: We should leave this in, for many people it will still take a long time to download.

merland: I agree. The commit removes an entire duplicated paragraph. As the commit comment says, the markdown capture  {{complete_list_of_commands_and_ibd}} already contains the same wording. 
@wbnns

PR #2648, opened 2018-09-14
Update ubuntu bitcoin version 0.16.2 from 0.14.2

mattbourke:
I've only tested this on Ubuntu so I've not bumped all the occurrences of 0.14.2, I'm very short for time at present, sorry.


wbnns:
@mattbourke Hello, thanks. Yes, this would need to be tested and bumped across the page - not just in one location.

PR #2633, opened 2018-09-06
Fix for overflow-x scrolling on mobile phones

jdabs:
Scroll bar would show after the table of contents snapped to bottom, namely on phones in portrait mode. The extra padding seemed to be causing this. That padding is not longer needed once the TOC is at the bottom.

responsive-container-x-scroll

PR #2653, opened 2018-09-14
Update indent

jdabs:
Per change request.


wbnns:
@jdabs Thanks. Could you please add this as a commit to #2633 instead of opening a new PR?

PR #2654, opened 2018-09-14
 buy: Non-renewal of Paxful Sponsorship

wbnns:
The Paxful sponsorship ends on October 11th. There are currently no plans or discussions to renew it, so this PR is scheduled to be merged on that date.

We were hoping there might be an opportunity to have more of a long-term relationship with them, however, at present, we feel it may not be in the best interest of trying to represent differing perspectives if we renew the sponsorship. That said, we hope there will be more opportunities for sponsorships on the site in the future. Bitcoin.org is one of the only leading/most-visited websites in the space that isn't directly or indirectly fueled by companies seeking to unilaterally press their own for-profit initiatives. 

We thank Paxful for their support.

### DIFF
--- a/_includes/templates/index.html
+++ b/_includes/templates/index.html
@@ -7,7 +7,6 @@
     <div class="btn-container">
       <a class="btn btn-bright btn-home" href="/{{ page.lang }}/{% translate getting-started url %}">{% translate getstarted layout %}</a>
       <a class="btn btn-light btn-home" href="/{{ page.lang }}/{% translate choose-your-wallet url %}">{% translate choosebut getting-started %}</a>
-      <a class="btn btn-light btn-home" href="/{{ page.lang }}/{% translate buy url %}" data-proofer-ignore>{% translate title buy %}</a>
     </div>
 
     <div class="mainvideo">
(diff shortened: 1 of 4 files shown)


Cobra-Bitcoin:
What about also removing the "Buy Bitcoin" button from the homepage?

wbnns:
@Cobra-Bitcoin We can do that, too, but one of the first questions a lot of people ask is where they can get bitcoin, and our exchanges page was already one of the top pages before the button was there. Was thinking leaving the button would be more user-friendly and save people time from having to figure it out.

Cobra-Bitcoin:
Fair enough, maybe we can look more into the traffic patterns, but just think it would be a shame if people come to Bitcoin.org having searched 'Bitcoin', click the buy button, and then immediately go to an exchange or something without letting us at least inform them about what Bitcoin actually is. Seems better to let them learn about what they're buying and then forward them to the relevant services.

Maybe we can have a link at the top of the buy page to the getting started page with anchor text "We suggest first learning about what Bitcoin actually is and how it works before buying". Because that buy button is attractive a lot of 'dumb' clicks.

wbnns:
@Cobra-Bitcoin Ok, let's go ahead and drop it (e58694c). People who are specifically looking for a way to buy bitcoin will continue to end up on that page as an entry point to the site since it is one of the first results when entering "buy bitcoin" into search engines. It also remains accessible via the dropdown menu.

Cobra-Bitcoin:
I would like to also thank Paxful for their professionalism and flexibility throughout the sponsorship. It's been great working with them.

PR #2655, opened 2018-09-15
Add Findbits to Peer-to-peer exchange list

chanar:
Findbits is a new peer to peer Bitcoin marketplace with efforts to provide more quality and be a safer environment for crypto trading.


wbnns:
@chanar Hello, thanks for sharing. It looks like there are only approximately 20 sellers at the moment, and the first two that were shown had 23% and 25% fees. There would need to be a deeper order book and lower fees before this could potentially be added to the site.

chanar:
@wbnns Thank you, will create a new pull request when ready.

PR #2658, opened 2018-09-17
Added Bitcoinregulation.world site as a resource

dannmat:
After taking @wbnns feedback into account regarding the way we display information (not using N/A, but providing detail as to why) and adding more countries for areas such as US, South America and Asia, we believe our site is an even more useful resource than it was last time we added to Bitcoin.org and is ready for publication here.


wbnns:
Hello, it seems the site is intermittently giving errors:

image

We're unable to review. Please let us know when it's fixed and we'll take another look.

PR #2662, opened 2018-09-18
Show donation modal in download page

Cobra-Bitcoin:
This imports main.js into the download page so we can call the function to show the donation modal, which currently isn't displaying.

I screwed up in https://github.com/bitcoin-dot-org/bitcoin.org/pull/2626 a bit, but this fixes that. Suggest to merge as soon as tests pass.


wbnns:
@Cobra-Bitcoin Ok, will do - thanks.

PR #2663, opened 2018-09-18
Fix bullet overflow> #2631

jdabs:
Fix for changes suggested in #2631


wbnns:
@jdabs Thanks. Can you please consolidate this in #2631 instead of opening a new PR?

You can drop the previous commit in there and use this one.

PR #2666, opened 2018-09-21
Add an alert about CVE-2018-17144

Cobra-Bitcoin:
Adds an alert warning about the DoS / inflation bug (CVE-2018-17144) vulnerability in Bitcoin Core.

First person to ACK that also has merge privileges should merge ASAP if the text is acceptable.


theymos:
I think that the wording is a little too panic-inducing. Miners are believed to have upgraded sufficiently, so any incident at all is unlikely, and even in the worst case any illegitimate inflation would eventually be undone. Whether or not you upgrade won't affect the security of your *stored* bitcoins, but only the security of the network as a whole and perhaps whether you accept an invalid transaction as legit.

It also fails Travis CI for some reason. It's not immediately clear to me why.

luke-jr:
@theymos Miners are the potential attacker. What matters is the network-wide deployment, which is still dangerously low.

The wording sounds fine to me (although a bit perhaps too negative on Core's responsible non-disclosure of the full extent of the risk initially).

theymos:
I looked into the Travis CI failure, and I think that it's a bug in alerts.html rather than in this PR. If this is merged as-is, I think that it might break the alerts.html listing of alerts, but it'll probably otherwise work. I'm not set up to test it right now, though.

@luke-jr Since the vast majority of big economic players will have upgraded by now due to the DoS, even if the majority of miners turn explicitly malicious, their blocks would be ignored by most of the economy. It seems like a very unlikely attack with limited potential for harm.

I especially dislike:

> Furthermore, it's possible for miners to exploit this vulnerability to maliciously inflate the supply of Bitcoin.

They *can't* do that in any meaningful sense. They can create invalid transactions, and some nodes might temporarily accept them, but there is **zero** chance that those will survive long-term.

It should also be emphasized that coins at rest are not at risk. It's mainly an issue of accepting counterfeit BTC, basically.

wbnns:
Yes, there's a bug in the alert itself (not the PR).

Due to the time-sensitive nature of this PR, I've added a note to the home page in bc14a96 to temporarily workaround the alert issue.

I've also ported the [reddit post](https://www.reddit.com/r/Bitcoin/comments/9hkoo6/new_info_escalates_importance_upgrading_to_0163/) that @theymos made to give additional clarity/context. This is also to have a consistent message across two very high-traffic sites in order to hopefully avoid confusion among users (especially those who aren't running the software - a significant number of people).

Tests are now passing.

[Link to Preview.](https://bitcoin.cryptopelago.com/)

theymos:
Bitcoin.org should probably not link to bitcoincore.org for downloads or Reddit for BitTorrent / verification instructions, but it's OK. IMO merge it after the build passes, and we can nitpick it later.

wbnns:
Ok, build passed. Merged.

PR #2693, opened 2018-10-04
index: Drop 0.16.3 Upgrade Alert Notice from Homepage

wbnns:
This drops the 0.16.3 upgrade alert notice from the homepage.

It has been up for the last couple of weeks.


wbnns:
This is scheduled to be merged on Sunday, October 7th.

PR #2696, opened 2018-10-05
Adding cryptex24.com link

cryptex24:
Kindly add Cryptex24.com to the exchanges list. The service is based in Russia, but operates worldwide since 2014.


wbnns:
@cryptex24 Hello. We were not able to test signing up because:

1. The mobile dropdown menu does not work; nothing seems to happen when it is touched.
2. When viewing the site from a non-mobile perspective, there is no option for the United States in the dropdown, where many users are going to come from if you are listed in the international category on the Exchanges page (the category this PR adds the exchange to).

Also, according to Alexa rankings, the site does not appear to receive much traffic outside of Russia or Egypt, which would exclude it from the international category at this time:

image

PR #2702, opened 2018-10-11
Fix bugs in devsearch

Cobra-Bitcoin:
Fixes a bug in the search widget that is messing up the suggestions, bug is described below:

- Go to a page with the search widget visible (https://bitcoin.org/en/developer-documentation)
- Type "bump" into the box
- **Expected behavior**: "BumpFee" is the only thing present in the suggestions
- **Actual behavior**: _Everything_ in the glossary is being returned in the suggestions

This pull request fixes that bug by making sure the right results are returned, additionally the UX/UI of the widget is improved with the following additions in look and behavior:

- The color of the text in the input field is changed to be darker, since the current color is too faint and not ideal
- When the user clicks back into the box, the suggestions are displayed again, as is the behavior on search engines like Google/Bing
- When the user clicks back into the box, the cursor is automatically placed at the end, so the user can easily backspace their search and make a new one

### DIFF
--- a/_sass/screen.scss
+++ b/_sass/screen.scss
@@ -1035,7 +1035,7 @@ h1 span.fa, h2 span.fa, h3 span.fa, h4 span.fa, h5 span.fa, h6 span.fa {
     margin: 0;
     font-size: 137.5%;
     line-height: 1.9;
-    color: #FFFFFF; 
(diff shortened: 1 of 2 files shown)


wbnns:
This is scheduled to be merged on Sunday, October 14th.

PR #2703, opened 2018-10-12
Add new exchange to Australian exchanges list

rokanost:
FastX.io is Australia's newest cryptocurrency trading platform. One of our goals is to allow buying & selling crypto an ease. (quick and simple verification).

We are also almost done implementing PayID - it will take seconds to receive AUD from your crypto to bank.

For more info, please DM me. 
Thank you :)

### DIFF
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -374,6 +374,8 @@ id: exchanges
     <br>
     <a class="marketplace-link" href="https://www.cointree.com.au/">CoinTree</a>
     <br>
+    <a class="marketplace-link" href="https://fastx.io/">FastX</a>
+    <br>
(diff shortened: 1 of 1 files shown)


wbnns:
@rokanost 

Hello, we were unable to complete a review of this site because when we tried to register we encountered an Internal Server Error. The issue continued to persist after several attempts.

Also, without logging in, we noted it appears on the home page that there is a $581.18 AUD spread between buying one bitcoin and selling one bitcoin.

rokanost:
Hey, thanks for your response. 

The registration issue was just resolved now. 

Regarding the price, we have a 3% spread. 
However, I'm working on a new pricing model where we can offer best prices for higher cost cryptos like BTC and fixed margin for lower ones like XRP, NANO.  (FYI: We've been losing money so far).

PR #2670, opened 2018-09-22
Adding news source in portuguese Brazil.

Cmte-Kirk:
Livecoins is one of the main news, prices, and information on bitcoin, digital
currencies, and the blockchain movement as a whole in Brazil.

Over 300k people visit the site every month, including fintech/bank executives,
technology entrepreneurs and investors, digital currency traders, lawmakers and
regulators from government agencies, and journalists from major technology
publications.
[https://livecoins.com.br/](url)

PR #2715, opened 2018-10-15
resources: Drop news section

### LINKED ISSUE #2697: Dropping the 'News' section from the Resources Page
Link: https://bitcoin.org/en/resources

We recently had a pull request to add a language-specific news resource to the News section of the Resources page (so that it would only appeared when the page is viewed in that language), and I noted that the majority of the news was more tailored to cryptocurrencies and blockchain-related topics in general, rather than a specific focus on Bitcoin. The current news sites that appear on the resources page also somewhat reflect this trend. 

There's no shortage of news about Bitcoin these days, so wondering if there should even be a section with a handful of links we try to maintain at this point. Especially since the ones we have aren't that super-focused about Bitcoin anymore - there's a lot of other coverage intermixed.

What's the sentiment on potentially dropping this section from the Resources page?

wbnns:
Closes #2697.
Closes #2670.

### DIFF
--- a/_templates/resources.html
+++ b/_templates/resources.html
@@ -64,62 +64,6 @@ id: resources
       </div>
 
       <div class="card resources-card">
-        <img src="/img/icons/ico_spread.svg?{{site.time | date: '%s'}}" alt="Icon">
-        <h2 id="news">{% translate news %}</h2>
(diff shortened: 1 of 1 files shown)


mateusnds:
:(

PR #2731, opened 2018-10-30
Update ie.css

AdvaitT17:
Looks elegant...


wbnns:
@AdvaitT17 Hello, thanks for the suggestion. Let's leave the site-wide styling as-is for now.

If others disagree, however, and would like to advocate the change we can reconsider.

PR #2732, opened 2018-10-30
Update sans.css

AdvaitT17:
For headline


wbnns:
@AdvaitT17 Virtually all browsers would support a sans-serif, which is the current fallback, so this third fallback of Times New Roman most likely wouldn't ever get triggered.

PR #2743, opened 2018-11-08
Add Exchange

victor-act:
- Added AltCoin Trader to South African list.


wbnns:
@victor-act 

Hello, thank you for the submission, however, this site layout appears to be broken, with several elements becoming overlaid on one another when viewed on mobile.

Please let us know when this is fixed so we can resume the review.

PR #2745, opened 2018-11-09
Add fastx.io to Australian Exchanges list

rokanost:
Hey,

We have fixed the issues we had on the first pull request attempt.
Our pricing model has been updated as well. 

Fiat to crypto conversion rates include 1.75% Braintree cc processing fee.

Please let me know if any issues arise.

Thank you

### DIFF
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -388,6 +388,8 @@ id: exchanges
     <br>
     <a class="marketplace-link" href="https://digitalsurge.com.au/">Digital Surge</a>
     <br>
+    <a class="marketplace-link" href="https://fastx.io/">FastX</a>
+    <br>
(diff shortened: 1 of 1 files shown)


wbnns:
@rokanost 

Hello, when we sign up, we're unable to progress and see the following information in the top-right corner:

image

Hovering over this area reveals an active state indicating that a link should be present, however, inspecting the code reveals that there is only a stub (#).

Please let us know when this is resolved so that we can resume the review.

rokanost:
@wbnns 

There shouldn't be anything under that link.
(We might add a profile page sometime in the future, but it's not required at the moment).

To proceed with our system, you need to create a new conversion.
If you are buying/selling crypto, we will require your ID.
Also when purchasing, a random amount will be deducted from your credit card and you will have to confirm it.

If crypto to crypto, no account required. (if you do have an account already, it's fine to have KYC lvl 0).

PR #2752, opened 2018-11-14
Replace Simple Bitcoin Wallet with Bitcoin Lightning Wallet

btcontract:
BLW is a successor of my previous wallet and the one I plan to work on in the future. Simple Bitcoin Wallet is in support mode only so I'd like to have it replaced if possible.


Cobra-Bitcoin:
I don't think Lightning is mature enough yet to have a Lightning wallet promoted to users on bitcoin.org. Also, when Lightning is ready, I think we'll need to adjust our wallets page to differentiate between Lightning compatible or not wallets. Seems a bit _too_ bleeding edge for now.

btcontract:
OK, except that "when Lightning is ready" feels a bit like "when Bitcoin is ready", there will never be an official announcement.

crwatkins:
First, @btcontract, thanks for all of your new exciting work here! 

My initial thought was "Should we list lightning wallets along with the other wallets?"  We currently have no criteria regarding lighting network support, just like we have no criteria for BIP70 payments, nor many other features.  By this reasoning I believe, while not on the same scope as something like BIP70, supporting the lightning network in any capacity should not disqualify a wallet for listing with the others as long as it meets our criteria and the lightning usage is optional.

However, currently this wallet only supports bech32 addresses.   We don't currently have any criteria that specifically relate to address format, however I've always assumed an unwritten generic "interoperability with other wallets" criteria.  Since many current wallets do not yet support sending to bech32 addresses, I would recommend continuing to list Simple Bitcoin Wallet, and not yet list the new wallet until there is more support for bech32 or the new wallet also supports a legacy address format. @btcontract has [explained](https://github.com/btcontract/lnwallet/issues/39#issuecomment-383816188) that there are currently technical implementation limitations.   Might those be resolved soon?

In the meantime, have you considered a more qualified (less generic) name for the wallet?  We've caused confusion in the past dealing such generic naming.

wbnns:
@btcontract 

Hi, closing this for now since it has been a couple of months and we haven't heard back from you. We can potentially reopen when you reply to the [last message](https://github.com/bitcoin-dot-org/bitcoin.org/pull/2752#issuecomment-447039715) from @crwatkins.

btcontract:
Sorry, forgot to reply to this last time.

> @btcontract has [explained](https://github.com/btcontract/lnwallet/issues/39#issuecomment-383816188) that there are currently technical implementation limitations. Might those be resolved soon?

No, the plan is to be bech32 only while bitcoinj is used. Long term plan is to migrate to other (yet to be developed) bitcoin library but it's unclear whether P2WPKH-in-P2SH support will be needed at all by then.

> In the meantime, have you considered a more qualified (less generic) name for the wallet? We've caused confusion in the past dealing with such generic naming.

User consensus seems to be that the name is BLW, I can use that in description.

crwatkins:
@btcontract, since we've removed Simple Bitcoin Wallet per your request in #2847, could you open a new PR to list BLW and update the listing to include your current wallet naming?

I believe that all the issues that we were debating about listings when you originally submitted are now resolved and I see no major issues with listing.  I will provide a review shortly after you create a new PR.  Thanks!

PR #2760, opened 2018-11-23
Resolves IE positioning issue of cookie consent popup.

vim88:
On IE for desktop resolutions, the cookie consent pop-up is displayed on the left side of the screen with no space to the left:

ie

as opposed to being displayed on the right side of the screen, like on Chrome, Firefox, Opera:

chrome

By changing the value of the "left" property on that element from "initial" to "auto" it will be displayed correctly.

This is happening because IE does not support the "initial" value:
https://caniuse.com/css-initial-value

Also, "auto" **is** the initial default value for the "left" property.

### DIFF
--- a/js/cookieconsent/light-floating.css
+++ b/js/cookieconsent/light-floating.css
@@ -78,7 +78,7 @@
 
 @media screen and (min-width: 500px){
 	.cc_container{
-		left:initial;
+		left:auto;
 		right:20px;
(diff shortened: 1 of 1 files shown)


Cobra-Bitcoin:
LGTM.

wbnns:
@vim88 Thank you for reporting this and also for fixing it.

Issue #2768, opened 2018-11-30
Add Paybis.com as International bitcoin Exchange


kvpaybis:
Hi Bitcoin.org team,

Please add Paybis.com as International Bitcoin exchange. We offer customers to purchase various cryptocurrencies using Debit/Credit Cards, Bank Transfers in EUR/USD, Skrill, Neteller and many other payment methods. We are able to work with USA customers as well.

We are in service for more than 3 years and have an excellent track record: https://www.trustpilot.com/review/paybis.com

In order to support your great website, we would like to offer our affiliate programme with 30% revenue share, please let us know if you are interested.

Thank you in advance! 
Konstantins Vasilenko
COO, Paybis.com

odarboe:
@kvpaybis you should create a pull request adding the url of your exchange in this [file](https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/_templates/exchanges.html). Note they are grouped by continent.

kvpaybis:
Done!

wbnns:
@kvpaybis 

Thank you for the offer, however, that's not necessary in order to be listed on the website.

I replied to your PR (#2770) with some concerns.

Others also may have feedback that they would like to add.

PR #2770, opened 2018-12-03
Add Paybis.com to International exchanges list

umeku-paybis:
This pull request adds Paybis to the list of international exchanges.

Please see #2768 for details.


wbnns:
Hello, thanks for your submission, however, not sure this is a good recommendation for an international exchange; the fees are very high:

image

In the above example on the homepage, the minimum buy is $70 and the fee is just under $10 - almost a 14% fee for a person to get some bitcoin.

Also, the text is clipped off on the top of the page:

image

Before making a decision, let's place this on hold until more contributors can add feedback.

PR #2779, opened 2018-12-09
Transaction version is a signed integer.

Coding-Enthusiast:


harding:
It's a bit more complicated than that.  Although originally defined as a signed number, BIP68 casts it to unsigned for the purposes of comparison.  E.g.,

        bool fEnforceBIP68 = static_cast<uint32_t>(tx.nVersion) >= 2
                      && flags & LOCKTIME_VERIFY_SEQUENCE;

Although I don't believe this forces future changes to use unsigned tx version numbers, I do suspect it's going to be normative and so we're best off treating this field as an unsigned int.  However, I accept that the point is debatable and I'm certainly not opposed to providing more description about the nuances in this table (or elsewhere in the documentation).

Issue #2739, opened 2018-11-07
Internal links to developer guide broken


harding:
Reported here by @xekyo: https://twitter.com/murchandamus/status/1059898461036863488

This appears to be a result of #2700, but what surprised me was that PR passing the internal link checker which should've caught the broken anchor links on Travis.  However, it looks like commit 2b7882952abd63e777627fb889fe1b8abdf7a8c9 disabled checking any link that includes an anchor.

wbnns:
@harding Hey, thank you (and Xekyo).

harding:
@Cobra-Bitcoin to answer your question about why we structured links that way: in the part of the documentation where we provide the best description of a term, we gave it an anchor named `term-` and made that a link.  Making it an anchor allowed other people to link to it, and making it a link made it easy for someone to click on it and get the URL to copy/paste.

That UI was implemented before we had a glossary.  I think it'd be fine to keep the anchor so that glossary entries can direct back to the extended definition but remove the link so that users always go first to the glossary.

Note: I think you're using the term "anchor" differently than I am, so just to be clear `<span id="term-anchor">this is an anchor</span>` and `<a href="#term-anchor">this is a link</a>`.

harding:
Note that the revert in #2740 also broke links, e.g. I just saw [this broken link](https://bitcoin.org/en/operating-modes-guide#simplified-payment-verification-spv) in a comment posted yesterday to [one](https://bitcoin.stackexchange.com/a/34156) of my Bitcoin StackExchange answers.

wbnns:
@harding Thanks for mentioning. Looks like this user posted this comment, yesterday, prior to the revert, which is why the link became broken. Is that the issue you're referring to?

If so, the original PR was merged approximately 8 days ago - links shared to that specific page in that time frame, other than this one you mentioned.

wbnns:
Closing since #2786 ensures something like this would get caught in the future.

PR #2786, opened 2018-12-13
Restore test functionality

Cobra-Bitcoin:
Fixes a broken test to ensure that internal links with anchors are checked as described in https://github.com/bitcoin-dot-org/bitcoin.org/issues/2739.


wbnns:
@Cobra-Bitcoin 

Thanks. Just FYI, looks like this is failing on Travis CI.

Cobra-Bitcoin:
For some reason the tests are timing out and going over the 50 min runtime limit that Travis allows. It's most likely because there's too many links to check, or there's something with the tests making them stuck in an infinite loop. We'll hold off on merging until we can get more reasonable performance.

wbnns:
I spoke w/ Travis CI this week about the timeout issue, which is now resolved, however, now that the test is running successfully it has uncovered some broken same-page anchor links on translated pages which need to be fixed. Will use this branch for that as well.

wbnns:
Broken anchor links are now fixed and all tests are green. Merging.

PR #2811, opened 2019-01-14
Add Lunes Wallet

fmgfelipefmg:
Lunes Wallet is an open-source and progressive web application that supports Bitcoin, and also has multi-services built-in.

It's totally decentralized, only the users has access to their BIP39 seed words.

It has been available since January 2018, and we made a major update on September.


fmgfelipefmg:
I tried, tried and tried again. Still trying to understand what's happening and didn't find a solution.

Could someone help me? Please...

_wallets/lunes.md
```diff
@@ -0,0 +1,65 @@
...
+          environment: "checkpassenvironmenttwofactor"
+          privacy: "checkpassprivacybasic"
+          fees: "checkpassfeecontroldynamic"
+        privacycheck:
+          privacyaddressreuse: "checkpassprivacyaddressrotation"
+          privacydisclosure: "checkfailprivacydisclosureaccount"
+          privacynetwork: "checkfailprivacynetworknosupporttor"
+      - name: android
+        text: "walletlunes"
+        link: "https://play.google.com/store/apps/details?id=com.luneswallet&hl=en"
```
wbnns: The unescaped ampersand in the URL could be the issue.

wbnns:
@fmgfelipefmg 

Hello, took a quick look and left a comment, in-line.

crwatkins:
@fmgfelipefmg Thank you for your submission.

I've spent an hour or so using the wallet and found that it does not meet many of our [criteria](https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/docs/managing-wallets.md) and has some general difficulties of use.  Some of the items I have noticed (some are criteria, some are not) are

* While using the wallet in English, the confirmation emails and instructions were in Portuguese.
* Many of the screens and instructions were confusing to an experienced wallet user; perhaps some of the issues were with English translations; other languages were not reviewed.
* When the wallet is initially started, Bitcoin is not enabled by default.  Lunes are enabled, which is the name of the wallet, so users could be confused that Bitcoin addresses are being displayed.  Enabling Bitcoin in the UI was confusing.
* It was not clear where or when the seed was saved and how that related to the password or logout.
* When a seed was entered in iOS, the first character was automatically capitalized which made the seed invalid and the error message did not provide any clues.
* Received funds were never displayed in any way until confirmed on the blockchain.
* When received funds were eventually reflected in the balance, the associated transactions still did not appear in the transaction list (inconsistent wallet display). The transaction list displayed "Nothing found".
* Even after transactions were displayed, the wallet continued to display the same receive address (address reuse violation).
* The wallet does not show the number of confirmations on received transactions.
* Active items in the UI menu appear grayed out which was confusing.
* The wallet displays "from addresses" in transactions.
* When the balance is zero, a button appears labelled "Unavailable" which is confusing (it's the Send button).
* A space on the end of a Bitcoin address will cause the Send operation to fail and it is not obvious to the user what the problem is from the message.
* When sending, it is not obvious that the words "Low", "Medium" and "High" are fee options which must be selected.
* The wallet disallows sending of funds less than 0.0002 BTC.  I could not send transactions that I desired to send and I had to add more funds to the wallet to be allowed to send; not acceptable.
* Using the Max send option does not allow for any fee, so the wallet refuses to send.
* When an amount was chosen as the wallet balance minus the selected fee, the wallet claimed there were insufficient funds (I suspect rounding or truncation errors).
* When trying to send funds, I received messages such as a "Shipping error" had occurred and "An error ocurred [sic] in our server, try again later on"
* When send errors occur, the UI spinner spins forever afterwards.
* I never was successful at getting my funds out of this wallet.

There is a lot of work to be done on this wallet, so I can't recommend it for listing at this time.  I suggest some serious testing and verifying that it meets our listing criteria before resubmitting it in six to nine months.

wbnns:
@crwatkins Thank you.

fmgfelipefmg:
@crwatkins To be honest, we didn't expect that. It means that we still have a lot of work to do.

We already started to work on all of those points. We have a mission here, and we're going to accomplish.

You said you weren't able to get your funds out, if you still have that problem, please, tell me that I'll help. If you prefer, you can import your seedwords in other BIP38 compatible wallet.

Thanks very much for pointing out all those problems. See you soon, the next time you test our wallet, we'll make you very surprised. ;)

Thanks @wbnns too!

crwatkins:
@fmgfelipefmg 

> If you prefer, you can import your seedwords in other BIP38 compatible wallet.

I did that.  All set.  Thanks.

> Thanks very much for pointing out all those problems. See you soon, the next time you test our wallet, we'll make you very surprised. ;)

Excellent!

PR #2813, opened 2019-01-15
add XCH4NGE to p2p exchanges list

XCH4NGEUK:


wbnns:
@XCH4NGEUK 

Hello, we were not able to sign up and perform an initial review of the site. On the sign up page, the "I agree to the Terms of Use and Privacy Policy" box could not be checked in order to continue to progress through the sign up.

Trying to do so, caused a modal to pop-up with the terms of use / privacy policy. Scrolling through the copy and then closing the modal would prompt the box to become unchecked, where after, we would need to check the box again, causing the modal to reappear. This created an endless loop of trying to accept the terms of use, being prompted to read them, and then having to accept them again, followed by having to read them.

This issue was experienced in Firefox 64.0.2 (the latest release).

Please let us know once this is resolved so we can resume the review.

PR #2828, opened 2019-01-26
Added PagCripto to exchanges

Raizen:
Adding Brazilian exchange PagCripto.com.br.

### DIFF
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -311,6 +311,8 @@ id: exchanges
           <br>
           <a class="marketplace-link" href="https://www.omnitrade.io/">OmniTrade</a>
           <br>
+          <a class="marketplace-link" href="https://www.pagcripto.com.br/">PagCripto</a>
+          <br>
           <a class="marketplace-link" href="https://walltime.info/">Walltime</a>
         </p>
       </div>


wbnns:
@Raizen 

Hello, when reviewing this exchange we encountered several issues.

1. The 24-hour volume listed on the home page for the exchange appears to be very low (0.01216 BTC / ~$40 USD):

screen shot 2019-02-04 at 20 48 29

2. When logging in, the user is instructed to enter a 2FA code. First-time users who registered do not have a 2FA code to enter here, since setting up 2FA was not part of the registration process:

screen shot 2019-02-04 at 21 04 38

3. Ignoring the box allowed us to be able to log in, however, there are text overflow issues on the user dashboard, for example, with the drop-down menu:

screen shot 2019-02-04 at 21 04 15

4. Also with the text instructing users to set up 2FA:

screen shot 2019-02-04 at 20 58 59

5. Lastly, when withdrawing bitcoin, it was not clear if there was any validation on ensuring the user-provided address was in fact a valid bitcoin address to be able to withdraw to.

Raizen:
Hi,
Today was bad for our volume. In normal days, our volume are about 1.5-2 BTC.
In the second point, there are a warning in portuguese who tell the customer to just set if he has 2FA.
The text overflow are beign solved in the new version of dashboard.
About the validation when trying to withdraw, this check are made before the request was accepted. The system check if there is a valid bitcoin address and if yes, it send the transaction, or if not, it is rejected and the balance return to the user.

wbnns:
@Raizen Ok, thanks.

PR #2830, opened 2019-01-27
Update full-node.md

behradkhodayar:
A super-minor edit to make things accurate for those suffering from Obsession! MB replaced with MiB


wbnns:
@behradkhodayar 

Thanks! :)

PR #2831, opened 2019-01-28
Added BitKassa to list of exchanges in Europe

annetdeboer:
Add BitKassa to the list of exchanges in Europe

### DIFF
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -146,6 +146,8 @@ id: exchanges
         <br>
         <a class="marketplace-link" href="https://www.bitcoin.de/">Bitcoin.de</a>
         <br>
+        <a class="marketplace-link" href="https://www.bitkassa.nl">BitKassa</a>
+        <br>
         <a class="marketplace-link" href="https://www.bitpanda.com/">BitPanda</a>
         <br>
         <a class="marketplace-link" href="https://bl3p.eu/">BL3P</a>


wbnns:
@annetdeboer 

Hi, when we reviewing BitKassa we encountered the following issues:

1. The home page is in English, but the sign up process / subsequent screens are in Dutch. If this was added as-is to the "Europe" category, anyone who doesn't speak Dutch may have difficulty moving beyond the home page to sign up.

2. The account activation email is also in Dutch by default.

3. Once logged in a notice states that "buying bitcoins is only possible from Dutch bank accounts. If you wish to buy bitcoins but you don't have a Dutch bank account yet you may open one for free at the Dutch bank bunq (install their android/iphone app called 'bunq' and you're good to go)". This excludes everyone in the Europe category with the exception of people in The Netherlands.

Based on the above, we recommend you open a new PR to potentially be included in a new category in Europe for The Netherlands (similarly to other countries that appear like Poland and the United Kingdom).

Once submitted, we'll continue with the review.

annetdeboer:
Thanks for your valuable review @wbnns ; I will make adjustions etc.
One quick follow up on the Bunq bank.. Everone in Europe can potentially open a Dutch bank account with them by just installing the app. But I can see how this is not clear from the text. 
Anyway, BitKassa targets the Dutch market so a sub category definitely seems more appropriate. Will take a look at it later! Thanks again for your time.

wbnns:
@annetdeboer Ok, understood, thanks.

PR #2834, opened 2019-01-30
Community page housekeeping | http -> https

annetdeboer:
Checked all links were still working. Removed one url that now redirects to an ATM manufacturer.
Changed to https:// where possible.


wbnns:
@annetdeboer Perfect, thanks.

PR #2846, opened 2019-02-08
Tests: fix wallet schema test

harding:
As mentioned by @crwatkins on https://github.com/bitcoin-dot-org/bitcoin.org/pull/2840#issuecomment-460851781 , the schema-validation test failed to catch a problem.  This appears to have been because of how the test was modified to run as a child of the `find` process.  `find` always returns true if it finds files, so its child processes failures wouldn't be propagated back up to the test process.  This fixes the oversight.

I've tested this by reintroducing the exact problem identified by Craig above, running the test, and ensuring that it raised a fault.  Upon removing the bug, the test passes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -304,4 +304,4 @@ check-for-too-many-wallets-on-one-platform:
 
 check-validate-yaml:
 ## Validate YAML files against schemas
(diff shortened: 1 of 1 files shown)


crwatkins:
@harding Thanks so much for fixing a problem that wasn't even yours.  Your schema check saves me countless hours of time.

wbnns:
@harding Thanks!

PR #2855, opened 2019-02-13
wallets-filter: Remove conditional to hide current wallet from selector

### LINKED ISSUE #2844: Not possible to navigate to a wallet's other platforms
Once a platform (e.g. iOS) is chosen for a particular wallet, we have no way for the user to navigate to another platform for that same wallet (e.g. Android).

1. Select "Choose your wallet"
2. Select the iOS option for wallet BRD  (and see the iOS entry with iOS download links)
3. Select Android from the chooser below BRD

Results:  Selection of wallets change, but BRD is not displayed in the choices and thus there is no way to switch the display (scoring and links) to the Android version of BRD.

Note that we don't ever display the selected wallet in the choices below which leads to this problem.  I believe the selected wallet should always appear in the choices.

wbnns:
Resolves #2844 and will be merged once tests pass. Thank you @ap4y for helping prepare the patch.

Cc: @crwatkins

### DIFF
--- a/_includes/layout/base/wallets-filter.html
+++ b/_includes/layout/base/wallets-filter.html
@@ -60,9 +60,6 @@ http://opensource.org/licenses/MIT.
 <div class="wallet-list wallet-filter-list">
     <!-- All wallets -->
     {% for wallet in site.wallets %}
-    {% if page.id contains wallet.id %}
-     {% continue %}
-    {% endif %}
     {% assign platformList = "" %}
     {% for platform in wallet.platform %}
      {% capture platformList %}{{platformList}}{{platform.name}} {% endcapture %}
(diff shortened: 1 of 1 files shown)


wbnns:
We discovered an additional bug in the selector when testing this; placing this PR on hold until fixed.

wbnns:
Following up on this, the issue that was discovered in testing was that the exact same Desktop wallets are all available for Linux, Mac and Windows. This created the impression that the filter wasn't working when toggling between the three because nothing changed.

crwatkins:
Yeh, there used to be quite the differentiation between desktop operating systems, but no longer since most developers are using OS independent development platforms.  None of our currently listed wallets have any different scores for the different desktop operating systems so we could perhaps combine them at some point.

PR #2859, opened 2019-02-16
Change Wasabi transparency to checkgoodtransparencydeterministic

nopara73:
Wasabi wallet's transparency score was added with `checkpasstransparencyopensource`. This PR updates its to `checkgoodtransparencydeterministic`.

The reason for `checkpasstransparencyopensource` is that I was not familiar with the deterministic builds concept at the time, thus I just opened an issue to research it: https://github.com/zkSNACKs/WalletWasabi/issues/1078 

I've done my investigation now and it turns out, our tooling is creating deterministic builds by default. While I went ahead and verified it, of which the rest of this comment is about, I'd like to note that in the commit https://github.com/zkSNACKs/WalletWasabi/commit/b4c86c83b978ab0a8ea1fbf2354ec41447634d4d I explicitly specified the deterministic compiler flag to make it future proof and to avoid relying on the default behavior.

Roslyn (C# compiler) uses deterministic builds by default: [Why is deterministic builds the new MS norm now?](https://www.reddit.com/r/dotnet/comments/9pje6r/why_is_deterministic_builds_the_new_ms_norm_now/)

There's also a `<Deterministic>True</Deterministic>` compiler option we can add to our .csprojs.  
I went ahead to verify it. I ran the packager 5 times and extracted the Linux tar.gz archive as a folder (extracted because the diff output will look nicer):

1. Explicitly specified deterministic build in the .csproj: `deterministic`
2. Explicitly specified non-deterministic build in the .csproj: `non-deterministic`
3. Explicitly specified deterministic build (again) in the .csproj: `deterministic2`
4. Explicitly specified non-deterministic build (again) in the .csproj: `non-deterministic2`
5. Did not specify explicitly anything: `default`

After that I ran `git diff --no-index folder1 folder2` to spot the difference:

# ✅ default vs deterministic

image

# ✅ default vs deterministic2

image

# ❌ default vs non-deterministic

image

# ❌ non-deterministic vs non-deterministic2

image

# Reproduce

After [getting Wasabi source code,](https://github.com/zkSNACKs/WalletWasabi#build-from-source-code) one can reproduce our builds with the same flags we are publishing with (instead of running `dotnet run`): `dotnet publish --configuration Release --force --self-contained true --runtime linux-x64 /p:VersionPrefix=1.1.1 --disable-parallel --no-cache`.  

Source: 

https://github.com/zkSNACKs/WalletWasabi/blob/59c2180556eae2176dc43b25c3ba3a56ea69c281/WalletWasabi.Packager/Program.cs#L172

### DIFF
--- a/_wallets/wasabi.md
+++ b/_wallets/wasabi.md
@@ -18,7 +18,7 @@ platform:
       check:
         control: "checkgoodcontrolfull"
         validation: "checkfailvalidationcentralized"
-        transparency: "checkpasstransparencyopensource"
+        transparency: "checkgoodtransparencydeterministic"
         environment: "checkfailenvironmentdesktop"
         privacy: "checkgoodprivacyimproved"
         fees: "checkpassfeecontroldynamic"


crwatkins:
utACK.  Great!  

@nopara73 Have you tested this on Mac and Windows also?
Thanks for all the documentation and explanation.

nopara73:
> @nopara73 Have you tested this on Mac and Windows also?

No, but I do it now. I write my steps here as I go:

## Step 1: Clone and Publish

I cloned the project on Windows 10 and published for targets: `linux-x64`, `win7-x64`, `osx-x64`. Source: 
https://github.com/zkSNACKs/WalletWasabi/blob/59c2180556eae2176dc43b25c3ba3a56ea69c281/WalletWasabi.Packager/Program.cs#L54-L59

```sh
git clone https://github.com/zkSNACKs/WalletWasabi.git --recursive
cd WalletWasabi/WalletWasabi.Gui
dotnet publish --configuration Release --force --self-contained true --runtime linux-x64 /p:VersionPrefix=1.1.1 --disable-parallel --no-cache
dotnet publish --configuration Release --force --self-contained true --runtime win7-x64 /p:VersionPrefix=1.1.1 --disable-parallel --no-cache
dotnet publish --configuration Release --force --self-contained true --runtime osx-x64 /p:VersionPrefix=1.1.1 --disable-parallel --no-cache
```

Next, I repeated the exact same commands in Ubuntu 18.04 VM and OSX High Sierra VM.  
The binaries can be found in `WalletWasabi/WalletWasabi.Gui/bin/Release/netcoreapp2.2/{linux-x64/win7-x64/osx-x64}`. I renamed the Release folders respectively to operating systems: `ReleaseLinux`, `ReleaseWin`, `ReleaseOsx` and first compared with `git diff --no-index` the binaries generated on Windows and Ubuntu. Note that I found a difference here in `netcoreapp2.2/WalletWasabi.Gui.xml` files, but this is just a distraction, because the final binaries, as I specified before are in `netcoreapp2.2/{linux-x64/win7-x64/osx-x64}` and whatever outside of them doesn't get shipped.  
For the record this file is contains comments and is generated so if we reference a project in another project, then we can read the comments in Visual Studio by hovering above a function or a variable. But again it's not important I'm not even sure why I'm expanding upon it here, I guess I just want to show my thinking process. Anyway I added a remove command to the above workflow and repeated everything:

- Linux & OSX & Windows PowerShell: `rm bin/Release/netcoreapp2.2/WalletWasabi.Gui.xml`
- Windows cmd: `del bin\Release\netcoreapp2.2\WalletWasabi.Gui.xml`

## Step 2: Compare

I built in all OS for all OS and compared them with `git diff --no-index`. This time I encountered differences, and @molnard and me were hunting them down all day, but it was unsuccessful. One was about different permissions after copying from Windows to Linux, which I believe is solvable somehow, but the other one was with line endings inside the binaries, which was quite strange. I'll get back to this some time in the future.

# Summary

If one builds on Linux for Windows, the binary will be different than building on Windows to Wondows, mainly due to because of permissions and line endings, but there may be more.

crwatkins:
Hey @nopara73 don't make this too hard on yourself!  All we as a community need to be able to do is reproduce your build and create an identical binary to the one on your distribution page.  You can specify the build environment.  You don't have to make it work on other environments just for us.

nopara73:
I reopened this PR, because I created a [guide on how to reproduce the builds.](https://github.com/zkSNACKs/WalletWasabi/blob/master/WalletWasabi.Documentation/Guides/DeterministicBuildGuide.md)

However I coded an `--onlybinaries` switch to our packager project to make it easy to reproduce our builds, which was of course not present at the time of our last release, so this guide won't work exactly as described until our new release, thus I changed the status of this PR to WIP and will change it back after we release.  

# Guide

> Reproducible [or deterministic] builds are a set of software development practices that create an independently-verifiable path from source to binary code.- https://reproducible-builds.org/

This guide describes how to reproduce Wasabi's builds. If you got stuck with these instructions, take a look at how to build Wasabi from source code: https://github.com/zkSNACKs/WalletWasabi#build-from-source-code

# 1: Assuert Correct Environment

In order to reproduce Wasabi's builds you need a Git, a Windows 10 and the version of .NET Core SDK that was the most recent in the time of building the release.

# 2. Reproduce Builds

```sh
git clone https://github.com/zkSNACKs/WalletWasabi.git --recursive
git checkout {hash of the release} // This works from 1.1.2 release, https://github.com/zkSNACKs/WalletWasabi/releases
cd WalletWasabi/WalletWasabi.Packager/
dotnet restore
dotnet build
dotnet run --onlybinaries
```

This will build our binaries for Windows, OSX and Linux from source code and open them in a file explorer for you.



# 3. Verify Builds

You can compare our binaries with the downloads we have on the website: https://wasabiwallet.io/
In order to end-to-end verify all the downloaded packages you need a Windows, a Linux and an OSX machine.



## Windows

After you installed Wasabi from the `.msi`, it'll be in `C:\Program Files\WasabiWallet` folder. You can compare it with your build:

```sh
git diff --no-index win7-x64 "C:\Program Files\WasabiWallet"
```

## Linux && OSX

You can use the Windows Subsystem for Linux to verify all the packages in one go. At the time of writing this guide we provide a `.tar.gz` and a `.deb` package for Linux and .dmg for OSX.  
Install the `.deb` package and extract the `tar.gz` and `.dmg` packages, then compare them with your build.

After installing WSL, just type `wsl` in explorer where your downloaded and built packages are located:



### .deb

```sh
sudo dpkg -i Wasabi-1.1.1.deb
git diff --no-index linux-x64/ /usr/local/bin/wasabiwallet/
```

### .tar.gz

```sh
tar -pxzf WasabiLinux-1.1.1.tar.gz
git diff --no-index linux-x64/ WasabiLinux-1.1.1
```

### .dmg

You'll need to install `7z` (or something else) to extract the `.dmg`: `sudo apt install p7zip-full`

```sh
7z x Wasabi-1.1.1.dmg -oWasabiOsx
git diff --no-index osx-x64/ WasabiOsx/Wasabi\ Wallet.App/Contents/MacOS/
```

crwatkins:
ACK 602dd15.  Let us know when you release.  Nice job!

nopara73:
This is not WIP anymore. @NicolasDorier @molnard would you be interested going through the [Deterministic Build Guide](https://github.com/zkSNACKs/WalletWasabi/blob/master/WalletWasabi.Documentation/Guides/DeterministicBuildGuide.md) to make sure everything works? (Make sure you have `dotnet --version` -> `2.2.105` (The latest .NET Core.))

crwatkins:
Nice work!  We were just waiting for your release.

LGTM 602dd15 for merge.

NicolasDorier:
@nopara73 I think for deterministic build you should just build Wasabi via docker so you can lock down the build environment.

nopara73:
I think using less dependencies for security is more important than the simplicity of deterministic builds, but I agree it's debatable. That being said, the blocking factor here is not really this design decision, but the fact that there's no OSX docker image and I want to do OSX packaging on OSX in the future, because there are a bunch of proprietary OSX build software we either had to hack around or couldn't do at all, like codesigning to avoid Apple Gatekeeper warnings or releasing `.pkg` instead of `.dmg`, so we wouldn't be able to use docker anyway.

NicolasDorier:
Docker is not a dependency, docker is an executable documentation for deterministic build.

PR #2871, opened 2019-02-22
Add Cryptoradar to Directories

caloon:
Cryptoradar is a popular tool that lists more than 30 international Bitcoin exchanges and allows users to compare these platforms based on a number of factors such as prices, fees, features and user reviews.


wbnns:
@caloon 

Hello, thank you for the submission. This (Cryptoradar) seems to contain a significant amount of similar material that another resource in this section, "Buy Bitcoin Worldwide" already provides. 

The latter also provides a helpful guide to assist new users buying bitcoin, which the former does not. Combine this with exposure to almost a dozen other altcoins via the top of page filter, we are hesitant to add this to the site at this time.

Closing this issue for now, however, if other contributors disagree we can reopen it for further discussion.

PR #2874, opened 2019-02-26
Fix number of allowed Inv. message type identifier

Coding-Enthusiast:
There are already 7 type identifier defined not 3.


wbnns:
@Coding-Enthusiast Thank you.

PR #2876, opened 2019-03-01
Fix Wrong Language Name

AlexKoala:
'한국의' means 'of Korea'.
The correct translate of 'Korean' is '한국어'.


3ggerman29:
btc add  bitcoincash:qprtuvakg7na6qdjf5alypgjrvcm6qj6nv0dwnpvsx

wbnns:
@AlexKoala 

Thank you for the suggested change. We're just waiting on one of our translators that specializes in Korean to review and concur with your recommendation.

PR #2886, opened 2019-03-07
Exchanges Request AlteumX

AUMXcquintana:
Exchange Request. 
Add into Mexico AlteumX Exchange. 
https://alteumx.com

### DIFF
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -251,6 +251,8 @@ id: exchanges
           <a class="marketplace-link" href="https://bitso.com/">Bitso</a>
           <br>
           <a class="marketplace-link" href="https://www.volabit.com/">Volabit</a>
+          <br>
+          <a class="marketplace-link" href="https://alteumx.com/">AlteumX</a>
         </p>
       </div>
     </div>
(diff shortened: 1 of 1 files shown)


_templates/exchanges.html
```diff
@@ -251,6 +251,8 @@ <h3 id="mexico" class="no_toc">Mexico</h3>
           <a class="marketplace-link" href="https://bitso.com/">Bitso</a>
           <br>
           <a class="marketplace-link" href="https://www.volabit.com/">Volabit</a>
+          <br>
```
wbnns: Exchanges should be added in alphanumeric order.

AUMXcquintana: Hello, thanks for your comment. We committed a change where the alphanumeric order is applied.

wbnns:
@AUMXcquintana 

Thank you for revising the PR. Unfortunately, we were unable to continue reviewing the exchange. When signing up a red "ERROR" button appeared when using the REGISTER button.

Trying to register again led to another error message stating that the user we had created already existed, with an option to resend an activation email (which we had never received previously). When entering in the email address to resend the activation email (to the email address that had been used to sign up that was showing the duplicate user error message), pressing the "RESEND" button to send the email caused nothing to happen. Examining the developer console yielded the following error:

image

Please let us know when these issues are resolved and we will resume the review of the exchange.

PR #2888, opened 2019-03-11
Changed verbal tense to reflect currently provided consensus sharing from v 0.10.

behradkhodayar:
As we are on 0.17 now, text updated to represent what has been already developed rather than tasks going to be developed


wbnns:
@behradkhodayar Thank you.

PR #2895, opened 2019-03-13
Update laanwj gpg keys

laanwj:
The expiry date was bumped, and the key is expired by now, and some people are running into issues due to this (https://github.com/bitcoin/bitcoin/issues/15592) so update the key to the most recent one.


wbnns:
@laanwj Thanks.

PR #2903, opened 2019-03-23
Corrected data message type identifiers

Coding-Enthusiast:
The type values with witness are set by combining existing type identifiers with a bit mask called `MSG_WITNESS_FLAG` in [bitcoin core](https://github.com/bitcoin/bitcoin/blob/3f125151998d9fead198fd44243dd64006b5a56b/src/protocol.h#L379-L381) instead of being separate values of 5, 6, 7.


wbnns:
@Coding-Enthusiast Thank you.

This is scheduled to be merged on Wednesday, April 24th.

PR #2912, opened 2019-04-04
Remove incorrect explanation of `verifymessage`

real-or-random:
It's not true that all four possible public keys (corresponding to recovery
ids) are tried. The recovery id is instead read from the signature, see
https://github.com/bitcoin/bitcoin/blob/78dae8caccd82cfbfd76557f1fb7d7557c7b5edb/src/pubkey.cpp#L189.


wbnns:
This is scheduled to be merged on Friday, April 12th.

PR #2901, opened 2019-03-20
update ios and android apps for GreenAddress rebranded to Blockstream…

greenaddress:
… Green

Hi, I hope I did all the updates correctly, if not, please let me know and I'll update asap.

I removed the GreenAddress iOS app as it has been replaced by Blockstream Green for iOS.
I also removed GreenBits (for Android) as likewise it has been replaced by Blockstream Green for Android.

We compressed the new screenshot and logo, let me know if anything is missing.

PR #2916, opened 2019-04-06
Temporarily remove GreenAddress until further review

Cobra-Bitcoin:
This is following on from a discussion in https://github.com/bitcoin-dot-org/bitcoin.org/pull/2901 and removes the GreenAddress and Greenbits wallets temporarily until further review and enough time has elapsed as the current listing is pointing users to the new versions.

The new version of these wallets is hosted on a new repository, uses a new codebase, and is also rebranded. All these changes at once makes it closer to a new wallet and product, not a slight incremental upgrade of an existing one. We can consider re-listing when the wallet is shown to function fine with no user issues for a few months and continues to meet the existing listing criteria.

Additionally, we might need new criteria to prevent existing wallets from being "acquired" in this way. If a wallet has a sudden substantial code/repo change along with a rebrand, our policy should be to remove it and make it go through the listing process again after a short period.


theymos:
I haven't looked into whether the recent changes in GreenAddress are substantial enough to make it essentially a new wallet, but note that GreenAddress was actually acquired by Blockstream in 2016, not recently. So if it's *merely* a rebrand, then the people behind it hasn't changed.

crwatkins:
I agree that some changes need to be made here.  For example our existing listings point to an iOS app that is no longer in the app store.  However, the desktop versions of GreenAddress are still the same apps that we've always listed.  I don't believe that there is any reason to remove them.

Cobra-Bitcoin:
@crwatkins Yeah I noticed the iOS link was dead, and I also saw the Greenbits pointing to the new version. Good catch on the desktop versions of Greenaddress still being the same, though I'm not sure if they'll update that too.

How about we remove the broken iOS link on Greenaddress and also remove the Greenbits listing since it's pointing to the new version, but we keep the desktop versions of Greenaddress?

crwatkins:
@Cobra-Bitcoin That sounds reasonable to me.  Input from @greenaddress welcome!

Cobra-Bitcoin:
@crwatkins I've made the necessary changes!

I recommend we merge as quickly as we can since the present situation is causing dead links on the site.

crwatkins:
LGTM.

greenaddress:
Note that the android app GreenBits now Blockstream Green wasn't rewritten from scratch - it was an update to the previous app with rebranding even if it has a new repo and brand new UX. The repos can be diffed (and I regret we didn't just transfer/renamed the old repo and kept history)

As for removing iOS and keeping desktop I don't have additional input/comments.

anonymous108:
Please forgive me GreenAddress,

I am confused as to why you choose to make a new repo. Can you create a video that goes through the diff in detail? Or have your wallet code been audited and by how many verified people in the space.

I see that the app is removed from the site bitcoin.org site, this is wise. I am keeping the old version. Will not update until I am sure that there is nothing undercover in the new version.

Just becuase something is open-source, that is no guarantee of safety. It is unfortunate that you chose a new repo. Makes checking more "diff" icult for most people (I hope you "app"reciate my pun)

crwatkins:
The new app will be reviewed soon.

PR #2920, opened 2019-04-10
Added Bitupper Explorer to Resources Page

andrewzhuk:


wbnns:
@andrewzhuk 

Hello, thank you for the suggested addition. We already have three explorers on the resources page. Not sure that adding a fourth is warranted at this time, so closing for now. If other contributors disagree, however, we can reopen for further discussion.

Also, as a side note, it looks like the logo on the homepage above the search field is slightly distorted:
image

It might be because it isn't retina-friendly. The current dimensions of the image are 240 x 66. You could try replacing it with an image that is 480 x 132 or an SVG to resolve the issue.

PR #2930, opened 2019-04-17
Remove Opendime from resources under "Vouchers" category

Cobra-Bitcoin:
A user in the Bitcoin Core community Slack (shinobimonkey) felt that we were miscategorizing Opendime by listing it as a form of voucher.

This pull request removes Opendime from that page until we find a better way to explain the concept to users.

### DIFF
--- a/_templates/resources.html
+++ b/_templates/resources.html
@@ -121,9 +121,6 @@ id: resources
         <p>
           <a href="https://gyft.com/bitcoin/">Gyft</a>
         </p>
-        <p>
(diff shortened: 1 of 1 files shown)


luke-jr:
IMO "voucher" is a reasonable classification.

wbnns:
@Cobra-Bitcoin and I chatted about this and were thinking perhaps it's best to just close and leave as-is for now, unless additional feedback from others is received that the current categorization/location isn't accurate.

PR #2833, opened 2019-01-29
Added Binance exchanges to list

tsbd:
Added Binance.com, Uganda, and Jersey.

PR #2934, opened 2019-04-19
Add Binance Exchange - Revised w/ new PR

JagerGizmo:
I have revised the comments in PR #2833 and fixed them in this PR along with an extra addition


wbnns:
@JagerGizmo 

Hello, thank you for the quick submission, however, @tsbd has already made several of these changes. Also, #2833 has been updated with additional requested changes - let's give that PR priority since it was opened first.

Lastly, not sure that the additional change in this PR of adding Binance Academy to the resources section on bitcoin.org would be a relevant inclusion alongside the other resources. There's no mention of bitcoin anywhere on the academy homepage; it's more broadly focused on blockchain technology and cryptocurrencies in general:

Screen Shot 2019-04-19 at 15 25 45

If other contributors disagree and think that's OK, we can reconsider the addition.

PR #2939, opened 2019-04-22
Update es.yml

alvarodh5:
Fix translation issue


wbnns:
@alvarodh5 Thank you.

PR #2949, opened 2019-05-02
Added Bitcoin UK & Bitcoin CA

andrewbtc:
Adding bitcoin.co.uk and bitcoin.ca


wbnns:
@andrewbtc 

Hello, in reviewing these two websites we see that it doesn't appear that they're exchanges by definition for the purposes of being present on the Exchanges page, in part since there is no way to sell bitcoin if needed in order to convert back to fiat.

PR #2955, opened 2019-05-05
correct GET getutxos output format description

0xB10C:
Location: https://bitcoin.org/en/developer-reference#get-getutxos

The current description seems to be copied from `GET Block`. Changed from 'block' to 'UTXO set' in the output format description.


harding:
Thanks for catching this!  I'd suggest `s/set content/entries/` for both changes as `UTXO set` sounds like you're getting the whole multi-GB set.  (Alternatively, maybe `s/set/subset/` ?)

0xB10C:
> s/set content/entries/

Thanks for the feedback! Sounds way better.

harding:
Untested ACK 3407bf6.  Thanks!

wbnns:
@0xB10C @harding Thank you.

This is scheduled to be merged on Wednesday, May 8.

Issue #2861, opened 2019-02-17
Need way to compare wallets


crwatkins:
We present a fairly good selection of wallets to choose from, but little guidance on the process of selecting a wallet. I suspect most users play [whac-a-mole](https://en.wikipedia.org/wiki/Whac-A-Mole) on the wallet listings going from wallet to wallet trying to differentiate them by reading our fairly accurate descriptions ("scoring").  As I (and others) have written in other issues and PRs I would like improve this situation in three stages:
1.  Expose existing wallet scoring information in a format that is easier make comparisons
2. Add a "Good for new users" badge for wallets that are best for first time users
3. Add a feature/capability chart for wallets (e.g. multisig, segwit, bech32, lightning, etc.)

I've opened this issue to discuss the first item; the other two will be the subject of subsequent issues and PRs.  We've already had a fair amount of discussion in another [issue](https://github.com/bitcoin-dot-org/bitcoin.org/issues/2347) that I originally opened almost a year ago to discuss specific bugs related to the wallet page redesign.  I would like to move that discussion here to focus on this one issue with a title that is much more obvious to interested parties.

**Background on existing scoring**

Wallets are "scored" on six categories, one of which is privacy. Privacy has three subcategories which are similarly scored. The privacy score is manually (not in code) determined by the algorithm on the [wallet criteria page](https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/docs/managing-wallets.md#score) based on the three privacy subcategory scores. Scores are not numbers, but rather are text labels which refer to two or more text descriptions of that category. Wallets are scored by choosing the text (and thus the label) that most accurately describes the wallet. All labels begin with the characters "check" followed by one of the four following sets of characters:
* good
* pass
* fail
* neutral

Currently on bitcoin.org, we indicate "fail" with an orange ball beside the text explaining the score while the other three ("good", "pass", and "neutral") are indicated with a green ball. 

**Proposal**

My proposal is that we expose our existing scoring in a way that is easy for users to consume and compare.  Except as I explained above, we don't currently expose the good/pass/fail/neutral levels.  I further propose that we expose this visually.

**Solicitation**

I'm looking for ideas on how to be express this.   Previously, @harding has suggested using [Harvey Balls](https://en.wikipedia.org/wiki/Harvey_Balls) to augment potential color presentation of these levels, for example

```
              | Control | Validation | Transparency | Environment | Privacy | Fees
[...]

PR #2957, opened 2019-05-05
wallets: Drop Windows for Mobile

wbnns:
This drops Windows for Mobile from the wallets page [as discussed](https://github.com/bitcoin-dot-org/bitcoin.org/issues/2861#issuecomment-484719171) in #2861. Microsoft will [no longer be supporting](https://support.microsoft.com/en-us/help/4485197/windows-10-mobile-end-of-support-faq) their Windows Mobile operating system and is recommending that users switch to Android or iOS:

image

Cc: @crwatkins


crwatkins:
utACK.   Thanks @wbnns 

One thing that I noticed is it that currently on desktop and iOS Chrome, BitPay is not displayed when filters are used to select **Windows Phone**.  In the full listing BitPay shows **Windows Phone** support, just not during filtering.  Perhaps we should look into where this bug lies before merging this to make sure that it isn't also affecting entries we are not removing.

wbnns:
@crwatkins

Thanks. With regard to:

>In the full listing BitPay shows Windows Phone support, just not during filtering. Perhaps we should look into where this bug lies before merging this to make sure that it isn't also affecting entries we are not removing.

The issue was that `windowsphone` was omitted from the `compat` field in `bitpay.md`, which is why it wasn't showing when filtered. If you noticed that Coin.Space, another wallet, did show when filtered, it was because `windowsphone` was present in that field.

image

This is scheduled to be merged on Wednesday, May 8.

Issue #2968, opened 2019-05-11
$1.2M Traded on Quebex.com!


davidquebex:
Dear bitcoin community,

Quebex.com P2P exchange is now at $1.2M traded since 2016. Last trade was made by MarcosRodrigues for $3,000.

Are there any contributors that would like to add Quebex.com to bitcoin.org?

Benefits include:
* Instant verification (< 5 minutes)
* No Photo ID required
* No deposit fees
* No fees for trading CAD
* No CAD deposit withholding time

Quebex is a federally registered MSB with FINTRAC Registration #M17947675 and licensed with AMF #903510.

We offer email support, live chat and phone support at (888) 661-3358.

Accepted payment methods include: Email Money Transfers, Bank Deposits and PayPal.

aaronkuchma:
As a contributor (of this PR https://github.com/bitcoin-dot-org/bitcoin.org/pull/2977 ), I agree that Quebex is great and should be reviewed.

achow101:
Bitcoin.org's issues are not the place to advertise your service.

davidquebex:
Andrew, this is not advertisement. We are just  giving facts as to why Quebex.com should be recommended to the Canadian bitcoin community. Thanks! David

wbnns:
Duplicate of #2960.

@davidquebex @aaronkuchma Please stop spamming the repository. Other contributors and people following along can see all of your messages in issues and pull requests, regardless of if they are open or closed, and can make their own informed decisions and recommendations if they would like.

davidquebex:
Hi Will,

Ok sure. Was not aware of that.

I have a feeling you are the boss on bitcoin.org. :)

Why are you closing the issue? You didn’t even reply?

Now that you have seen all the scam and fraud complaints, will you be removing the scam sites?

Cheers!

David

davidquebex:
Well, at least you fixed the space in the documentation :)

Issue #2967, opened 2019-05-10
Quebex.com P2P exchange eliminates e-transfer fraud


davidquebex:
Dear Bitcoin community,

Quebex.com is working really hard to create a safe P2P environment for both sellers and buyers.

Many P2P exchange like localbitcoins have huge issues with e-transfer fraud resulting in sellers loosing bitcoins and their bank account.

Just google “localbitcoin e-transfer fraud”

Quebex.com has made it very hard for fraudsters to signup and get their accounts approved.

Since 2016, we have had over $1,194,000 CAD in P2P trades on Quebex without one single instance of Fraud.

So now traders can focus on selling bitcoins instead of wasting time and worrying about asking buyers to provide selfies and photo IDs. This is most of the time useless as these fraudsters hide behind burner phones and proxy’s and provide falsified ID.

The proof that we have a good system in place is that Will Binns has tried to signup many times with burner phones and proxy services but can’t get access to the site features.

This is because it’s typical behaviour of fraudsters.

If he tries to signup and provide real info the system will automatically approve him and he can start trading in less than 5 minutes, without even uploading any IDs.

I hope the bitcoin community will find a P2P exchange that cares about its sellers useful and hopefully bitcoin.org will also agree.

Will Binns, please try to create a real account and not hide as you will never have access to site feature and will always encounter issues.

Best,

David

wbnns:
@davidquebex Hello, sorry, but we disagree.

In case any other contributors would like to review your exchange and potentially recommend it for listing, here are links to other related issues:

+ #2421 
+ #2559
+ #2960

In addition to the above, we noted we weren't the only ones who have [experienced issues](https://www.reddit.com/r/BitcoinCA/comments/7k2w8x/has_anyone_here_used_quebex_to_buy_or_sell_crypto/) trying to use the platform.

davidquebex:
Hi Will,

Thank you for your quick reply!

I respect that you disagree, but you must understand that as long as you try to hide when signing up you will experience issues.

Please understand that protecting the traders on Quebex.com is our number one priority.

Quebex has over $1M traded since 2016, without e-transfer fraud and think that the bitcoin community would really appreciate a service that can provide a safe P2P trading environment.

As per reddit post, these are very old posts. Since all issues have been resolved. We don’t use webcam verification anymore, there’s no ID required to get verified and support has been greatly improved.

[...]

Issue #2976, opened 2019-05-12
Remove bitbuy for signup issues?


aaronkuchma:
As per the bitbuy.ca website, Trading is "[restricted] to Canadian clients only with Canadian IPs," (source: https://bitbuy.ca/en/compliance ).

This makes it difficult for me to review it with a non-Canadian IP as the site will not allow non-Canadians to transact.

As such, I don't think it is a good fit for bitcoin.org.

achow101:
NACK. It's in the Canadian section for a reason..

PR #2978, opened 2019-05-12
Add Quebex

aaronkuchma:
Add the Quebex bitcoin exchange to the list of exchanges in Canada.

See https://github.com/bitcoin-dot-org/bitcoin.org/issues/2968 and https://github.com/bitcoin-dot-org/bitcoin.org/issues/2967

### DIFF
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -236,6 +236,8 @@ id: exchanges
       <div>
         <h3 id="canada" class="no_toc">Canada</h3>
         <p>
+          <a class="marketplace-link" href="https://quebex.com/">Quebex</a>
+          <br>
           <a class="marketplace-link" href="https://bitbuy.ca/">Bitbuy</a>
           <br>
           <a class="marketplace-link" href="https://www.canadianbitcoins.com/">Canadian Bitcoins</a>


maxwellmons:
NACK

aaronkuchma:
I see an objection, but I don't see an accompanying explanation for it.

The notion of decentralization is an acknowledgement that only anarchy can ever exist. I understand this. But the ethos of Bitcoin's decentralization wasn't "early adopters should have unlimited power and yield it in such a way as to illustrate the ills perpetrated by fiat incumbents". 

Correct me if I'm wrong (don't just plaster NACK like it were Elon's face and this PR was your meme canvas), but Bitcoin was supposed to protect people from being debanked, from being scammed by a central authority, and from being marginalized and subjected to selectively enforced rules.

It appears that bitcoin.org plays a role in the Bitcoin ecosystem like media outlets or banks play to fiat. You don't control bitcoin. You are a "last mile" service provider.

It also seems that bitcoin.org's actions may have led to Bitcoin adopters being scammed (we have tried to assist by pointing out some of the scams being recommended). A community policing approach was probably part of Satoshi's vision, even if that vision involved Philosopher Kings doing that policing.

Quebex has tried to follow the undocumented rules for inclusion. We have reviewed the other exchanges listed in an attempt to create a deterministic algorithm that defines what should and should not be included (e.g. https://github.com/bitcoin-dot-org/bitcoin.org/issues/2976 was a way of eliminating the possibility that exchanges can't be listed if there are "signup issues when using suspicious information that violates the exchange's TOS"). Given the opposition without clear guidance on what would be required to be listed, it certainly seems like this is a string rules for some, no rules for other situation.

This seems more like an ETH Classic situation than a Bitcoin situation. "bitcoin.org was available for the taking; now we are the Borg, and we do not chose you for assimilation".

Kind regards,

- Aaron

achow101:
There are already many issues that have been opened to add Quebex, no need to open yet another one.

aaronkuchma:
This is a pull request though.

PR #2983, opened 2019-05-16
download: Replace ppa with snap

MarcoFalke:
See https://github.com/bitcoin-core/bitcoincore.org/pull/639


Cobra-Bitcoin:
We should keep linking to the PPA so long as @TheBlueMatt continues to update it. If he eventually doesn't want to do it, then he should pass the responsibility onto someone else.

I don't see the need to remove ways of downloading the software easily, it will only make it harder for users to run and keep their full nodes up to date.

MarcoFalke:
Of course, it will be updated on versions of Ubuntu where it has previously been released/installed. However, there is no need to force every Ubuntu user into downloading the ppa by default. Just to give some reasons why the ppa is broken:

* In addition to trusting Ubuntu, you have to trust Launchpad/the ppa owner
* The binaries itself are not deterministically built, nor signed by our release process
* The binaries aren't tested as part of the release process
* Building for all Ubuntu versions often takes days to get it built with all distro-provided dependencies (and it has to be supported for 10 years at minimum, see Xenial EOL cycle)
* Some users attempt(ed) to add the ppa to a debian host, which will obviously lead to issues
* There is no way to opt out of updates
* ...

I am aware that some of the issues can be solved by ditching the ppa build and replacing it with a simple downloader that fetches the gitian binaries, but no work has been done to achieve that and I doubt it will ever happen.

Also, there is a bitcoin-core snap package (that uses the gitian binaries), so while not perfect, it solves some issues.

Cobra-Bitcoin:
I understand all of your concerns but if someone is doing all that work _anyway_ to get the PPA working, it seems strange to me to not have it at least as an option.

How about we have the Ubuntu button in the download page link to both the Snap and PPA? Also with a warning icon next to the link to the PPA saying  "This is not a secure way to download Bitcoin Core and may not be reliable in the future" when the user hovers over it. I think thats better than outright removing it.

MarcoFalke:
> I understand all of your concerns but if someone is doing all that work anyway to get the PPA working, it seems strange to me to not have it at least as an option.

This reasoning is circular. It needs to be supported on released platforms, because it has been listed in the past. But because it is supported shouldn't mean that it should be listed.

Also be reminded that `add-apt-repository ppa:bitcoin/bitcoin` *fails* on the two latest Ubuntu versions (disco, eon). So all this has as an effect is to corrupt the users apt config file and print an error on each `apt update`. See

```
apt update
Ign:1 http://ppa.launchpad.net/bitcoin/bitcoin/ubuntu disco InRelease
Hit:2 http://archive.ubuntu.com/ubuntu disco InRelease
Hit:3 http://security.ubuntu.com/ubuntu disco-security InRelease
Hit:4 http://archive.ubuntu.com/ubuntu disco-updates InRelease
Hit:5 http://archive.ubuntu.com/ubuntu disco-backports InRelease
Err:6 http://ppa.launchpad.net/bitcoin/bitcoin/ubuntu disco Release
  404  Not Found [IP: 91.189.95.83 80]
Reading package lists... Done
E: The repository 'http://ppa.launchpad.net/bitcoin/bitcoin/ubuntu disco Release' does not have a Release file.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.

wbnns:
@MarcoFalke Thanks.

Based on the conversation over at https://github.com/bitcoin-core/bitcoincore.org/pull/635, it seemed like this was kind of on hold / in limbo pending results on the outcome from conversations between Canonical and @TheBlueMatt - until a subsequent PR removed the PPA option. Since he originally opened the issue and also maintains the PPA, would be nice to get his input, if possible. 18.04 LTS  users (the current LTS) can download via the PPA, but yes, it's a problem if anyone using the latest version (19.04) is going to encounter errors, especially since it is the latest version being promoted on the download page of the Ubuntu website:

image

If support for that isn't going to be updated on Launchpad then perhaps we could see if Matt is open to letting others who are trusted help maintain it. Otherwise, unless there's reason to believe that the PPA is at risk of being compromised or something like that, would agree with @Cobra-Bitcoin that this only makes it harder for people to get Bitcoin up and running on their systems. There are millions of Ubuntu users, many of whom could become (and already are) interested in running Bitcoin but don't know how to compile from source.

MarcoFalke:
The conclusion of the phone call was that the `bitcoin` snap package will be removed and replaced by the `bitcoin-core` snap package. See https://github.com/bitcoin-core/packaging/pull/13

harding:
@wbnns 

> There are millions of Ubuntu users, many of whom could become (and already are) interested in running Bitcoin but don't know how to compile from source.

Compiling from source is unnecessary.  The Bitcoin.org download page provides Linux binaries that work on Ubuntu.  There's also the Ubuntu Snap package that will turn up in any search an Ubuntu does for "bitcoin" from Ubuntu's built-in software center thing.

TheBlueMatt:
As others stated, for Ubuntu users we'd prefer they migrate to the snap which should have a better management story going forward. It should also be as easy, or easier, to install, so no reason to maintain both, really. The PPA will be maintained for existing platforms but to avoid duplicated effort we'd like to deprecate it, and removing installation instructions is the first step.

Cobra-Bitcoin:
Maybe a more worthwhile discussion is replacing the PPA with Snap rather than removing the entire Ubuntu button/support entirely as this PR does?

MarcoFalke:
Done

wbnns:
@MarcoFalke Awesome, thank you.

Tested, LGTM:

image

wbnns:
This is scheduled to be merged on Wednesday, May 29th.

PR #2984, opened 2019-05-17
added Dasset to the list of NZ exchanges

phoenix-dasset:


wbnns:
@phoenix-dasset 

Hello, when reviewing the exchange, we tried to sign up but were unable to do so. We encountered the same issue as before that is noted in #2498.

sm-dasset:
We've resolved this issue and have sent a password reset request to the email address that was used to sign up using the front-end interface. You can also make a reset password request at https://dassetx.com/problem-logging-in if you would like to make the request again. If you would like for us to delete your account so that you can start the process from scratch, please let us know.

PR #2990, opened 2019-05-18
added bitfinex

DorothyRMartin:
added bitfinex to the international exchanges list

### DIFF
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -13,6 +13,8 @@ id: exchanges
   <p>
     <a class="marketplace-link" href="https://www.bitstamp.net/">Bitstamp</a>
     <br>
+    <a class="marketplace-link" href="https://www.bitfinex.com/">Bitfinex</a>
+    <br>
     <a class="marketplace-link" href="https://www.coinbase.com/">Coinbase</a>
     <br>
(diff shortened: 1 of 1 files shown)


Cobra-Bitcoin:
NACK. Bitfinex has major problems right now.

DorothyRMartin:
You can deposit fiat, you can withdraw fiat.
You can deposit crypto, you can withdraw crypto.
Yes they have some legal action, regardless, NYAG has said trade on, the 'loan' they didn't like was returned and money was raised from investors.

So whats the problem, or is this just some kind of personal issue you have with them which you're unable to articulate ?

wbnns:
@DorothyRMartin Thanks for the PR, but agree with @Cobra-Bitcoin and also think the exchange shouldn't be added at this time.

DorothyRMartin:
Okay, what requirements are we waiting for them to fill exactly to be added ? 
Like I said, neither has provided any reason to not to add them... 
In the spirit of being fair and unbiased, i do not see why not.
They help finance bitcoin core development, first to implement liquid / blockstream elements, have a program for funding lnd development, their platform works, better than stamp/coinbase in terms of liquidity, spread, reliability, ws api etc.

PR #2992, opened 2019-05-21
Fix quoting

cornelius:
Remove redundant trailing double quote


wbnns:
@cornelius Thank you.

PR #2998, opened 2019-06-03
Fix incorrect colors associated with wallet scores

wbnns:
As part of the wallet resdesign that is currently ongoing (#2861), it was discovered that scoring colors (green, orange and red dots) are being misreported on the individual wallet pages alongside their associated scoring criteria. While scoring criteria rated as `good` were showing associated with the color green, `pass` ratings were incorrectly displaying as green as well, when in fact they should have been orange, and `fail` ratings were showing as orange, when in fact they should have been red:

Screen Shot 2019-06-03 at 23 03 02

This PR corrects the mismatches so that the correct colors are displayed in association with the three aforementioned rating classifications:

Screen Shot 2019-06-03 at 23 03 42

Cc: @crwatkins, @Cobra-Bitcoin, @natiwa, @alexcherman, @maxwellmons


Cobra-Bitcoin:
LGTM. Good catch by the way. I don't know why I never noticed this before.

crwatkins:
Here is my recollection of the history related to these colors:  The original (2014) design displayed scoring levels (good, pass, fail, neutral)  in colored text (green, orange, or gray).  There was no distinction between good and pass; fail was shown as orange.  I believe this was the intended behavior.

During the recent redesign, we changed to colored balls.  Good and pass remained green, and fail remained orange, but neutral became a green ball.  I've documented that [here](https://github.com/bitcoin-dot-org/bitcoin.org/issues/2861#issue-411244142).

Note that as best as I recall, we've never displayed fail in red, so this PR introduces a change in design.  I'm OK with trying that out as I have proposed this exact behavior for our next redesign.  However, I would also request regressing back to the original design of gray for neutral.

wbnns:
@Cobra-Bitcoin Thanks.

@crwatkins 

> I would also request regressing back to the original design of gray for neutral.

Pushed 2829ac5 to ensure gray is displaying when neutral ratings are present (e.g. hardware wallets):

Screen Shot 2019-06-04 at 10 36 11

harding:
@crwatkins wrote:

> The original (2014) design displayed scoring levels (good, pass, fail, neutral) in colored text (green, orange, or gray). There was no distinction between good and pass; fail was shown as orange. I believe this was the intended behavior.

Yeah.  The idea was that `good` is for advanced features and `pass` is for acceptable features, so they're both green.  Here's @saivann discussing the earlier iteration of the page in #473:

> basically, present a good score (bold green) to wallets supporting a defined set of advanced privacy features like those mentioned above, and a passing score (normal green) to wallets supporting privacy features all wallets should be supporting. Keep it simple.

Whether or not the colors should be changed, I have no opinion.  I'm here only as the ghost of history past.

wbnns:
Here's a live preview:
https://bitcoin.cryptopelago.com/en/wallets/mobile/ios/brd/

...compared to the current version:
https://bitcoin.org/en/wallets/mobile/ios/brd/

@harding Thanks.

crwatkins:
@wbnns thanks for preparing the preview.  That helps visualize the presentation.  I can see why the original design used bold green and normal green (for **good** and **pass**).  I believe that when both green and orange are presented on the same listing (in the current PR), orange might appear to be a warning as opposed to the implied intent of "perfectly fine."  I think this will particularly be an issue when we implement wallet comparisons in #2861.

I'm hoping someone has a great idea how to represent this better.  Lacking any other suggestions, I would probably fall back to two shades of green (or an otherwise stylized dot) that both imply OK.

wbnns:
@crwatkins 

Thanks for following up. Regarding:

> Note that as best as I recall, we've never displayed fail in red, so this PR introduces a change in design. I'm OK with trying that out as I have proposed this exact behavior for our next redesign. However, I would also request regressing back to the original design of gray for neutral.

And:

> Lacking any other suggestions, I would probably fall back to two shades of green (or an otherwise stylized dot) that both imply OK.

Is your present recommendation to not show green, orange, gray and red colors on the current site (this PR), or in the redesign (#2861), and to show bold green, green, gray and orange?

maxwellmons:
Hi @wbnns and @crwatkins  I do not think it makes sense if you are showing two different shades of the same color for two different ratings. People are just going to see the color green and are not even going to read the explanation. Wallets should want to have good scores. If they don't want to see the color orange or red associated with their wallet, they should make the required changes to get a good rating instead of a passing or failing one.

crwatkins:
> Is your present recommendation to not show green, orange, gray and red colors on the current site (this PR), or in the redesign (#2861), and to show bold green, green, gray and orange?

I would recommend "fixing" our current site with the original color scheme of bold green, green, gray, and orange.  That will correct a perhaps unintended change that occurred during the previous redesign and we can see how that works with the new balls in anticipation of the comparison pages.

wbnns:
@crwatkins 

Here's an updated screen shot using dark green:

Screen Shot 2019-06-06 at 12 53 11

...compared to green, orange and red:

Screen Shot 2019-06-06 at 12 53 58

crwatkins:
I like that better.  “Secure environment” should not be an orange “warning.”

wbnns:
@crwatkins:

Pushed aa70cb68 to change scoring colors to dark green (good), green (pass), orange (fail) and gray (neutral).

@maxwellmons:

Thanks for the feedback. Agree that there is room for improvement - let's use #2861 for that.

---

Unless others object, this is scheduled to be merged on Monday, June 10th.

crwatkins:
utACK aa70cb6

PR #3002, opened 2019-06-06
HODLER Wallet

zurekpiotr:


crwatkins:
@zurekpiotr  Thanks for the submission!

Regretfully, there are a few PRs currently in the queue, so the review will not start immediately.
You can help speed up the process by making sure that the wallet strictly complies with all of the [listing criteria](https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/docs/managing-wallets.md) and requesting any clarifications or discussing any discrepancies here in the meantime.

One issue to note is that on bitcoin.org we use the [Wikipedia Manual of Style](https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Trademarks) for trademark representation (for example, we list [TREZOR as Trezor](https://github.com/bitcoin-dot-org/bitcoin.org/pull/921)).  Could you make the appropriate changes to your submission?

zurekpiotr:
Thanks for your reply,
We are working on a new UI/UX. We try to release new version as soon as possible.
If the new version will be available, I will reopen the issue with the correct commit.

PR #3016, opened 2019-06-17
Added link to price comparison

ton77v:
ExchangeRates.Pro is a new website allowing to compare prices across exchanges, brokers and P2P market


wbnns:
@ton77v 

Thank you, however, the current phrasing is sufficient. FYI, we are going to open a separate PR to remove Cryptoradar since it is more about cryptocurrencies in general rather than specific to bitcoin.

PR #3017, opened 2019-06-17
Wallet selector redesign

alexcherman:


wbnns:
Adding some TODOs, below, for reference purposes on our side:

- [ ] Add pros/cons underneath each different operating system choices
- [ ] Add copy for beginners
- [ ] Add copy for advanced users
- [ ] Add copy for discover more
- [ ] Finalize criteria
- [ ] Add copy for various criteria
- [ ] Finalize features
- [ ] Add copy for various features
- [ ] Finalize desktop flow
- [ ] Finalize tablet flow
- [ ] Finalize mobile flow

Cobra-Bitcoin:
Can't seem to see the new redesign in the preview, was eager to start testing some stuff.

wbnns:
@Cobra-Bitcoin Hey, try here: https://bitcoin.cryptopelago.com/en/choose-your-wallet

Cobra-Bitcoin:
@wbnns Thanks, works now! :)

_includes/layout/base/wallets-selector.html
```diff
@@ -0,0 +1,601 @@
...
+        <h3 class="sidebar-selector-title">Need some help?</h3>
+        <p class="sidebar-selector-text">You can find more information and help on the resources and community pages or on the Wiki FAQ</p>
+      </div>
+
+      <div class="wallet-selector-content">
+        <div class="wallet-table-wrapper">        
+          <div class="wallet-table">        
+            <div class="wallet-table-row">
+              <div class="wallet-table-head">
+                <p class="wallet-table-title">iOS Wallets:</p>
```
wbnns: Let's make sure that this:

1. Updates dynamically based on the Operating System that is chosen, and,
2. Doesn't appear if there is no Operating System chosen yet

_includes/layout/base/wallets-selector.html
```diff
@@ -0,0 +1,601 @@
...
+              </div>
+              <div class="wallet-table-head">
```
wbnns: Let's not show the table headers until an operating system has been chosen.

_includes/layout/base/wallets-selector.html
```diff
@@ -0,0 +1,601 @@
...
+                      <div class="wallet-table-data">-</div>
+                      <div class="wallet-table-data">-</div>
+                      <div class="wallet-table-data">-</div>
+                    {% endif %}
+                  </a>
+                {% endfor %}
+              {% endfor %}
+            {% endfor %}
+          </div>
+          <div class="wallet-hints-row">
```
wbnns: Let's not show the hints until an operating system has been chosen.

_includes/layout/base/wallets-selector.html
alexcherman: Yes

_includes/layout/base/wallets-selector.html
alexcherman: Are they visible? I hid them by default. They are visible only if the operating system is selected and at least one wallet matches the filter values.

_includes/layout/base/wallets-selector.html
alexcherman: Are they visible? I hid them by default. They are visible only if the operating system is selected and at least one wallet matches the filter values.

wbnns: Apologies, you're right - I wasn't looking at an updated preview. Thanks for taking care of this.

_includes/layout/base/wallets-selector.html
wbnns: All good. 👍

_includes/layout/base/wallets-selector.html
wbnns: Thanks

crwatkins:
Shall we use the wallet **titleshort** instead **title** in the listing?  In the past, I believe we've used **titleshort** when multiple wallets are listed.

crwatkins:
We currently filter the categories on **good** vs **pass** but we don't visually distinguish between **good** and **pass** in the listings which looks confusing.  Shall we add the dark green here for **good**?

wbnns:
@crwatkins 

Thanks, pushed 31e4381 to utilize `titleshort` and a [TODO](https://github.com/bitcoin-dot-org/bitcoin.org/pull/3017#issuecomment-503289755) to distinguish between good and pass.

crwatkins:
I believe all the colons on the new page are unnecessary and could probably be removed.

wbnns:
@crwatkins 

> I believe all the colons on the new page are unnecessary and could probably be removed.

59f6000

Cobra-Bitcoin:
Noticed that if you scroll down so you can see the full list of features, some of the results table is cut off and users will need to scroll back up to see the effects of their filter. I'm assuming that eventually the filter will float next to the content always like the other sidebars on the site do?

wbnns:
@Cobra-Bitcoin 

Ok, I've added a [TODO](https://github.com/bitcoin-dot-org/bitcoin.org/pull/3017#issuecomment-503289755) to ensure the selector sidebar remains in a fixed position when scrolling.

crwatkins:
Is it time to discuss the "User type" or is that still WIP?  Currently the UX is a bit strange in that it is a radio button so it has to be either one or the other, unless it is in the original state of neither (there's no way to get back to that).   Another concern is that "New" wallets aren't displayed for "Experienced" users leading to a situation where, for example, we don't display Electrum for "Experienced" users.  Perhaps Experienced users should see them all?   Or maybe a single filter checkbox for "Show only wallets suitable for new users" or something like that?

wbnns:
@crwatkins 

This is still being worked on, but yes, we should show all wallets when a user chooses Experienced as their user type.

Cc: @alexcherman

Cobra-Bitcoin:
Let's keep in mind the wizard which doesn't seem to be implemented yet will be the main way users end up on the choice of wallets page with the filter options pre-filled based on what they chose on the wizard? Am I correct to assume this is still the intended plan? Majority of users probably shouldn't interact with the filter side-menu directly, since the wizard can dumb it down for them in an easier to digest way.

Cobra-Bitcoin:
By the wizard I mean the main "Choose your wallet" page will look like this: https://projects.invisionapp.com/share/EVPQ1WPSBG8#/screens/356509387_Bitcoin-org_-_Wallet_Flow_2-01_-_Desktop

Users will then just follow the wizard, and then land on a pre-filtered list of results, the majority won't mess around with the filters after that unless they feel adventurous.

wbnns:
@Cobra-Bitcoin Hey, yes, that's correct regarding the intended plan - the wizard is still being implemented and the current page is just one component of it.

_includes/layout/base/wallets-selector.html
```diff
@@ -0,0 +1,561 @@
...
+              </div>
+            </div>
+          
+            {% for wallet in site.wallets %}
+              {% for platform in wallet.platform %}
+                {% for os in platform.os %}
+                  <a
+                    href="{% if platform.name == os.name %}/{{page.lang}}/wallets/{{platform.name}}/{{wallet.id}}/{% else %}/{{page.lang}}/wallets/{{platform.name}}/{{os.name}}/{{wallet.id}}/{% endif %}"
+                    class="wallet-table-row wallet-link"
+                    data-categories="{{wallet.user}} {% for check in os.check %}{% assign checkValue = check[1] %}{% assign checkName = check[0] %}{% if checkValue contains 'checkgood' or checkValue contains 'checkpass' %}{{checkName}} {% endif %}{% endfor %}{{os.name}}"
```
wbnns: @alexcherman 

Let's show all wallets for the `experienced` user type, but continue to only show wallets with the `new` value for the `new` user type.

alexcherman: ok, I'll do it

wbnns:
@crwatkins 

For the setup wizard, could you please let us know what would you recommend be the pros/cons to list under each type of wallet on this page:

image

_includes/layout/base/wallets-selector.html
wbnns: Thank you.

Cobra-Bitcoin:
The helper/wizard should be on /choose-your-wallet, not on its own URL. Otherwise all the existing links to /choose-your-wallet won't benefit from visitors being guided by the wizard. When they click "Skip" it should just hide the wizard on the _same_ page, and show the table. Basically I think it makes the most sense to have everything happening on the /choose-your-wallet page. Otherwise it's the worst experience possible since visitors arriving straight to the /choose-your-wallet page from external sites will see an empty table and a whole bunch of options/icons and be confused. Is having the wizard on its own separate page just a temporary measure?

wbnns:
@cobra-bitcoin Yes, we need to clean that up - it's only temporary.

alexcherman:
> The helper/wizard should be on /choose-your-wallet, not on its own URL. Otherwise all the existing links to /choose-your-wallet won't benefit from visitors being guided by the wizard. When they click "Skip" it should just hide the wizard on the _same_ page, and show the table. Basically I think it makes the most sense to have everything happening on the /choose-your-wallet page. Otherwise it's the worst experience possible since visitors arriving straight to the /choose-your-wallet page from external sites will see an empty table and a whole bunch of options/icons and be confused. Is having the wizard on its own separate page just a temporary measure?

No, because I thought that the user should have a choice either to be navigated to the wizard page or to be navigated directly to the choose-your-wallet page. If everything is on the same page and the user doesn't want to use the wizard, the browser will render a lot of unnecessary HTML and javascript. But if you want I can do it on the same page. Should I do it on the same page?

wbnns:
@alexcherman

The preference will be for users to start with the wizard, like in the flow. The 'choose-your-wallet' URL will be the first page in the flow.

alexcherman:
> The preference will be for users to start with the wizard, like in the flow. The 'choose-your-wallet' URL will be the first page in the flow.

ok, so I'll move everything on the same page(choose-your-wallet)

wbnns:
@alexcherman Ok, thanks.

_includes/layout/base/helper-step-2.html
```diff
@@ -0,0 +1,64 @@
...
+      <div class="helper-page-number">4</div>
+    </div>
+  
+    <div class="helper-content">
+      <div class="helper-header">
+        <div class="helper-header-row">
+          <h2 class="helper-title previous">What’s your operating system?</h2> 
+          <div class="row">
+            <div class="helper-selected-filter">Desktop wallets / OS X</div>
+            <button class="helper-change-btn">Change</button>
```
wbnns: @alexcherman Just a reminder that we need to make sure these `Change` buttons function as intended as users progress through the wizard.

wbnns: Let's also make sure that the parameters are updating correctly to show the selected values:

image

alexcherman: Yes, I'm still working on it

wbnns: Ok, thanks.

crwatkins:
The criteria chooser requires me to select at least one criteria before going forward.  Most new users aren't going to know and we should let them continue without any filtering.  Also, there should be some instructions such as "Select 0-6 criteria which are important to you".

How are we going to help the user when they select some combination in step 2, 3 or 4 that results in no items to display in step 5?

In the feature list, I regret we may have to add the word "Support" to every feature title.  Features such as "Hardware Wallet" is too confusing otherwise.  It should be "Hardware Wallet Support"  Likewise, some other suggested text changes:

Legacy Addresses -> Legacy Address Support
Bech32 -> Bech32 Address Support

wbnns:
@crwatkins 

> The criteria chooser requires me to select at least one criteria before going forward. Most new users aren't going to know and we should let them continue without any filtering. Also, there should be some instructions such as "Select 0-6 criteria which are important to you".

For the instruction, we ask the question, "Which criteria are the most important to me?" with checkboxes next to each item. This lets users know they need to make a selection. Alternatively, they can skip the helper.

> How are we going to help the user when they select some combination in step 2, 3 or 4 that results in no items to display in step 5?

In this case, a user sees a message informing them that no wallets match their query to please update their search criteria.

> In the feature list, I regret we may have to add the word "Support" to every feature title. Features such as "Hardware Wallet" is too confusing otherwise. It should be "Hardware Wallet Support" Likewise, some other suggested text changes:

Instead of adding "Support" to every title, thinking we should update the section from "What features am I looking for?" to "What features would you like your wallet to support?"

crwatkins:
Is the old filter page still available somewhere in the live preview?  _Skip helper_ doesn't seem to work.

alexcherman:
> Is the old filter page still available somewhere in the live preview? _Skip helper_ doesn't seem to work.

No,  I hid this screen while I'm developing wizard screen because right now they on the same page.

crwatkins:
>For the instruction, we ask the question, "Which criteria are the most important to me?" with checkboxes and descriptions next to each item. This lets users know they need to make a selection. Alternatively, they can skip the helper.

I'm happy to see how this looks as work progresses, however I have some hunches on what we might find.  Since we are focusing the wizard on new users that once "new user" is chosen

1.  A new user won't be able knowledgeable enough to choose criteria.  Even as an expert, I have a really hard time choosing which one or two are most important to me.

2.  After a new user filters on new user wallets, there are are so few wallets to choose from that selecting criteria is most likely to be moot.  It's likely to prune the entire list or none at all.

Perhaps selecting "New user" should skip the selecting criteria page.

PS As I mentioned before Environment is an awkward criteria to allow selection on since it was basically already selected when the platform was chosen.

wbnns:
@crwatkins 

> A new user won't be able knowledgeable enough to choose criteria. Even as an expert, I have a really hard time choosing which one or two are most important to me.

Pushed 627d291c to simplify this further; we explain all of the terms in the section so they can decide what's important. Or again, they will be able to skip the wizard should they desire to do so at that step.

> After a new user filters on new user wallets, there are are so few wallets to choose from that selecting criteria is most likely to be moot. It's likely to prune the entire list or none at all.

Would recommend using #3022 to refine and potentially widen the scope of what wallets should be shown to new users if this is a concern (e.g. considering the inclusion of user-friendly hardware wallets since they provide a lot of security and other benefits), instead of skipping over all criteria and features and not helping users (especially new ones) become more well-informed about them.

Cobra-Bitcoin:
On the wizard we have to remember to filter out "impossible" combinations. Maybe make the box un-selectable? It could look like this:

impossible

The logic should be: if selecting that feature/criteria combination leads to an empty page, then make it impossible to select. For example if the user selects iOS as their platform, and then specifies they want "full node" functionality from a wallet, it should be impossible to select since there's no wallets that can offer this. Maybe we can also display a small message saying something to the effect of "Your choice of platform makes it impossible for you to enjoy certain features and criteria"? Without that it's going to be misleading a bit since we'll offer up all these criteria/features to users, but if they select through them willy-nilly not understanding fully what they're doing, they'll land on an empty results page a lot of the time. Ideally it should be impossible for people to land on empty results when they're being guided by the wizard.

I don't know how it could be done, maybe after each click/choice, do a quick look-ahead of the remaining options and filter out those ones who if selected would reduce the matching wallets count to 0. Then do the reverse when an option is un-selected. I know it's not nice to implement something like this and it's probably going to be annoying to test, but it makes the wizard much smarter and more useful.

wbnns:
@Cobra-Bitcoin 

Yes. Maybe we should experiment with just not showing a selectable option at all, if it's going to be unavailable - that might be the most straightforward solution. Users would only see what's available based on what they have selected so far.

Cc: @alexcherman

crwatkins:
@Cobra-Bitcoin I agree.  I've commented before that we shouldn't allow users to go through the process and end up "empty handed."  I think my following suggestion is a bit overkill, but I'll mention it for consideration:  Each criteria could have an indication of what effect that selection would have.  For example we could indicated if you select Full Node there will be zero wallets to choose from, and if you select 2FA, there will be one.

@wbnns I think removing the options completely based on previous choices would even be more confusing (perhaps not to us, but probably to users not familiar with the wallets already).

wbnns:
@crwatkins Ok. 

@natiwa, can you please update the desktop flow, based on the feedback from @Cobra-Bitcoin and @crwatkins to handle this use case?

wbnns:
Maybe we could present features that aren't available based on what has been selected by the user, like this:

image

Cc: @Cobra-Bitcoin @crwatkins @natiwa

_wallets/electrum.md
```diff
@@ -15,6 +16,7 @@ platform:
       link: "https://electrum.org"
       source: "https://github.com/spesmilo/electrum"
       screenshot: "electrum.png?1528322191"
+      features: "2fa bech32 hardware_wallet legacy_addresses multisig segwit"
```
wbnns: @alexcherman 

Added the features for each wallet.

_wallets/electrum.md
```diff
@@ -41,6 +43,7 @@ platform:
         link: "https://play.google.com/store/apps/details?id=org.electrum.electrum"
         source: "https://github.com/spesmilo/electrum"
         screenshot: "electrumandroid.png?1528322191"
+        features: "bech32 legacy_addresses segwit"
```
wbnns: @alexcherman 

Features may differ depending on environment - added values to reference underneath desktop and mobile os type (if a wallet is available in multiple environments).

Cobra-Bitcoin:
@wbnns That seems ideal as they can see and know they're missing out on a feature.

wbnns:
@Cobra-Bitcoin Ok. 👍

Cc: @natiwa, @alexcherman

crwatkins:
* Wizard: Which criteria are important to me?

I remain concerned that we actually force users to choose at least one criteria before they can continue to use the wizard and proceed to select which features they are interested in.

1.  Maybe none of those things are important to a particular user

2. Or more likely, brand new users will have no basis to choose some of these criteria over others and if we force them to make a choice, it will simply artificially limit  or entirely remove the wallets matching their feature criteria.

Cobra-Bitcoin:
Been testing this every so often, and it's looking really good after the most recent commits. One suggestion though, on the helper let's make the feature selection optional, so show the "Next" button and let the user go to the wallet results with an empty "features" selection. I've been kind of forced to pick _something_ from the features just to progress, even if I'm not particularly interested in that feature. Even worse when with certain combinations, there's only 1 or 2 features to pick from. We don't want users just picking any random feature to progress since it isn't helpful to them and they'd benefit more just seeing the full list of wallets based on their other criteria.

Looks very close to complete though. Really nice job on this!

alexcherman:
> Been testing this every so often, and it's looking really good after the most recent commits. One suggestion though, on the helper let's make the feature selection optional, so show the "Next" button and let the user go to the wallet results with an empty "features" selection. I've been kind of forced to pick _something_ from the features just to progress, even if I'm not particularly interested in that feature. Even worse when with certain combinations, there's only 1 or 2 features to pick from. We don't want users just picking any random feature to progress since it isn't helpful to them and they'd benefit more just seeing the full list of wallets based on their other criteria.
> 
@wbnns Do you confirm it?

> Looks very close to complete though. Really nice job on this!
@Cobra-Bitcoin Thanks!

wbnns:
@alexcherman Thanks, yes - let's try it.

wbnns:
@Cobra-Bitcoin 

Hey, just wanted to follow-up regarding the fixed sidebar. We were looking into this, and noted that unlike some of the other pages on the site (e.g. Vocabulary, Exchanges) where there is a fixed sidebar, on the wallet selector page the sidebar is larger than the center content area. In other words, the table where the wallets appear takes up less area than would be seen on a page like the Vocabulary page, for example, where there is a lot of content to scroll through. In this case, there would be very little if any scrolling at all - so we were thinking it may be ideal to leave as-is rather than fixed.

Cc: @alexcherman

Cobra-Bitcoin:
@wbnns No problem, that makes sense. Let's leave it as is then.

crwatkins:
@wbnns During this refactor, could we remove the dependency of the wallet files on having a cache buster parameter on images?  For example:
```
      screenshot: "bitcoincore.png?1528322191"
```
That would be great because of the frequent questions I get from contributors about that.  If they are still necessary and for some reason we can't centralize the query string, let me know and I'll write up a separate issue requesting proper documentation of it for contributors.

wbnns:
@crwatkins 

Thanks for mentioning - yes, let's do this.

crwatkins:
Just a reminder that we have not yet resolved the platform comparison issues.  Recall that when I tried to fix all the problems that I noticed with the current pros and cons, I could not figure out how to fix them all and retain the current format of pros and cons so I rewrote the descriptions in a more narrative format.  If someone really wants to retain the current format, I can start enumerating the issues, but I don't know how to solve them all in that format.

wbnns:
@crwatkins 

Can you please reply with a checklist or bulleted list?

crwatkins:
Here are a few of the things I tried to solve:

* App marketplaces can delist/remove wallet making it difficult to receive future updates

That doesn't seem very important; at least not the top reason not to use a mobile wallet.  Apple did this five years ago and I know of no such problems since.  Even if a wallet is delisted, the existing apps continue to work.  I've never heard of Google or Apple removing a wallet from client devices.  This sounds like FUD.

* Damage, loss of device or mobile viruses can potentially lead to total loss of funds

Mobile viruses affecting wallets (on non-jailbroken phone -- which most wallets won't run on) are virtually non-existent.   Again, certainly not a compelling reason to avoid a mobile wallet.

* Designed to be able to use QR codes to make quick and seamless transactions

I don't know what seamless transactions are, but if I were to guess, I don't see how this is a plus for mobile and not for any of the other types of wallets.   

* Funds can potentially be recovered if a device is damaged or lost

I hope that's true for any of the categories.  It's not a pro just for web.

* Requires trust in a third party and shared control over funds

Every solution here requires some significant trust in a third party.  Web wallets may in general (but not certainly) require some degree more.   Web wallets do not necessarily imply shared control over funds.

* Web wallets compromised by hackers can potentially result in loss of funds

Any wallet compromised by hackers can.  This is not a unique characteristic of web wallets.

* Environment enables users to have complete control over funds

I think I would argue that's not a unique characteristic.

* Some desktop wallets offer hardware wallet support, or can operate as full nodes

Some mobile wallets offer hardware support.  There are certainly web wallets that offer hardware support.  This is not a unique qualification.

* Loss of device without proper backup can make funds unrecoverable

Again, I don't find this unique to any one category and not the others.

-----

I could go on, but you get the idea.

crwatkins:
@wbnns Thanks for 3539f77 !!

wbnns:
@crwatkins 

Thanks. On my side, we've encountered lots of users who don't know anything about bitcoin wallets, and also don't know which platform they would select. Thinking it would be better to come up with some different pros/cons rather than redoing the section and/or dropping them altogether. 

For example, helping people know they might want a mobile wallet since it is ideal for scanning QR codes, etc (pro). Or a hardware wallet if they are going to be storing large amounts (pro). But not a desktop wallet if they plan to use it to buy something when visiting an establishment that accepts bitcoin (con). Or that there can be scam apps and wallets are sometimes impersonated in App marketplaces (con). Stuff like that.

But let's see what other people say/think when we solicit more feedback from the community.

crwatkins:
@wbnns What's wrong with what I wrote up?

wbnns:
@crwatkins 

Apologies, I wasn't criticising anything you wrote - was saying let's open this up to community feedback, and that if you'd like to see more changes to the section because you feel there are issues ([1](https://github.com/bitcoin-dot-org/bitcoin.org/pull/3017/#issuecomment-517253498), [2](https://github.com/bitcoin-dot-org/bitcoin.org/pull/3017/#issuecomment-517427721)), was thinking we could further refine the pros/cons (along with some ideas), rather than redesigning how users select a platform, or dropping them.

crwatkins:
Thanks @wbnns. No need for any apologies!

wbnns:
Just an FYI to all who are following this PR, we've added a new TODO for something we've discovered in testing, which is that we need to have a fallback for people who have JavaScript disabled. We're thinking we'll just fallback to the current design (https://bitcoin.org/en/choose-your-wallet), instead of a message or something to that effect which doesn't let them browse wallets and requires them to enable JS.

wbnns:
@alexcherman Thanks.

crwatkins:
Here are some issues I notice in the preview (some I've brought up before and don't recall a resolution (apologies if I missed it) and some are new things I've just noticed):

1.  The Coin (wallet) logo is broken
1. On a wallet page, the layout of the Privacy subcategory explanations is broken (on desktop and mobile).  The headers don't look like headers  because of layout.
1. We still force a user to select at least one Criteria before being able to continue in the wizard. In a tiny bit of user testing, I got complaints about that.  I thought we changed that.  In any case, it's very awkward, particularly apparent  when there are only two choices or when actually making the choice doesn't even affect the list.
1. We should decide if bech32 should be capitalized and be consistent.
1. I still think our feature named "Hardware Wallet" will be confused by many thinking they are selecting hardware wallets (they won't remember that was a choice two steps ago and analyze the fact that one or the other must be wrong and come to the correct conclusion).
1. In the list of wallets, I find the current icons mildly disturbing.  I know that sounds overly dramatic, but that's the best word that I can use to describe the cognitive dissonance I experience when looking at the chart.  My mind tries to match patterns and discern better and worse until it finally gives up.  I fear that the only reason I figure it out at all is because I've been working with this for months.  I regret that I think I'm much better at spotting these types of problems than I am at fixing them.  I've made some suggestions in the past, most of which I think are better than the current situation, but I really hope someone can come up with something better here. 
1.  Likewise, the good and pass labels are confusing.  Based on an extremely small, non-scientific survey, I found people disagreed on which was better.  I propose we either leave off the key altogether, or change them to better and good, or come up with something entirely different (just don't leave them they way they are).   
1. I'm glad we were able to "shoehorn"  the hardware wallets into the scheme once again, but that leaves us with some confusing selections (such as many of the features, particularly "Hardware Wallet" support), so we may way to revisit splitting out hardware wallets in the future.
1.  Likewise for the future, we probably should expose the feature attributes of wallets (we don't do that now, right?).  I've seen users toggle features on and off one after another to try to determine which features a particular wallet supports.

crwatkins:
In the wizard, I've heard complaints of not being able to find the [Next] buttons because it is below the fold on mobile and even on some desktops.  People think they've already clicked a button/link (e.g. I'm a new user) and don't know what to do next.

In a related layout note, it would be best if our [Back] button didn't move.  When the [Next] button is hidden, the [Next] button replaces the [Back] button when it becomes visible and the [Back] button moves.  It is a better UX to not have the [Back] button move on the user.

crwatkins:
@wbnns Oh Nooooo!.... Don't drop the Hardware Wallet support feature!  That's a very important feature. (or did I misunderstand the commit?)

wbnns:
@crwatkins 

Thanks for the comments. Not sure you understood the commit. I've elaborated on the HW wallet change, below.

Regarding the numbered items (5, 8):

1. Can you please try clearing your cookies and cache and let us know if you still don't see it? Are you referring to the thumbnail in the selector/table or the screenshot on the individual wallet page for Coin?
2. Resolved w/  [244ef4e](https://github.com/bitcoin-dot-org/bitcoin.org/pull/3017/commits/244ef4e36b5d0aaaeeba5f297d86f4fccf10c69e) .
3. Users should be able to proceed without choosing any criteria; if you're still experiencing this, please let us know the steps needed to reproduce the issue and what browser/device it's being encountered on.
4. There are instances in the repository aside from the wallet pages where it is both capitalized (`_releases/v0.16.0.md`) and not capitalized (also `_releases/v0.16.0.md`). We recommend this to be a opened as a separate, non-blocking issue unrelated to the wallet page redesign if it is a concern. 
5. Since we only have one wallet that would appear when this feature is checked, we think it would be ideal to drop this feature for now, until at least an additional wallet is available. Also, since most wallets will be able to send to a hardware wallet (a subjective level of "Hardware Wallet Support" from the perspective of the site visitor), we feel leaving as-is could be even more confusing since unchecking the box might create some type of subjective insinuation that other wallets don't work with a Hardware Wallet [05a5366](https://github.com/bitcoin-dot-org/bitcoin.org/pull/3017/commits/05a5366a4690fc06be46a28efa297d3c74fa1457). Users will still see all Hardware Wallets when they select that as their environment. This only refers to the feature, where only one wallet was available.
6. There is a legend that explains the symbols. We haven't encountered any feedback from multiple people that they don't understand the meaning. Also, we feel the current symbols are a significant improvement over the current wallet pages since people with color sensitivities may only be able to see circles. Similarly, when looking into Harvey Balls as an alternate solution we found that they were not very prevalent in their use, especially outside of the United States (even in the United States, aside from Consumer Reports magazine) - whereas shapes are very common around the world in terms of conveying meaning. This, combined with the legend, we feel is an upgrade over the current layout. We propose to leave as-is, and/or to create a subsequent issue for "further improvement of scoring colors/shapes" once the new wallet pages go live, since they are already better than the current pages - especially for people with color sensitivities.
7. We don't feel that "Pass" is Good as per the suggestion ("Better", "Good"), for example, especially in the context of Privacy. Alternatively, perhaps we could use the word "Acceptable" instead of "Pass"?
8. Pushed [05a5366](https://github.com/bitcoin-dot-org/bitcoin.org/pull/3017/commits/05a5366a4690fc06be46a28efa297d3c74fa1457) as an additional user experience tweak to further eliminate any potential confusion.
9. @natiwa @alexcherman Let's add a bulleted list of features to the individual wallet pages. We're thinking this isn't a blocking addition to prevent us from going live with the pages, so would propose we do this as a separate PR on top of the new design once we're live.
10. @alexcherman When people choose an option, please jump the user down the page to the next button.
11. Thinking we can probably drop the "Back" button, since we already have a "Change" button next to each section heading - it's the same as the "Back" button and is always in the same place.

crwatkins:
@wbnns 

1.  No longer an issue.
2.  OK
3. Now I don't have to select a criteria, but on Chrome on desktop and iOS I have to select the [Next] button twice (the first time just scrolls the page).
4. OK
5.  I disagree.   I believe that being able to find a choice compatible with hardware wallets is an extremely important feature.   If we go by that logic, we should also remove the Lighting feature and the Mixing feature because we only have one wallet in those categories.   There's another wallet almost ready to be listed (Green) which also supports hardware.
6. I recommend circles of different sizes, or rings around the circles.  
7.  I could go with "acceptable"
8. Removing it is much much worse than any of the issues I encountered
9. Agreed
10.  Let's see how that plays
11. I have no objection to the back button itself

Thanks!  Can't wait to get this live!

crwatkins:
This is another one of those nits that should not block release, but perhaps "Neutral" should be considered in the same category as "Good" for filtering (of course this applies only to the niche of hardware wallets).  In other words, checking Fees should not remove all the wallets from the listing.  Perhaps an even better solution would be to disable Fees, Validation, and Privacy Criteria if hardware wallets are selected, but I would be happy with my first suggestion also.

maxwellmons:
I really like the shapes that are being used for good, pass, neutral, and caution. They look great.

crwatkins:
A couple more thoughts:

We've always randomized the order of the wallet listing.  The listings here seem always alphabetical.  Should we randomize?

Secondly, our current interface allows a user to see all the wallets, and find a wallet that they can use on multiple platforms (for example maybe a wallet that works on both Android and Windows).  In the new interface, I haven't found a way to determine which platforms a given wallet supports.  Maybe we could add that information back at some point in the future.

wbnns:
@crwatkins 

3. Ok, thanks for the info - @alexcherman is checking into this.

--

5. Dropped the commit to remove and pushed bcc4910 to further revise the text to help eliminate confusion

--

6. Circles of different sizes may look odd in the table and make it look like the stylesheet is broken or not rendering correctly for the associated viewport. This becomes more problematic on smaller devices, like phones. Rings around circles are hard to see because of the size of the circle (also problematic on smaller devices like phones), and lastly, can make it more difficult for people with visual sensitivities. This is why we recommend shapes. They are easy to distinguish from one another and are also the same proportionate height and width. Can open a separate issue to see how we can potentially iterate further post launch if you'd still like to pursue something different?

--

7. Pushed 300da0b to change 'Pass' to 'Acceptable'.

--

8. No longer removed (bcc4910).

--

12. @alexcherman Let's update this logic as described (for hardware wallets only, include wallets in the list if a criteria is selected and the hardware wallet has a neutral score associated with the selected criteria).

--

13. Since the selector is designed to be used to be able to quickly and easily compare and select wallets, we don't think so. It is confusing if each time users return to the page, the wallets they were looking at last time are appearing in different places. It can also give the false impression that the criteria or feature selections aren't the same as what they previously thought they selected before, since the results will appear in a different order. Lastly, it also makes it harder to find a specific wallet when browsing all wallets for an operating system, since wallets that start with "B" wouldn't necessarily be at the beginning of the list, or those that start with "T" wouldn't necessarily be at the end. The user would have to skim up and down looking one by one. 

--

14. We could allow multiple selections for the operating system on the selector to accomplish this, but not sure it would be clear that the user could select multiple operating systems. Or alternatively, that the the user would know that they need to first unselect the previously selected environment  if they wanted to see a list of unique wallets for another one. The current setup seems much easier - to just let them click between operating systems to see what's available. There aren't that many wallets and it only takes a second or so to switch and compare.

@maxwellmons Thanks.

wbnns:
@alexcherman, @Cobra-Bitcoin, @crwatkins, @maxwellmons, @natiwa

Unless you all, or any others who would like to join the thread to share feedback, object, we would like to propose scheduling a merge date on this PR for Wednesday, August 14th.

In the interim we'll continue to do testing (along with any other fixes should they arise).

Cobra-Bitcoin:
I think we've all looked and played around with the new wallet pages so much that any feedback will be mostly nit picking and minor issues, what's key IMO is going live as soon as we can so we can get the feedback of less experienced or knowledgable users, and just get a sense of the community's reaction. But here are some thoughts:

- When we go live, we really want to understand _how_ users are going through the wizard, what criteria are they selecting the most? Which features do they want the most? What's the ratio between new/experienced and do the experienced users pick different criteria and/or features? For this I'd recommend we use some Google Analytics event or action tracking on when the user clicks the "next" button and saving their checkbox selections so we can study further. 

- I agree that shapes are better differentiators than colour, for the obvious reason that it's better for colour blind folks and it's a lot easier to distinguish on first glance.

- Regarding the ordering I feel like it's almost inevitable that users will click the first wallet in the top of the list, perhaps thinking that this wallet is the best _because_ it's at the top. Is there some way to emphasise that we're ordering alphabetically? I agree that ordering randomly could be confusing, but at the same time we could I guess somewhere have a small [a-z icon](https://cdn3.iconfinder.com/data/icons/text-icons-1/512/BT_sort_az-512.png) in light grey that shows we're ordering by name.

But generally I think the priority is to go live as soon as we can. It's looking really good and essentially complete now.

crwatkins:
@wbnns 
>  It is confusing if each time users return to the page, the wallets they were looking at last time are appearing in different places.

I was recommending the layout order that we’ve “always” done and that I thought we were still doing.  We order wallets by level and then choose a random order of wallets within that level by day. That way the order remains static for a day and I believe addresses @wbnns’ concerns as well as @Cobra-Bitcoin ‘s very valid concerns about preferential ordering. 

We occasionally get questions about why we are listing one particular wallet first or last or whatever, however everyone seems happy when we explain it is random.

wbnns:
Alphanumeric ordering is a pretty standard convention on the internet and across many other forms of media; let's add a UI element like @Cobra-Bitcoin suggested, to denote this. This will make it pretty clear that it's not preferential (unless people take issue with the alphabet).

crwatkins:
@wbnns Why not do what has worked so well for years?  Or are you just rejecting it for now for resource reasons?

wbnns:
@crwatkins 

With the new design, the selection pages are not the same as they have been before. As mentioned above, since the selector is designed to be used to be able to quickly and easily compare and select wallets, it is confusing if when users return to the page, the wallets they were looking at last time are going to be appearing in different places. It can also give the false impression that the criteria or feature selections aren't the same as what they previously thought they selected before, since the results will appear in a different order. Lastly, it also makes it harder to find a specific wallet when browsing all wallets for an operating system, since wallets that start with "B" wouldn't necessarily be at the beginning of the list, or those that start with "T" wouldn't necessarily be at the end. The user would have to skim up and down looking one by one.

I disagree that randomizing the previous list has worked really well; we've seen people asking and/or starting to brigade on social media in the past about why they are presented in the order that they are and accusing us of playing favorites of listing one wallet before another (e.g. Bitcoin Knots before Bitcoin Core), because they don't realize it's random and see that it also isn't alphabetical. 

The UX issues above aren't resolved even if a randomly sorted list remains in a static state for a period of 24 hours before getting reshuffled. We get close to a million visitors to the site each month and we can't assume everyone returns to the website in a 24 hour period. 

We can assume that most people understand how the alphabet works, though - especially when it is clearly sorted that way (as suggested).

wbnns:
Hello, just a note to everyone that we discovered an issue, where-in the old wallet category pages are superceded by the wizard and are no longer accessible:

Old pages:

- https://bitcoin.org/en/wallets/desktop/mac/
- https://bitcoin.org/en/wallets/desktop/linux/
- https://bitcoin.org/en/wallets/desktop/windows/
- https://bitcoin.org/en/wallets/hardware/
- https://bitcoin.org/en/wallets/mobile/android/
- https://bitcoin.org/en/wallets/mobile/ios/
- https://bitcoin.org/en/wallets/web/

We're the first search result for a lot of phrases related to those pages (linux bitcoin wallet, mac bitcoin wallet, windows bitcoin wallet, android bitcoin wallet, etc.), so we need to make sure they are preserved across languages and still accessible to crawlers, before going live.

wbnns:
This has been rescheduled to be merged on Monday, August 19th.

crwatkins:
>This has been rescheduled to be merged on Monday, August 19th.

Yay!  (I still totally object the pros and cons of the different platforms.  I think they're just plain incorrect and embarrassing as I have previously explained in detail and written corrections for.)

Cobra-Bitcoin:
Awesome work everyone! Excited to see this get merged and go live today! :)

wbnns:
Yes, everyone has done some great work. We're looking forward to going live as well, however, apologies, we're going to need to postpone a bit longer.

We've done another round of testing and have discovered what appear to be some JavaScript-related compatibility issues with Internet Explorer that need to be resolved, and we're still able to reproduce an issue with loading older pages that have been replaced by the wizard/new design.

wbnns:
The redirect problem has been resolved; it was related to caching on a testing machine. Clearing the cache resolved the problem.

We're now retesting with Internet Explorer and ensuring there are no additional issues.

wbnns:
The Internet Explorer issues have been resolved and everything has been retested. 

This PR is rescheduled to be merged on Tuesday, August 27th.

PR #3018, opened 2019-06-18
exchanges: Drop Cryptoradar and revise string

wbnns:
In the spirit of what was previously discussed in #2697 where we decided to remove the News section from the Resources page due to an observed trend away from a specific focus on Bitcoin and more-so toward blockchain technologies and cryptocurrencies in general, this drops a link to Cryptoradar off of the Exchanges page.

The page that is linked to is more-so of a directory of price comparisons for many cryptocurrencies and where to buy them (bitcoin being just one), rather than a specific site where users can focus primarily on finding the best rate when exchanging fiat for bitcoin.

Unless others object, this is scheduled to be merged on Monday, June 24.

### DIFF
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -402,7 +402,7 @@ id: exchanges
 </div>
 
 <p class="introlink exchanges-introlink exchanges-buy-link">Visit
-  <a href="https://www.buybitcoinworldwide.com/">Buy Bitcoin Worldwide</a> for user reviews on some of the above exchanges, or <a href="https://cryptoradar.co">Cryptoradar</a> for comparisons based on prices, fees and features.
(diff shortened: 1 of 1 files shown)


crwatkins:
I don't believe we have to remove Cryptoradar (cryptoradar.co) as a resource.  I have not personally verified the quality of the ratings, nor the prices on the site, but if they are accurate, this seems to be a valuable resource for users attempting to choose an exchange.  The existence of other cryptocurrency listings on the site does not seem to detract from the quality or the presentation of the Bitcoin information (in fact, Bitcoin is listed first and is preloaded into the search selector).

I'm OK with the other revision to the string.

wbnns:
@crwatkins Ok, thanks. Will place this on hold until we get more feedback.

wbnns:
Closing since no additional feedback was received.

PR #3024, opened 2019-06-21
Update exchanges.html

nexwave-mat:
Adding Bull Bitcoin for Canada.


jb55:
Did you just add your referral codes?

wbnns:
@jb55 Seem so.

PR #3031, opened 2019-06-27
Rework exchanges page structure

josiasmontag:
The current [Bitcoin Exchanges page](https://bitcoin.org/en/exchanges) has been forked many times in recent years and grown a lot without paying too much consideration on site structure. In my opinion the page is a bit cluttered and there are multiple things that could be improved:
* The structure of the page is not consistent: main sidebar menu items are “International”, “P2P”, continents and even small countries like “New Zealand” (no offense); there is no logic to the order of menu items either (such as alphabetical listing).
* The structure within the sections is not consistent: exchanges in “North America” are listed by country-of-origin, whereas there is no sub-section in “Europe” for many exchanges (e.g. AnyCoin Direct should be listed as “Netherlands” if we follow a country-of-origin-approach).
* Many of the exchanges listed in local markets have been listed in their country of origin (e.g. itBit, Coinfloor, BitBay), but are available internationally; other exchanges have been listed multiple times based on availability in different countries (e.g. Bitpanda, Buda).
* There are additional platforms listed without a headline at the very bottom of the page.
 
Therefore, I would like to propose a new structure for the site, listing exchanges based on type and availability. Exchanges operating in more than one continent should be listed as “International”. Each exchange will only be listed once and in alphabetical order.

Proposed new structure:
* International Exchanges
* Local Exchanges
  * Africa
  * Americas
  * Asia
  * Europe
  * Oceania
* P2P Directories
* Review & Comparison Platforms

To accomodate the new structure, I moved a few exchanges which are available internationally from local sections to the “International Exchanges” section. Furthermore, the following sites were removed:
* Cryptobuyer - reason for removal: no information on supported countries and currencies available
* Bitcoin ATM Radar - reason for removal: this website is neither a Bitcoin exchange nor a comparison platform; the alternative of using ATMs to buy Bitcoin is already discussed at https://bitcoin.org/en/buy

Screenshot of the new structure:
image

### DIFF
--- a/_sass/screen.scss
+++ b/_sass/screen.scss
@@ -2484,100 +2484,14 @@ br.big {
     margin-top: 70px;
     font-weight: 400;
 }
-.toc-row .toccontent .exchanges-tab-title {
-    position: relative;
-    padding-left: 95px;
-}
-.toc-row .toccontent .expanded .exchanges-tab-title {
-    padding-left: 95px;
-}
-.exchanges-tab-title::before {
-    content: '';
-    position: absolute;
-    top: 40px;
-    left: 25px;
-    width: 40px;
-    height: 40px;
-}
-.exchanges-international-title::before {
-    background: url(../img/flags/ALL.svg?1528322191) center no-repeat;
-    background-size: contain;
-}
-.exchanges-europe-title::before {
-    background: url(../img/flags/EU.svg?1528322191) center no-repeat;
-    background-size: contain;
-}
-.exchanges-asia-title::before {
-    background: url(../img/flags/asia.svg?1528322191) center no-repeat;
-    background-size: contain;
-}
-.exchanges-south-america-title::before {
-    background: url(../img/flags/south-america.svg?1528322191) center no-repeat;
-    background-size: contain;
-}
-.exchanges-north-america-title::before {
-    background: url(../img/flags/north-america.svg?1528322191) center no-repeat;
-    background-size: contain;
-}
-.exchanges-australia-title::before {
-    background: url(../img/flags/australia.svg?1528322191) center no-repeat;
-    background-size: contain;
-}
-.exchanges-newzealand-title::before {
-    background: url(../img/flags/newzealand.svg?1528322191) center no-repeat;
-    background-size: contain;
-}
-.exchanges-africa-title::before {
-    background: url(../img/flags/africa.svg?1528322191) center no-repeat;
-    background-size: contain;
-}
-.exchanges-p2p-title::before {
-    background: url(../img/flags/ALL.svg?1528322191) center no-repeat;
-    background-size: contain;
-}
-.expanded .exchanges-content-row {
-    display: -webkit-box;
-    display: -ms-flexbox;
-    display: flex;
-    -webkit-box-pack: justify;
-        -ms-flex-pack: justify;
-            justify-content: space-between;
-}
-.general-marketplace {
-    width: 240px;
-}
-.expanded .marketplace-row {
-    display: -webkit-box;
-    display: -ms-flexbox;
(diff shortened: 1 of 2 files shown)


Cobra-Bitcoin:
I think we should list international exchanges in order of volume and respectability, maybe with a small paragraph underneath each one describing it. Then for the local exchanges we can have the user select a country from a drop down, and then we can have appear in a table or something any local exchanges that service that country. I like your solution but it also in some ways looks more cluttered and like a long list.

josiasmontag:
@Cobra-Bitcoin I agree with your point that a description of the international exchanges would be nice.

However, building such a country dropdown would require a lot of logic. These European exchanges would require that you store all European countries (sometimes exemptions) and so on. Building and maintaining this logic is a bit out of scope of this rather simple page. Furthermore, sites like BuyBitcoinWorldWide.com and Cryptoradar do a pretty good job in showing the available exchanges for your country - and this is why the page mentions them in the "Comparison platforms" section.

wbnns:
@josiasmontag 

Hello, thank you for taking time to suggest improvements to the page. Regarding:

> The structure of the page is not consistent: main sidebar menu items are “International”, “P2P”, continents and even small countries like “New Zealand” (no offense); there is no logic to the order of menu items either (such as alphabetical listing).

The sections aren't listed alphabetically, because International exchanges (currently listed first), would receive more traffic and be applicable/relevant to more users visiting the site. Second to that, P2P exchanges are listed to encourage their use. Agree that we should consider the ordering, however. Would still recommend International and P2P, followed by an alphabetical listing of the subsequent categories / continents. We can also combine Australia and New Zealand as Oceania.

> The structure within the sections is not consistent: exchanges in “North America” are listed by country-of-origin, whereas there is no sub-section in “Europe” for many exchanges (e.g. AnyCoin Direct should be listed as “Netherlands” if we follow a country-of-origin-approach).

This is because if an exchange is in Europe, it can be opened to many users due to being in the EU/EEA (thereby leveraging SEPA transfers), whereas with exchanges in North America, this would not be the case unless the user was in the same country as the exchange (e.g. Canada vs Mexico vs United States). We should clean up the European category, though.

> Many of the exchanges listed in local markets have been listed in their country of origin (e.g. itBit, Coinfloor, BitBay), but are available internationally; other exchanges have been listed multiple times based on availability in different countries (e.g. Bitpanda, Buda).

This is because some exchanges are available in multiple countries but the countries they are available in still do not apply to many users who would be visiting the Exchanges page, going to exchanges in the International category. (e.g. Buda is available in three countries in South America but would not be able to service most "International" customers to appear in that section.

> There are additional platforms listed without a headline at the very bottom of the page.

They are not headlined because they are not exchanges, and the main goal of the page when it was created was to quickly help direct people to an exchange in their country, not to send them to another website to browse exchanges by country.

**Regarding the screenshot:**

We agree that we can improve the way exchanges are listed within each continent, however, the proposed changes add unnecessary repetition to the page (showing the same country flags multiple times, with the text "available in" and "throughout the" written over and over). Also, when users come to the Exchange page, many aren't familiar with the names of the exchanges. In this example, the exchange names are listed alphabetically, whereas the countries are not, which would cause users to need to dig through each continent to find where in a list of exchanges they can find ones in their country. In Asia, exchanges for South Korea are listed in the beginning, middle and the end of the list of exchanges. In the Americas, a user in Mexico would have to look back and forth through the list of exchanges to dig out the exchange he/she could use in Mexico. This in our opinion is more difficult to parse than the current layout.

This all said, however, we are going to open a new issue so we can continue to improve the Exchanges page based on your suggestions, as well as what @Cobra-Bitcoin has mentioned. Please feel free to continue to help and contribute to discussion, there.

PR #3036, opened 2019-06-30
Reflect correct transaction version type

amoskyler:
Per bitcoin source code, the transaction version is of type `int32_t`. Documentation currently lists it as a `uint32_t`.

bitcoin source: https://github.com/bitcoin/bitcoin/blob/c65c77c721b69f1782509145d62978d31c9a1d78/src/primitives/transaction.h#L289


wbnns:
@amoskyler Thank you.

PR #3042, opened 2019-07-07
Add Corazon hardware wallet

DarrenSG:


crwatkins:
Hey @DarrenSG Thanks for the submission.

Be sure to check out the [HSTS criteria here](https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/docs/managing-wallets.md#wallets) to make sure gray.sg meets them.  The easiest way to test for best practices is to use https://hstspreload.org/?domain=gray.sg

DarrenSG:
Hey @crwatkins, thanks for the feedback.

From the requirement I see
"Website serving executable code or requiring authentication uses HSTS with a max-age of at least 180 days"

- The site Gray.sg is HSTS enabled, but the max-age is 90 days, it's hosted with Shopify and there is no way to modify that header currently.
- Gray.sg is only the website to purchase the device, the device will not have any interactions with gray.sg

crwatkins:
Thanks for the update @DarrenSG 

> * The site Gray.sg is HSTS enabled, but the max-age is 90 days, it's hosted with Shopify and there is no way to modify that header currently.

Thanks for the information; 90 days is such a curious max-age that I've never seen before on a site.  I'm going to infer from the [Shopify help center](https://help.shopify.com/en/manual/domains/managing-domains/ssl) that they do this to reduce the pain if their customers move off Shopify to a non-SSL service.  However, that page also states that 90 days is the default and implies it can be changed (which I did not verify).  For more than one reason, including being able to add yourself to the HSTS preload list (which is currently optional, but may become a listing criteria in the near future), you may want to contact Shopify to find out how to change the default.

> * Gray.sg is only the website to purchase the device, the device will not have any interactions with gray.sg

Our goal is is to maintain a chain of trust from bitcoin.org, which will link to your site, to the downloaded code (whether it be a companion wallet app, or new firmware updates), or to the site that links to this code.  If you didn't have links to trezor.io on your site, I could probably argue that this requirement doesn't strictly apply to you (although it still would be a very good idea).

DarrenSG:
Thanks for the reply and explanation @crwatkins 

I have contacted Shopify support to update our HSTS max-age to 180 days.
Unfortunately the HSTS preloading is currently not supported by Shopify.

crwatkins:
@DarrenSG Thanks for the update.  As I mentioned above

> Our goal is is to maintain a chain of trust from bitcoin.org, which will link to your site, to the downloaded code (whether it be a companion wallet app, or new firmware updates), or to the site that links to this code.

The site trezor.io, which you link to, should also have HSTS enabled, but it does not.  Could you have them correct that?

DarrenSG:
Isn't trezor.io already listed on the website?

Why don't they have to follow the HSTS requirement?

crwatkins:
@DarrenSG 
> Isn't trezor.io already listed on the website?
> Why don't they have to follow the HSTS requirement?

Trezor (a.k.a. Trezor One) is listed with a reference to the trezor.io web site.  When they "recently" switched their domain to trezor.io @prusnak requested some time to get the new site in order.

_wallets/corazon.md
```diff
@@ -0,0 +1,26 @@
...
+compat: "hardware"
+level: 2
+platform:
+  - hardware:
+    name: hardware
+    os:
+      - name: hardware
+        text: "walletcorazon"
+        link: "https://gray.sg/collections/corazon-wallet"
+        source: "https://github.com/trezor/"
```
wbnns: This should link to the open source repository corresponding to the firmware, instead of the GitHub overview page for the Trezor organization.

wbnns:
Closing this issue, since it has been approximately five weeks following our request to update this PR with the correct source code link (still pending), and also in lieu of the sentiment in #3060 that wallets should be functionally unique compared to other listings. As this device is the same as a Trezor Model T, albeit with a different encasement, this would not adhere to the definition.

If other contributors disagree with this conclusion, we can reopen this issue for further discussion.

Issue #3032, opened 2019-06-28
Clarify wallet listing requirements for code being open source / publicly available


wbnns:
We noted that at least a couple of wallets listed on the site do not have code that is open source and/or publicly available:

+ [BitGo](https://bitcoin.org/en/wallets/web/bitgo/) - no link to a code repository is present in the [wallet metadata](https://raw.githubusercontent.com/bitcoin-dot-org/bitcoin.org/master/_wallets/bitgo.md).
+ [BTC.com](https://wallet.btc.com) - [code repository](https://github.com/blocktrail/blocktrail-webwallet) links to [Blocktrail](https://www.blocktrail.com), which has a notice that they have ceased operations. The wallet at wallet.btc.com doesn't appear to share the same codebase. For example, it has a buy bitcoin feature which isn't present in the GitHub repository for the aforementioned web wallet by Blocktrail.

Since wallet providers are being instructed that their codebase must be publicly available, we wanted to clarify how this requirement should pertain to existing wallets on the site.

crwatkins:
> Since wallet providers are being instructed that their codebase must be publicly available, we wanted to clarify how this requirement should pertain to existing wallets on the site.

I can comment more on this later, but I believe that we expect to have source code links for web wallet code that handles private keys.  On the two issues mentioned, I would like to clarify:

>BTC.com - code repository links to Blocktrail, which has a notice that they have ceased operations. The wallet at wallet.btc.com doesn't appear to share the same codebase. For example, it has a buy bitcoin feature which isn't present in the GitHub repository for the aforementioned web wallet by Blocktrail.

We have a source code link in the web btc.com wallet listing.  Are you stating that the code we link to is not the current code being used for the wallet?

> BitGo - no link to a code repository is present in the wallet metadata.

I believe "we" (i.e. not BitGo) removed the source link when we dropped the chrome extension from the BitGo listing.  We should ask for the correct link.

wbnns:
@crwatkins 

> We have a source code link in the web btc.com wallet listing. Are you stating that the code we link to is not the current code being used for the wallet?

Yes, the wallet at wallet.btc.com doesn't appear to share the same codebase as the source code link. For example, it has a buy bitcoin feature which doesn't seem to be present in the [GitHub repository](https://github.com/blocktrail/blocktrail-webwallet).

crwatkins:
[...]

PR #3045, opened 2019-07-09
Update bitgo page to add source for wallet library

bencxr:
As per https://github.com/bitcoin-dot-org/bitcoin.org/issues/3032, update the bitgo entry to include a link to our source

### DIFF
--- a/_wallets/bitgo.md
+++ b/_wallets/bitgo.md
@@ -15,6 +15,7 @@ platform:
         text: "walletbitgo"
         link: "https://www.bitgo.com/"
         screenshot: "bitgo.png?1528322191"
+        source: "https://github.com/bitgo/bitgojs"
         check:
           control: "checkpasscontrolmulti"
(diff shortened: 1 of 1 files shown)


wbnns:
@bencxr 

Thank you for the PR. Based on the [conversation](https://github.com/bitcoin-dot-org/bitcoin.org/issues/3032#issuecomment-509718509) in #3032, not sure that an open source library interacting with a closed source wallet would satisfy the open source requirement for listing.

CC: @crwatkins

crwatkins:
@bencxr I agree with @wbnns that we need a link to the repository for the web client that handles private keys.  In most cases this would be the JavaScript/HTML/CSS that runs in the user's browser.  In the past, in the case of BitGo, this was the Chrome client's source code [here](https://github.com/BitGo/bitgo-chrome) which was deprecated in 2017 and removed from the bitcoin.org listing in #2024.  Can you provide the source code for the current client?

PR #3048, opened 2019-07-15
Add guidance for wallet descriptions

crwatkins:
Currently we provide no guidance for developers submitting wallet descriptions for listing and recently we have been getting descriptions that seem too long.  Without some limit, we will have descriptions that don't currently lay out well, or may not lay out well in the future.  Also, longer descriptions are obviously more verbose and are likely not in the style of the other descriptions.  Currently we have no descriptions longer than 320 characters which I propose as our limit (arbitrary sum of two powers of two).

@natiwa @alexcherman Any thoughts on the layout considerations?

When this is merged, perhaps we could ask @harding to update the schema checker.

### DIFF
--- a/docs/managing-wallets.md
+++ b/docs/managing-wallets.md
@@ -92,7 +92,7 @@ with `optipng -o7 file.png`. The icon must fit within 96 X 96 px inside the png,
 or 85 X 85 px for square icons.
 
 **Description**: The text must go in `_translations/en.yml` alongside other
-wallets' descriptions.
+wallets' descriptions.  The description must be less than 320 characters and be consistant in style with the other descriptions.
 
 **Level**: Each wallet must have a level property assigned. A value must be in a range
 between 1 and 4. Level represents a category of a wallet:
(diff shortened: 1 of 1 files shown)


wbnns:
@crwatkins 

As part of this, am wondering if we should also consider adding guidance for the type of phrasing that can be used in the descriptions. For example, at least a few wallets include phrases like:

+ "is one of the easiest wallets"
+ "ultra-secure"
+ "secure even when using a compromised computer"

These statements are subjective, and to a site visitor, might come across as being some kind of editorial-like type of endorsement from bitcoin.org, when they are actually just the description that the vendor supplied.

harding:
@wbnns FYI, previous phrases like that were removed, so there's precedence for that (I actually thought it was in the wallet policy not to allow superlatives, but I don't see it).  See https://github.com/bitcoin-dot-org/bitcoin.org/pull/543

wbnns:
@harding Ok, thanks for mentioning.

crwatkins:
@harding I agree.  I've updated the instructions to include this.  It will definitely help as I have been doing an ad hoc job on each new wallet listing as it is submitted to try to maintain this currently unwritten policy.

wbnns:
Unless others object, this is scheduled to be merged on Monday, July 29th.

Issue #1341, opened 2016-08-17
Incorrect protocol version documented for version.relay.


evoskuil:
[Bounty](https://github.com/bitcoin-dot-org/bitcoin.org#earn-bitcoin-for-contributing): **25,000 bits / ~$250 USD**

According to other bitcoin.org [documentation](https://bitcoin.org/en/release/v0.9.0) and observation of the network, `version.relay` was not enabled in Bitcoin Core until protocol version 70002, but is [documented](https://bitcoin.org/en/developer-reference#version) as version 70001:

> Added in protocol version 70001 as described by BIP37.

evoskuil:
Reviewing the sources shows that `version.relay` is read **but not written** in version 0.8.x (70001) and it is read **and** written in version 0.9.0 (70002).

Mirobit:
Agreed it doesn't work with version 70001.

evoskuil:
To be clear, it half-works, which can be a bit confusing.

wbnns:
@evoskuil @Mirobit Thanks for bringing this up - what do you all propose to resolve this issue?

evoskuil:
@wbnns There is proper Bitcoin Protocol behavior and there is the Bitcoin Core implementation. Assuming you are documenting the former I would document the proper behavior according to the BIP. Even so it might be worth a note on the bug since it needs to be worked around for useful operation against these older nodes. I've found a few of these types of issues in various clients because libbitcoin has configurable min and max protocol levels.

wbnns:
[Bounty](https://github.com/bitcoin-dot-org/bitcoin.org#earn-bitcoin-for-contributing): **5,000 bits / ~$10 USD**

achow101:
The field is optional. It does not need to be set even if the protocol version is 70001 or higher.

evoskuil:
@achow101 Agree, it being optional makes it unimportant that Bitcoin Core reads it in 70001 but doesn't write it until 70002. I had been operating in consideration of BIP60, which makes it mandatory. Since BIP37 is "final" and BIP60 is "draft", your interpretation is correct.

evoskuil:
Just for the record I would point out that the `version.relay` field is not valid at any protocol level prior to 70001. I mention this because I do encounter the occasional peer that sets it at these lower levels.

evoskuil:
[...]

PR #3074, opened 2019-08-18
Clarify note about relay field

cornelius:
Clarify the note about the relay field to include a description of the exact issue with the incompatibility with older protocol versions as discussed in issue https://github.com/bitcoin-dot-org/bitcoin.org/issues/1341


wbnns:
This is scheduled to be merged on Wednesday, August 21st.

PR #3041, opened 2019-07-03
Add Zebpay to the list of Australia exchanges

nirajd-zebpay:

PR #3082, opened 2019-08-27
Add Zebpay to the list of Australia exchanges

nirajd-zebpay:
Request to list Zebpay in Australia section of List of exchanges page (https://bitcoin.org/en/exchanges#australia).

### DIFF
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -385,6 +385,8 @@ id: exchanges
     <a class="marketplace-link" href="https://www.hardblock.net/">HardBlock</a>
     <br>
     <a class="marketplace-link" href="https://www.independentreserve.com/">Independent Reserve</a>
+    <br>
+    <a class="marketplace-link" href="https://www.zebpay.com/au/">Zebpay</a>
   </p>
 </div>
 


wbnns:
@nirajd-zebpay 

Hello, we've attempted to review the exchange again, however, this time we've encountered a different issue. When signing up, a message appears informing us that you all are "currently not accepting new registrations":

image

Due to the continued issues, combined with what we already mentioned in #3041, we wouldn't be able to recommend Zebpay for listing at this time. We recommend that a period of 12 months pass, after which, we can review for potential inclusion again.

If other active contributors disagree with this conclusion in the interim, we can reopen this PR for additional consideration.

nirajd-zebpay:
Hello @wbnns - This error message has been triggered because the number in use is not an Australian number. Are you certain that you've used an Australian number?

Please note that at the moment we are only accepting registrations from Australia - hence the error message that you see.  This is the reason for the issues in #3041 as well.

wbnns:
@nirajd-zebpay 

Hello. As mentioned in #3041, the user interface accepts numbers from over 200 countries. This would be problematic if you all were to be added under Australia, because not everyone in Australia uses an Australian mobile number. People who don't would follow the link, try to sign up with their preferred mobile number, and see a message that you all aren't accepting new registrations.

Issue #1141, opened 2015-11-14
Many of the torrent trackers are broken


laanwj:
I always seed the torrent. There have been some peers. But my torrent client tells me:
- udp://tracker.publicbt.com:80 Announce error: Connection failed - Today 06:13:12 AM
- udp://tracker.ccc.de:80 Announce error: Connection failed - Today 06:13:28 AM
- udp://open.demonii.com:1337 Announce error: Connection failed - Today 06:12:56 AM

udp://tracker.openbittorrent.com:80 and udp://tracker.coppersurfer.tk:6969 are working.

Don't know enough about bittorrent if this is a problem with the torrent file, or with my client, but I heard it from someone else on IRC too.

saivann:
Thanks, agreed! If someone can come up with a good suggestion for better open trackers, I can take care of updating the configuration and scripts.

Having dead trackers does not make any harm though, that I know, but it seems clear that some of them have been down for way too long to be of any value.

theymos:
torrentfreak.com recently [listed](https://torrentfreak.com/demonii-torrent-tracker-shuts-down-for-good-151116/?utm_source=feedburner&utm_medium=feed&utm_campaign=Feed%3A+Torrentfreak+%28Torrentfreak%29) three alternative trackers after demonii.com shut down: 

*  udp://tracker.openbittorrent.com:80
* udp://tracker.coppersurfer.tk:6969
* udp://tracker.leechers-paradise.org:6969

Might be nice to add these for the next release.

saivann:
@theymos Thanks, I replaced demonii by leechers-paradise for now, not sure if we should drop the other two ones (openbittorrent was down for a while but eventually recovered for example).

wbnns:
Just wanted to follow up on this - is it still a concern? If not, will close. FWIW, I just downloaded Core using the tracker and it seemed like it worked well:

![image](https://cloud.githubusercontent.com/assets/1130872/22736481/ef8ff564-edc3-11e6-89e4-f493c0d4bb0d.png)

laanwj:
No, this is no longer a concern, had forgot about this a bit.

wbnns:
@laanwj Ok, great, thanks for the update.

Sjors:
Some of these trackers are (still) dead, see #3087. However the remaining trackers are still good enough.

PR #3087, opened 2019-08-29
Remove disappeared torrent trackers

Sjors:
The following trackers are dead:

* `publicbt.com`: unreachable, since at least 2016: https://gist.github.com/mcandre/eab4166938ed4205bef4#gistcomment-1687275 
* `tracker.ccc.de:80`: same as above
* `open.demonii.com:1337`: unreachable, permanently closed in 2015: https://torrents.io/site/demonii

I tested that the other trackers still work as of today.

See also #1141 (@laanwj).

Not included in this PR, if people are interested, I setup my own tracker at `udp://tracker.bitcoin.sprovoost.nl:6969`. It's whitelisted, so shouldn't suffer the same fate as these folks: https://torrentfreak.com/running-a-torrent-tracker-for-fun-can-be-a-headache-160828/

Currently it tracks the 0.18.1 and 0.17.1 release torrents. I'm happy to keep whitelisting new releases.

### DIFF
--- a/_build/update_torrent.sh
+++ b/_build/update_torrent.sh
@@ -23,7 +23,7 @@ version=''
 for f in `find "$BINDIR" -maxdepth 1 ! -path "$BINDIR"`; do
 
 	f=${f##*/}
-	
+
 	# Ignore directories that don't end with a version number.
 	if [[ $f =~ [^0-9]$ ]]; then
 		continue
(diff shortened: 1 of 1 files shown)


wbnns:
@Sjors 

Thank you for removing the dead trackers and testing the two that remain. We yield to waiting to see what other people's opinions would be regarding adding another tracker and would recommend that it first be discussed as an issue, if you'd like to see the change.

PR #3103, opened 2019-09-11
Cleanup unused wallet assets

wbnns:
This removes media that was present in `img/screenshots` and `img/wallet`, associated with wallets that are no longer on the site, and will be merged once tests pass.


crwatkins:
ACK

PR #3106, opened 2019-09-12
Add optional indicator to wallet selection wizard

wbnns:
This adds a note under the Criteria and Feature selection portions in the new wallet selection wizard, so that people know they don't have to select criteria or features that are important to them, if they don't want to. This also covers a use case where people might not necessarily be sure which criteria or features are important to them.

### Screenshots

__Desktop__:

image

__Tablet__:

image

__Handheld__:

image


wbnns:
Unless others object, this is scheduled to be merged on Monday, September 17th.

PR #3107, opened 2019-09-14
Add InstaEx Exchange

instaex:


wbnns:
@instaex 

Hello, when reviewing the exchange, we noted that it is not possible to use fiat, which would be a requirement in order to be potentially included on the page. In the future, should a C2C be added, you all are welcome to open a new pull request for potential inclusion, there.

On a separate note, we noted that the Currencies page on your site is broken in Safari 13:

image

instaex:
@wbnns We are working on fiat integration and it should be live very soon :)

And thanks for reporting the issue on currencies page.

PR #3112, opened 2019-09-17
Require bug reporting method

crwatkins:
A bug reporting method is currently an optional criterion.  This PR makes it required.
All listed wallets have bug reporting methods.

(Another criterion was moved in the list to group similar criteria.)

### DIFF
--- a/docs/managing-wallets.md
+++ b/docs/managing-wallets.md
@@ -21,6 +21,7 @@ in the future
 - No indication that changes to the code are not properly tested
 - Wallet was publicly announced and released since at least 3 months
 - No concerning bug is found when testing the wallet
+- Provides a bug reporting method on the website and in app
 - Website supports HTTPS and 301 redirects HTTP requests
 - SSL certificate passes [Qualys SSL Labs SSL
   test](https://www.ssllabs.com/ssltest/)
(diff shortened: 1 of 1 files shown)


wbnns:
@crwatkins 

Just to confirm, all of the listed wallets currently have a bug reporting method in their apps in addition to their websites? It looks like the previous optional criteria was that they only have a policy listed on their respective websites.

docs/managing-wallets.md
```diff
@@ -21,6 +21,7 @@ in the future
 - No indication that changes to the code are not properly tested
 - Wallet was publicly announced and released since at least 3 months
 - No concerning bug is found when testing the wallet
+- Provides a bug reporting method on the website and in the app
```
wbnns: We should determine what qualifies as a bug reporting method in order for the requirement to be satisfied and to provide clarity to the wallet provider. For example, whether it's a specific email address that's dedicated to bugs, a link to a bug bounty / responsible disclosure program, or something else.

crwatkins: We have been accepting anything that allow contacting the developer and I'm OK with that.  What do you think?

crwatkins:
@wbnns I'm not positive all apps have it, so I changed to it and/or in 3536a0f.

wbnns:
This is scheduled to be merged on Wednesday, September 25.

docs/managing-wallets.md
wbnns: Since we've revised the requirement to "and/or" (app/website) for the method, SGTM; thinking we can just review on a case by case basis. Was previously concerned that if it needed to be in both places it might not be clear exactly what we are looking for in order to satisfy the requirement (it sounded a bit more stringent if it had to be in all places).

PR #3114, opened 2019-09-19
Add BitcoinVN to list of exchanges

sonofliberty:
This PR adds [BitcoinVN](https://bitcoinvn.io/en) as the first Vietnamese exchange to the list.

BitcoinVN


wbnns:
@sonofliberty 

Hello, thank you. When reviewing the exchange we noted the following issues:

1. When purchasing bitcoin, users are instructed to make deposits to what appears to be an individual's bank account, rather than the exchange. This is also a different name compared to the company name mentioned in the site terms of service. Due to international KYC/AML-related procedures, this would put the exchange at high risk for having its bank accounts closed and/or frozen due to the use of an individual account being used as a money services business.

2. The exchange advertises trades being subject to a 0.2% exchange fee. When attempting to execute a trade, we noted that the spot price being offered by the exchange, 211,122,460 dong ($9,107.08 USD), is 6.9% higher than the current global average spot price of $8,478.48 USD.

Based on the above issues, we wouldn't be able to recommend including the exchange at this time. If other contributors disagree with this conclusion, we can reopen the PR.

sonofliberty:
Fair enough, thanks for the feedback!

PR #3128, opened 2019-09-28
wallet-selector: Do not override anchor link style

wbnns:
This ensures that the wallet name that appears in the wallet selector
results table, which is an active link, isn't overridden to appear as
black text, potentially making it more difficult to discern as something
that can be clicked on or touched to navigate to an individual wallet's
landing page:

image

Cc: @crwatkins


crwatkins:
Concept ACK.  Thanks!

PR #3129, opened 2019-09-30
Remove optional security audit wallet criteria

crwatkins:
This PR is to remove the optional wallet security audit criterion.   Note that we still have a required criterion that specifies

* Sufficient users and/or developers feedback can be found without concerning issues, or independent security audit(s) is available

such that a security audit is an acceptable alternative to sufficient feedback.

Over the last few years, our optional criteria section has served as a place to list criteria that we think are good ideas and we want to convey that to the community and encourage developers to consider them, while not precluding any wallets that have not yet met the optional criteria.

I believe that security audit is certainly a good idea, but I don't believe we will be making it a requirement in the foreseeable future.    One reason is because of the difficulty and the cost to developers.  However the major reason is that I don't think that we can reasonably specify the definition and the scope of an audit to the point that this requirement is actually meaningful.
For that reason, and for the reason above that it is already mentioned elsewhere in the requirements, I recommend removing it from the optional section.

### DIFF
--- a/docs/managing-wallets.md
+++ b/docs/managing-wallets.md
@@ -58,7 +58,6 @@ recovery process.
 
 Optional criteria (some could become requirements):
 
-- Received independent security audit(s)
 - Does not show "received from" Bitcoin addresses in the UI
 - Website serving executable code or requiring authentication is included in the
   [HSTS preload list](https://hstspreload.org/)


wbnns:
It would be nice if we could come to some sort of standard for requiring third party reviews of wallets in order to be listed, especially since we're talking about products and services that are storing people's money. Not sure the best way to get to that.

In the interim, since it's optional, maybe we could leave as-is and see what kinds of third party reviews wallets might potentially provide, rather than discarding the mention altogether. It might yield something useful that we could potentially apply to reviews of other wallets, even if we don't have a hard requirement.

crwatkins:
> It would be nice if we could come to some sort of standard for requiring third party reviews of wallets in order to be listed, especially since we're talking about products and services that are storing people's money. Not sure the best way to get to that.

We would have to delist every software wallet.  None of the listed software wallets have provided a third party review (unless you count Edge, which had a review of their since-removed Airbitz wallet).

> In the interim, since it's optional, maybe we could leave as-is and see what kinds of third party reviews wallets might potentially provide, rather than discarding the mention altogether. It might yield something useful that we could potentially apply to reviews of other wallets, even if we don't have a hard requirement.

It hasn't happened in five years of trying that.  As a matter of fact, the only wallet submitted that claimed to have such a review in recent history is not listed because of a serious security issue which we found. 

By curating our optional list (moving some criteria to required, and removing others) I hope to achieve a realistic and useful list of optional criteria that we actually expect to be required at some point.  I don't realistically ever expect this to be a required criterion.

wbnns:
My recommendation would be to leave it as optional, rather than to assert that it wouldn't ever be useful material that could be submitted and reviewed alongside other things to determine if a wallet should be listed.

Do you still want to remove it?

crwatkins:
Since we have a required criteria that references independent security audits, I believe that should be sufficient.  Current reviews have to address the issue twice (once in each section) giving too much significance to an optional criteria that no one supports, so I recommend removing it.

wbnns:
@crwatkins Ok.

Unless others object, this is scheduled to be merged on Monday, October 7th.

PR #3133, opened 2019-09-30
Update exchanges.html - Add Coinmate to your list

vrsasi:
Thanks for adding Coinmate

### DIFF
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -139,6 +139,8 @@ id: exchanges
 
     <div class="general-marketplace">
       <p>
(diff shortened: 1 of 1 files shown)


syaifudin857:
Join

wbnns:
@vrsasi 

Hello, when reviewing the exchange we noted that there is no address validation when attempting to withdraw funds to a bitcoin address:

image

Please let us know when this is resolved and we will resume the review.

vrsasi:
@wbnns hi, if you would try to click on withdraw it would have informed you that your address is in incorrect format. So there is validation of the address

PR #3139, opened 2019-10-14
Add Binance.US to United States exchanges

JagerGizmo:
Binance.US was recently released using technology licensed from Binance - Capable of the same speed and matching in security. 

It currently services most states in the US, with the exception of a few. However, they're expanding to new states regularly.

### DIFF
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -275,6 +275,8 @@ id: exchanges
       <div>
         <h3 id="united-states" class="no_toc">United States</h3>
         <p>
+          <a class="marketplace-link" href="https://www.binance.us/en">Binance US</a>
+          <br>
           <a class="marketplace-link" href="https://bitflyer.com/en-us/">bitFlyer</a>
           <br>
           <a class="marketplace-link" href="https://bittrex.com/">Bittrex</a>


wbnns:
@JagerGizmo 

Thank you for the update. In reviewing the new exchange, we noted the Binance licensee, BAM Trading Services, Inc. (DBA Binance US), is using a third party, [Prime Trust](https://primetrust.com/), to [handle all user funds](https://www.binance.us/en/prime-trust-linked-wallet-addendum) (both fiat and crypto):

image

...and that the owner / responsible party with control over funds is not Binance US, but Prime Trust. In reviewing this, we noted the following [waiver](https://primetrust.com/tos) on Prime Trust's website:

image

We were also unable to find any reference to the protections (e.g. FDIC insurance) alluded to in the first TOS screenshot excerpt above for Binance US (1.3).

We are inferring from the above that Binance US is a licensee of the Binance brand, and that neither Binance nor Binance US are responsible for user funds - that Prime Trust is. And that on Prime Trust's website, their terms of service seem to indicate they have no responsibility over their products and services which are used in association with External Sites.

Could you please elaborate on if any of these inferences are incorrect and provide further clarification?

JagerGizmo:
Hey @wbnns - Thanks for posting your concerns, i've spoken with that team and got some details together.

1. Prime trust is the fiat service provider for Binance.US - All fiat exchanges have a fiat service provider, prime trust is Binance.US'.

2. The ToU on Binance.US now states that fiat deposits through their fiat service provider (PT) all sit in FDIC protected institutions. All USD deposits are FDIC insured.

If there are any other questions, feel free to connect with me on telegram @jagerBNB or post directly here.

wbnns:
We've completed an in-depth review of Binance.us and have encountered issues that we feel would preclude it from being listed at this time. The most notable of them were that:

1. We tested out the fiat withdraw functionality using a fully verified account, but only received one of the withdraws (small withdraws, $12 and $36, respectively - the $36 withdraw was never received).

2. We opened a succinct and polite support ticket regarding the issue (the missing withdraw) and did not receive any reply and/or acknowledgement from the support team in a timely manner. We still have not received a reply after multiple days.

We recommend that there should be a waiting period of at least six months before being re-reviewed for listing as a US exchange option.

PR #3154, opened 2019-11-01
Add leading Thai exchange to exchanges section

ercumenteskar:
Adding Bitkub Thai exchange


wbnns:
We contacted this exchange multiple times but did not receive acknowledgements and/or replies in a timely manner (> 1 month).

We recommend that a review be deferred for at least six months.

PR #3159, opened 2019-11-07
Add Bitclude.com to Polish exchanges

wierzcholski:
Bitclude is first Polish crypto exchange which receive license 
for financial activities in Poland and is regulated under 
financial supervision commission in terms of FIAT operations.


wbnns:
We reached out to Bitclude in late December and again in mid January, but have not heard back.

Closing for now.

PR #3169, opened 2019-11-11
added Dasset to the list of NZ exchanges

sm-dasset:
We've resolved this issue and have sent a password reset request to the email address that was used to sign up using the front-end interface. You can also make a reset password request at https://dassetx.com/problem-logging-in if you would like to make the request again. If you would like for us to delete your account so that you can start the process from scratch, please let us know.


wbnns:
@sm-dasset 

Hello, we attempted to contact your exchange via email but never received a reply.

PR #3170, opened 2019-11-12
Added New Exchange to South African Listings

victor-act:
- Added AltCoinTrader to South African list.


felipelalli:
nACK improve the commit title message.

wbnns:
We contacted this exchange but never received a reply; closing.

PR #3185, opened 2019-11-19
Better Hebrew whitepaper translation

sha-265:
Thanks for adding a Hebrew translation of the whitepaper to bitcoin.org, but I think there's a better Hebrew version translated by Meni Rosenfeld back in 2012 ([as mentioned in BitcoinTalk](https://bitcointalk.org/index.php?topic=46624.msg764564#msg764564)). This translation has been used by the Hebrew community ever since.

After a quick flutter in the new version you uploaded, I found Meni's translation linguistically better. But maybe we should wait for more comments on this.

Meni's translation can be found on Meni's website, [Bitcoil](https://bitcoil.co.il/bitcoin-hebrew.pdf), and on [Satoshi Nakamoto Institute website](https://nakamotoinstitute.org/static/docs/bitcoin-he.pdf).


MeniRosenfeld:
Thanks. I'm generally a "why not both" kind of guy, and if it's possible to feature both translations on the webpage, I think it might be best. This would give due credit to both my original work, and Zeev for bringing to our attention that the Hebrew version was so far absent from the bitcoin.org website.

Furthermore, the translations have some differences between them. I've made some particular design decisions in my translations (notably, I tried to match Satoshi's precise wording as closely as possible, and I made an effort to stick to Hebrew words and avoid loanwords) - which I think best capture the spirit of the work, but other people might prefer different styles.

But if it has to be one or the other, I dare say - my personal stake on the matter notwithstanding - that I strongly support featuring my original translation.

First, considering that the whole point of the whitepaper is that it's the original text that started it all, I think that if we offer a translation, it is better to offer the original translation (or at least a version of it).

Second, my version is already widely circulated, in both international and Hebrew sources. In addition to the sources linked by sha-265, it has appeared in the Bitcoin wiki for years - https://en.bitcoin.it/wiki/Bitcoin_whitepaper. It was part of jimbobway's translation project and published accordingly. In Israel we've printed booklets containing the English and Hebrew versions of the whitepaper, there's a famous introductory Bitcoin book which features the translation, and it was featured in our 10th whitepaper anniversary event.

Pushing a new translation when there's already a well-known one feels disrespectful (even if everyone's intentions were honest).

The final question is regarding the relative quality of the two translations, on which I am not at liberty to comment. But perhaps others might.

wbnns:
@sha-265 Thank you.

@MeniRosenfeld Thanks for joining the thread. Do you have a non-PDF version (i.e. doc, odt) of the translation you made that you can share?

shesek:
> After a quick flutter in the new version you uploaded, I found Meni's translation linguistically better. 

+1, I have a copy of the [booklet](https://imgur.com/a/zFSBweM) with Meni's translation that I've read a few times over the years and I also find this translation better.

MeniRosenfeld:
> @MeniRosenfeld Thanks for joining the thread. Do you have a non-PDF version (i.e. doc, odt) of the translation you made that you can share?

Sure, here's a link to the odg (OpenOffice Draw) file - https://bitcoin.org.il/files/bitcoin-he.odg

Also, here's the most recent pdf (it is preferred over the one sha-265 previously uploaded) - https://bitcoin.org.il/files/bitcoin-he.pdf

ZeevBB:
When deciding to do the translation, I was not aware of Meni's previous work. The slightly different taste in my approach  is legitimate as there are cases where the meaning is not reflected in the word by word translation. Sometime I found it beneficial  to make a connection to the original term. If both versions can be handled, that is fine; otherwise It is fine with me that Meni's version will replace mine. Meni's work for Bitcoin Israel is well appreciated while I am, in comparison, a newbie. 

 

Regards,

 

Ze'ev

wbnns:
We'll open a separate pull request to include both versions.

ZeevBB:
Great, thank you.

 

Ze'ev

wbnns:
Hello. After additional review when drafting the page to accomodate two versions, we've opened #3336 for further discussion. Let's confirm next steps there. If after doing so, the decision is to go with Meni's, we can just reopen this pull request.

PR #3186, opened 2019-11-19
Clarify wallet private key criterion

### LINKED ISSUE #3173: Listing custodial Lightning wallets
For years now, we [have required](https://github.com/bitcoin-dot-org/bitcoin.org/pull/1706) user access to private keys for wallet listings and the community has been very adamant on the subject.  In early discussions when there were few Lightning Network wallets available, [I have argued](https://github.com/bitcoin-dot-org/bitcoin.org/pull/3000#issuecomment-499097115) both pro and con positions for listing custodial Lightning wallets.  As many more wallets have been released and since we recently [added a new selection filter](https://github.com/bitcoin-dot-org/bitcoin.org/pull/3017) that allows users to specifically select wallets listed with Lightning capability, it is probably time to revisit the topic.

1. The community has been very adamant on the custodial topic for on-chain wallets
1. There are two non-custodial LN wallets listed
2. Users can now come to bitcoin.org to specifically request LN wallets
2. There have been custodial [issues](https://github.com/BlueWallet/LndHub/issues/8) confirmed with LN wallets that are not expected with wallets based on consensus
2. Many, if not all, released LN wallets provide Lightning functionality as an obvious major feature of the application, not a hidden optional function

There have been discussions that by making LN wallets custodial, on-boarding and general user experience can be easily improved.  I think the same can be said for on-chain wallets, and those arguments have been previously rejected.   The point could also be made for using PayPal instead.

For these reasons, I'm leaning toward applying the "access to private key" criterion to major off-chain features (including LN) of wallets as well as on-chain features.  This seems quite consistent with what we do today with the concept of exchange features and whether the custodial exchange is a major feature of the wallet or not.

Thoughts and discussion are very welcome.

crwatkins:
Resolves #3173 with a clarification that can apply to layer 2 components, such as Lightning Network, as well as clarify how I believe exchange components have been handled for the last few years.  The new clarified wording of the criterion is 

* User has access to private keys for all major components of the wallet

### DIFF
--- a/docs/managing-wallets.md
+++ b/docs/managing-wallets.md
@@ -32,7 +32,7 @@ in the future
   in the wallet UI
 - Avoid address reuse by using a new change address for each transaction
 - Uses deterministic ECDSA nonces (RFC 6979)
-- User has access to private keys
+- User has access to private keys for all major components of the wallet
 - If private keys or encryption keys are stored online:
   - Refuses weak passwords (short passwords and/or common passwords) used to
     secure access to any funds, or provides an aggressive account lock-out


wbnns:
This is scheduled to be merged on Monday, November 22nd.

PR #3187, opened 2019-11-20
Add Australian exchange BTC Markets to exchange page.

JarrodCrane:


wbnns:
We've completed an initial review of the exchange and encountered no major issues that we felt would preclude it from being listed. This is scheduled to be merged on Monday, February 17th.

PR #3188, opened 2019-11-20
Add Newton to Canadian Exchanges 

MichelleNewton:
Newton is Canada's first no-fee cryptocurrency exchange: https://www.newton.co/. 
Here's a list of pricing, including BTC: https://www.newton.co/prices.html.


wbnns:
We haven't heard back from Newton exchange since speaking with them in late December, so we're closing this PR for now.

PR #3193, opened 2019-11-24
update uk.yml

pavlowd:
- Added new translations for Buy Bitcoin page 
- Added new translations for Choose Your Wallet page 
- Added missed donation variables translation
- replaced small to capital letter when mention person: "your" to "Your" and "you" to "You" it`s not rule, but I noticed that sometime it`s was used as with capital and other time with small. Made one "standard".
- changed "плата" to "комісія" ("плата" is payment and fee in same time - so it`s homonym, and it`s not ok translate "payment fee" as "плата плата").  "комісія" - widely used by banks and it`s ok


pavlowd:
will make new pull request with validated yml and correct description

PR #3198, opened 2019-11-25
Update resources.html

fromtimwithcode:
Added The Rise and Rise of Bitcoin to list of documentaries


wbnns:
@fromtimwithcode 

Hello, thank you for the suggestion, however, it seems this is only available to users in select countries:

image

The site receives many visits from people all over the world, so we wouldn't recommend potentially including this if a significant number of people aren't going to be able to access it.

PR #3201, opened 2019-11-29
Add coinmate to exchanges

vrsasi:
Hi, last time you canceled my pull request with no valid reason. It's true that you can type anything into btc address when making withdrawal, but when you try to confirm it, it wouldn't let you do so and coinmate would request correct btc address.


wbnns:
@vrsasi 

Hello, we have reopened #3133 for the potential inclusion of Coinmate.

PR #3207, opened 2019-12-03
Add CashApp to exchange list

nickycutesc:
CashApp is awesome!


wbnns:
We've made multiple attempts to contact this exchange regarding listing requirements, however, they've yet to reply / acknowledge receipt, so we're closing for now.

PR #3209, opened 2019-12-04
Add Bitfinex to exchange list

nickycutesc:
Adding Bitfinex to the exchange list


wbnns:
@nickycutesc 

Hello, thank you for the suggestion, however, we've previously decided not to add Bitfinex.

Related: #2990

PR #3218, opened 2019-12-12
Update spend-bitcoin.html with bitcoin businesses search system

BitcoinWide:


BitcoinWide:
Svg icon file is ready and attached

BitcoinWide:
Link to download svg icon file is ready. Check previous post.

jdabs:
@wbnns, what is the reason for the delay on this PR and nice icon?

BitcoinWide:
It is Christmas :)

wbnns:
@BitcoinWide @jdabs 

Hello, we reviewed the website and noted several issues, for example:

**Typos**

+ On the homepage, "Who accept Bitcoins?"
+ On the homepage, "Once somebody will start to accept Bitcoins on the Mars"

**Search**

+ Using the first category link on the homepage, we are taken to view "Exchanges". The first result is an OTC exchange in Montreal, Canada - which may not be relevant for most people. The second result is for 796 Exchange, which we were unable to visit because the website is down/does not respond. The third result is for an exchange that has a letter to all users on the homepage noting that it has ceased operations.
+ We navigated through other suggested categories via the homepage (e.g. food, hotels, atm), and searching for "new york food" returned results with the top three being results for places in Germany, Massachusetts and California. Searching for "new york hotels" returned results with the top three being two results in Italy, and another in Switzerland. Searching for "los angeles atm" returned top results for a nail salon, soap, and moving companies that accept bitcoin, and a bitcoin atm in Arizona. We were able to find a result for Los Angeles, but it was several more results further down the page.
+ Entering a physical address for a previous search result as a new search query, does not return it as one of the top matches in a new list of search results.
+ The Back button link that appears above the search dialogue box in the breadcrumbs, refreshes the same page with the same query, rather than taking the user back to the previous page (e.g. homepage).

**Map**

+ The map feature doesn't seem to work. We entered "new york" and no results were shown in the state of New York (including New York City). For states where there is a result using the search engine (e.g. [New Jersey](https://bitcoinwide.com/product-of-italy-39c4ed82-f4f0-496b-a55f-65cba86630b5), there is no result on the map.

* * *

Based on the above issues, we would recommend waiting at least a few months and then proceeding to review the site again for potential inclusion, to see how things improve.

If other contributors disagree or recommend otherwise in the interim, we can reopen.

Related: #3217, #3216

BitcoinWide:
Hello,

Thanks for your review.
Many of the mentioned issues are really reasonable, but there is few things I would like to clarify.

 > There is a two reasons why search results seems unrelevant during your review.

1. While we add new businesses and organizations to our database everyday, it is still very small compared to the databases of search systems you use daily to find businesses of your interest (let's say Google maps). 

2. At the moment, we do not recognize a geo relevant requests in user's queries at the home page (map mode works in a different way). 
That is why when you have entered "new york food" you have got a results with businesses containing "new" "york" and/or "food" in the business name and/or business category name.
The geo relevant requests recognition will be definitely implemented later. It is a quite difficult task as we are building BitcoinWide from the scratch. 
Meanwhile, there is a map feature to find local businesses.


Entering a physical address as a search query does not have sense for user, as user perform search with a query relevant to the service, business category or product, together with geo relevant appendix or without.




 > Map feature.
You have entered "new york" and didn't got any results because you shouldn't. 
If you are in New York, open a map and try to search for a "restaurant" instead. Of course, if your IP is in NY area or Geo location shared with website.

We are focused to make a user's expierience assimple as possible. Using map tool usually mean that the user is looking for a local business, product or service in his area. 
He will not look for a "new york" as it will not contain any exact query related to the search target. 
To be clear, when you are located in New York and you looking for a restaurant, you will enter "restaurant" or "restaurant new york", not a "new york" query. 

User can only search for a local business on the map, as online business with no address will not have any relation to the address or location. 
 
Our map feature determine visitor's city or region and return nearest results related to your query first of all. Because using a map is always means you need a local business instead of online.

Moreover, when you move or zoom a map to any area, you may press "Search in this area" button which will return results related to your search query but located in the area currently catched on your screen. 

Please keep in mind, we are working on geo relevant queries recognition as well. 


 > You should take into considiration a very important aspect of the problems we are trying to solve. Due to the nature of Bitcoin use cases, there is a different kinds of businesses accepting and working with cryptocurrency. While some of them are local business, another ones are stricly online oriented (VPN service for example). 


I would like to appreciate your comments as some of them came us across a thought of an important improvements. 

We will be happy to see your comments to my reply.
BitcoinWide Team

wbnns:
@BitcoinWide 

Thank you for your reply. We understand your perspective, however, we still recommend waiting at least a few months to see how the site changes beyond the status quo.

BitcoinWide:
We are currently working on a number of improvements which are also cover the issues mentioned by you. I believe the release will be published in a month or so. We will re-submit the site once done.

PR #3206, opened 2019-12-01
Update mSigna link

nickycutesc:
The original link for mSigna throws a connection error. Perhaps the website's certificate expired?

PR #3223, opened 2019-12-17
wallets: Drop mSIGNA

wbnns:
Closes #3206.


wbnns:
We're noting some build-related issues occurring with Travis CI causing tests to fail on this PR, which we're currently looking into. The issues don't appear to be specifically related to any changes contained here-in, but rather with the build environment.

crwatkins:
utACK.  This issue #3206 makes the documentation for the wallet unavailable, as well as access to the information on obtaining help.  In addition the info email address (listed on the inaccessible web site) has not responded.

PR #3227, opened 2019-12-18
Add NovaDAX to exchange list

Elsadeng1020:


wbnns:
We've completed an initial review of NovaDAX and encountered no major issues that we felt would preclude it from being listed.

NovaDAX is scheduled to be listed on bitcoin.org on Wednesday, March 4th.

PR #3231, opened 2020-01-01
removes facebook and twitter sharing links

priem19:
1. Social media sharing links are redundant; copy-pasting isn't that hard and doesn't fall victim to trackers.
2. Does the facebook logo really deserve space on bitcoin.org? After all their scandals? These sharing links serve facebook more as free advertisement than it serves the visitors' convenience.
3. It's aesthetically unpleasing; it breaks the symmetry and minimalistic look of the website.
comparison


crwatkins:
utACK.  I agree with this concept.

jdabs:
I agree. At least should not be in the header. The data collection practices also fly in the face of Bitcoin philosophy

priem19:
> I also believe Google analytics should be removed or replaced

I agree wholeheartedly.

laanwj:
Agree, I doubt anyone uses these to promote bitcoin or bitcoin.org anyhow

PR #3235, opened 2020-01-12
Add eToroX

etorox:


wbnns:
We tried signing up with eToroX from three different browsers, on three different devices, from three different locations in Europe, and were shown an error message each time stating "We have not been able to create your account," and that we should contact customer service.

In addition to that, we noted the sign up page has styling issues, where the icons are overlapping onto the words they correspond with:

Screen Shot 2020-01-31 at 17 35 53

Based on the above, we recommend that a review be deferred for at least six months.

PR #3237, opened 2020-01-16
Correct the inclusion height for locktime

Xekyo:
Bitcoin transactions locked to a block height `H` become valid at height
`H` and can be relayed, but they may only be included in a block of
height `H + 1`.

H/T: Pieter Wuille

Refs:
* https://github.com/bitcoin/bitcoin/blob/v0.19.0.1/src/consensus/tx_verify.cpp#L21L22
* https://bitcoin.stackexchange.com/q/92412/5406


wbnns:
@Xekyo 

Thank you.

PR #3253, opened 2020-01-27
Switch privacynetwork score of Bitcoin Wallet to checkpassprivacynetw…

schildbach:
…orksupporttorproxy

Bitcoin Wallet has supported Tor via the Orbot app for years. All you need to do is enable
VPN mode and select Bitcoin Wallet (amongst other apps).


crwatkins:
ACK That works for me.

PR #3257, opened 2020-02-02
update bitcoin version to 0.19.0.1

ethicnology:


laanwj:
Changes LGTM though I'm not sure the idea is to update this for every version, as it's only an example it might be better to substitute `x.x.x` or something.

PR #3270, opened 2020-02-12
Add wallet features

alexcherman:


_layouts/wallet-container.html
```diff
@@ -15,11 +15,44 @@ <h1 class="wallet-title">
...
+          <div class="information-icon">
+            <img src="/img/icons/information.svg" alt="information" onclick="showModal({{ feature }})" />
+            <div class="features-tooltip-overlay"></div>
+            <div id="{{ feature }}" class="tooltip features-tooltip">
+              <div class="close-btn" onclick="closeModal({{feature}})">×</div>
+              <div class="tooltip-title">
+                {{ feature }}:
+              </div>
+              <div class="tooltip-text">
+                Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et
```
wbnns: We need to update to show the tooltips.

wbnns:
@alexcherman 

Thanks for the updates. It seems there may be some sort of issue where a blank feature and tooltip are appearing, in some cases. We should remove it:

For example, it appears when looking at Bitcoin Core:

Screen Shot 2020-03-05 at 14 42 59

But is not present when looking at Bitcoin Wallet:

Screen Shot 2020-03-05 at 14 25 32

wbnns:
@alexcherman 

Thanks for fixing the issue with the missing feature. There's an additional customization that needs to be made, which is that features should be unique based on operating system.

For example, looking at the wallet metadata, [Bither wallet](https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/_wallets/bither.md) supports Legacy Addresses and SegWit for Android and iOS, but not Multi-Signature transactions (which is only available on Desktop). Currently, however, multisig is showing as a feature for all operating system landing pages for the wallet.

wbnns:
@alexcherman 

It appears there's a styling issue when viewing the features on desktop browsers. There are large amounts of space appearing in between each feature, causing the section to not match the design.

**Current state (incorrect):**

Screen Shot 2020-03-11 at 19 19 40

Screen Shot 2020-03-11 at 19 21 35

**Correct state:**

Screen Shot 2020-03-11 at 19 20 22

Please also note that the features section shouldn't take up the width of the entire hero.

wbnns:
@alexcherman 

Thanks for the updates. Looks close to ready, but there's a bug where the tooltip isn't displaying for the `2FA` feature.

wbnns:
@alexcherman 

Thanks, looks good. Let's also make an additional update to ensure that only one feature tooltip can be opened at a time.

wbnns:
@alexcherman Thanks for the additional update.

* * *

This is scheduled to be merged on Friday, March 20th.

PR #3273, opened 2020-02-14
Small P2P Network document fix

Coding-Enthusiast:
Headers in a `headers` message are fixed 81 bytes (80 bytes block header + 1 byte tx count equal to zero).


wbnns:
@Coding-Enthusiast 

Thank you.

PR #3295, opened 2020-03-17
Update Events Page

wbnns:
This revises the date for the upcoming Value of Bitcoin Conference, and removes two events that have been postponed due to coronavirus (Mallorca Blockchain Days & Lightning Hackday Barcelona).

This will be merged once tests pass.


wbnns:
We've updated this PR just now to also remove two events that have been postponed due to coronavirus.

PR #3333, opened 2020-05-12
Switch privacydisclosure score of Bitcoin Wallet to checkpassprivacydisclosurefullnode

schildbach:
Since version 8, an improved privacy mode of blockchain sync can be selected. It does
not set a filter and reads all blocks and transactions. This makes it privacy-equivalent
to a full node. Everyone with a high enough data allowance is recommended to use this
mode.

### DIFF
--- a/_wallets/bitcoinwallet.md
+++ b/_wallets/bitcoinwallet.md
@@ -27,6 +27,6 @@ platform:
           fees: "checkgoodfeecontrolfull"
         privacycheck:
           privacyaddressreuse: "checkpassprivacyaddressrotation"
-          privacydisclosure: "checkfailprivacydisclosurespv"
+          privacydisclosure: "checkpassprivacydisclosurefullnode"
           privacynetwork: "checkpassprivacynetworksupporttorproxy"
 ---


crwatkins:
Nice job Andreas, I like the new feature!

However, we score (internally we call it score; the current user facing term is "criteria") wallets based on the default behavior and settings of wallet (and not on potential capabilities -- more on that below).  Your "full node privacy mode" is not the default, for obvious reasons as most users will not be doing 10GB or so of traffic from their phones each month.   Also, our full node privacy description requires full node behavior for sending transactions as well as downloading blocks.

While on one hand, I think we should represent the default behavior in the scoring, I definitely see the need for a way to express features that may be available in the wallet.   That's one of the reasons that we recently created the binary "Features" list for wallets.  Wallets are marked as having specific features without requiring them to be the default, or even requiring them to be a commonly used behavior.  We tried to create categories of features that would apply to multiple wallets or that we anticipated would in the future.  Suggestions for new categories are welcome and appreciated.

schildbach:
What special sauce do full nodes apply to sending transactions?

crwatkins:
> What special sauce do full nodes apply to sending transactions?

A full node not only originates its own transactions, but also forwards transactions from others in the P2P network, making it harder to determine which transactions are originating from that node.  Using a method sometimes referred to as trickling, Bitcoin Core nodes introduce some random delays in forwarding to further obscure which transactions originate locally.   Dandelion (BIP 156) should further that protection.

schildbach:
Thanks for your review!

PR #963, opened 2015-07-14
Added Coin.Space wallet

nikashitsa:
Coin.Space HD Wallet is a free online bitcoin wallet, which you can use to make worldwide payments for free. It makes paying with bitcoins easy and secure available anywhere on your phone or desktop.

PR #3335, opened 2020-05-14
Update Coin wallet

nikashitsa:
This PR updates mobile screenshot and adds features SegWit, Bech32 to Coin wallet.  
In last version v2.19.0 we added SegWit and Bech32 address support (send / receive).


wbnns:
@nikashitsa Thank you.

Cc: @crwatkins

wbnns:
@nikashitsa 

We were reviewing this just now and it seems the mobile screenshot wasn't updated as mentioned, and that it was removed instead. Could you please check?

nikashitsa:
@wbnns Ok, all looks good. Please use rich diff to view the difference.  
image

wbnns:
@nikashitsa 

Great, yes, I see it now. Thanks.

crwatkins:
@nikashitsa Congrats on the segwit upgrade!

It seems that your bech32 addresses are derived from a path of m/0'/0, the same as your legacy P2PKH, and the same for P2WPKH-P2SH addresses.   Is that correct?  Is there a reason that you didn't use  BIP49 and BIP84 paths?

nikashitsa:
@crwatkins Thank you!  
Yes, that's correct. It was just simpler to implement. Also, using single path simplifies life for users who want to export their private keys from our wallet (lower number of private keys).

crwatkins:
The way the receive address generation seems to be implemented is such that HD derived keys are reused for both bech32 and p2sh segwit addresses.  For example, a wallet with a long ago used a P2PKH address based on the private key for m/0'/0/0 reuses that index (and public key) when a new bech32 address is requested.  Then after that address is used, a newly requested p2sh (segwit) address will again reuse that public key.

As we discussed in the initial review of the Coin.Space wallet, we have listing requirements to avoid public key reuse mainly for privacy reasons, but also for security reasons.  Choosing to use BIP49 and BIP84 paths would easily avoid this issue.  Simply not using an already used index for new addresses would also address the issue.

I would recommend using BIP49 and BIP84 for compatibility with other wallets (although let me be clear, we have no listing requirements mandating that use; only that public keys not be reused).

nikashitsa:
Ok, we will fix it and let you know.

nikashitsa:
@crwatkins 
> As we discussed in the initial review of the Coin.Space wallet, we have listing requirements to avoid public key reuse mainly for privacy reasons.

It was many years ago... I found only "Avoid address reuse" here https://github.com/bitcoin-dot-org/bitcoin.org/pull/963
We don't reuse addresses. Should we fix public key reuse to get this PR approved?

crwatkins:
@nikashitsa 

> We don't reuse addresses. Should we fix public key reuse to get this PR approved?

Right.  The addresses can be derived from the public keys.

nikashitsa:
Ok, we will fix it and let you know.

nikashitsa:
@crwatkins I'm happy to tell you that we implemented BIP49 and BIP84 in v2.20.1.

Users who have already used old implementation could recover missing funds with our special open source tool ([more info](https://coinapp.zendesk.com/hc/en-us/articles/360049288993-Missing-funds-on-P2SH-and-Bech32-addresses-How-to-recover-)).

crwatkins:
@nikashitsa Your update looks good to me.  It's unfortunate that you had to create a tool to recover the funds, but I did quickly test it and it worked for me.

Are you still using 21.co/earn for your fee estimation?  Your fees always appear unusually high.

In any case, I recommend this update.

crwatkins:
@nikashitsa I'm seeing some serious problems with this wallet.  I sent a transaction to the wallet and did a refresh on Coin Wallet and the balance updated and the transaction showed as Pending.   Then after it was confirmed in one block, I refreshed Coin Wallet again (the circle-arrows) and the new transaction was removed from the balance, and the transaction was gone in the history.   After I verified, using the explorer at blockchair.com that you link to that the transaction was indeed confirmed, I refreshed Coin Wallet again and it came back.   I consider incorrect balances and incorrect history a very serious issue, similar to what we had seen during the early days of Coin.Space.

As for fees, the fee that I used in the above transaction was 678 sat at about 3 sat/B and was mined in the next block.  The current suggested fee by Coin.Wallet for a one hour confirmation is 54269 sat, 80 times what I just paid.   Is something seriously broken here?

crwatkins:
@nikashitsa In transactions being sent by Coin Wallet, it appears that there is a charge going to a third party (I assume Coin) on each transaction.  Is this correct?

1.  When did this start?   I see some code commit related to it in July 2019
2.  Did you inform your users that you were now charging for each transaction placed in the wallet?  I don't recall any such notice and the last transaction I sent before then did not require any such charges.
3. Where in the Coin Wallet is the purpose and the amount of this charge disclosed?
4. I see a discussion at https://www.coin.space/all-about-bitcoin-fees/ about Bitcoin fees.  There are over 20 paragraphs dedicated to discussing items such as "Why are fees so damn high?" and there is one sentence after many pages of information stating:  `From a Coin perspective, we add a fee based on a value of a transaction. Coin fees 0.5% (min. $0.30, max $100) per transaction.`  Is this the only notice you provide and is this notice accurate?
5. I just sent a transaction of $0.50 (I'm stating this in dollars since your above page references dollars).   It looks like the wallet sent $0.49 to a third party (again I assume Coin) and the miner fee was $3.91 (which was extremely high for current conditions, but that's another matter).  The charge to Coin seemed to be 98% of my transaction and over 60% over the claimed minimum.  If my understanding is correct, please explain why this charge is significantly greater than the description I quoted above.
6. The charge in the above transaction was sent to an address to which nearly 2,000 other transactions sent to.  I assume you can now go and view all of your users' transactions by looking at some small number of addresses that are used for these charges.  Please explain how this is not destroying the privacy of all of the users of the wallet.

nikashitsa:
@crwatkins thank you for your review!

> Are you still using 21.co/earn for your fee estimation? Your fees always appear unusually high.

Yes, we are still using this fee [estimation](https://bitcoinfees.earn.com/). Perhaps, it works incorrect? 🤔

Our backend is not perfect, some glitches as you got during new block processing have place. We are working on it.

> In transactions being sent by Coin Wallet, it appears that there is a charge going to a third party (I assume Coin) on each transaction. Is this correct?

Yes, it's our address.

1. Correct, summer 2019.
2. Our [recover tool](https://recover.coin.space/) doesn't charge such fee. Before user confirms transaction we show total fee amount again. User can click cancel, if fee is too high.
3. In order to keep things simple in wallet we show final fee amount = miner fee + our fee. For the same reason we don't show change address.
4. Yep, we missed there `+ 1920 satoshi` per transaction. Already [fixed](https://www.coin.space/all-about-bitcoin-fees/).
5. Point 4 gives answer.
6. Sorry, but I can't understand how this info could destroy the privacy of our users. It doesn't give access to user passphrase or private keys. Could you please clarify?

My thoughts... When I withdraw crypto from almost any exchange they put my transfer into transaction with other users transfers. Single transaction with multiple outputs. I know other user addresses and that they used same exchange as me. But how this info can be used to disclose those users?

crwatkins:
@nikashitsa 
> Sorry, but I can't understand how this info could destroy the privacy of our users. It doesn't give access to user passphrase or private keys. Could you please clarify?

Please see https://en.bitcoin.it/wiki/Address_reuse

> When I withdraw crypto from almost any exchange they put my transfer into transaction with other users transfers.

Exchanges may involve an intrinsic loss of privacy.  Wallets should not.

nikashitsa:
@crwatkins 

From https://en.bitcoin.it/wiki/Address_reuse  

> When addresses are re-used, they allow others to much more easily and reliably determine that the address being reused is yours.

So, it's possible even if you don't re-used addresses. Just need to spend little more time.  
We have only single output address re-used for all users. Making relationship graph will take same time even without this address. Perhaps, I still didn't get your idea, please share.

I see that Coin Wallet is already removed (#3369) from your list. What should we do to get back there?

crwatkins:
It makes wallet fingerprinting completely trivial.   See https://en.bitcoin.it/wiki/Privacy#Wallet_fingerprinting Many privacy issues as well as countermeasures are cumulative, however if I simply know that you use Coin Wallet and I see you make a transaction or even know that you are planning to pay someone around 2pm, I can be fairly certain of how much you paid to what address by simply looking on the blockchain.

nikashitsa:
How can you be so sure, that it is my payment around 2pm? Perhaps, I forgot to make it and another Coin Wallet user did it?

Just imagine that 99% of transactions in blockchain have same single address in outputs. How would you detect that two transactions belong to same user? We have a lot of users :)

From `Open Bitcoin Privacy Project`
> Sending funds to the same Bitcoin address in multiple transactions provides strong indication to privacy attackers that the funds are owned by the same entity.

Yes, attackers know our fee address for sure. Our privacy is damaged, not users.

crwatkins:
Maybe this will help https://bitcoinops.org/en/topics/output-linking/

nikashitsa:
It's all about when you receive funds on same address.

> Output linking, also called address reuse, occurs when a user receives two or more payments to the same public key or other unique script element.

Not about when user send funds to same address multiple times.

So, @crwatkins what should we do to get back to your list?

Perhaps, we should show in details the fee which user send to our address? Display **miner fee** and **Coin Wallet fee** in order to make it more transparent?

We also have ideas about how to make Coin Wallet fee address unique for every user payment. But it will reduce our income a bit.

crwatkins:
I suggest to start by addressing the current [site owner's concerns](https://github.com/bitcoin-dot-org/bitcoin.org/pull/3368#issuecomment-637672130).  I have already given you [my recommendations](https://github.com/bitcoin-dot-org/bitcoin.org/pull/3368#issue-426678275).  That said, I am embarrassed to have recommended this listing and I personally won't be doing any more reviews on it for at least two years.  Anyone can write reviews and make recommendations and I suspect there are at least hundreds of people that the community would consider quite qualified.

nikashitsa:
Ok, thank you Craig!   
You are doing a great job 😊 

I hope we will do some update and get back soon.

PR #3368, opened 2020-06-02
Remove Coin Wallet

crwatkins:
This PR removes Coin Wallet from the wallet listings.

I believe that Coin Wallet violates the intent of the bitcoin.org address reuse criteria by [attaching a charge](https://github.com/bitcoin-dot-org/bitcoin.org/pull/3335#issuecomment-637437354) to every transaction which the wallet creates and sending that charge to a common address (or addresses) controlled by CoinSpace.

Note that there is currently no criteria precluding wallets from charging for their service and that in itself is not an issue.  That said, I don't believe any other currently listed wallets charge for sending simple transactions.

There are also currently a number of issues with Coin Wallet that I personally find troubling and I would highly recommend addressing:

1. The wallet [does not consistently display correct balance and history](https://github.com/bitcoin-dot-org/bitcoin.org/pull/3335#issuecomment-637219838).  This issue [was reported](https://github.com/bitcoin-dot-org/bitcoin.org/pull/963#issuecomment-124629555) in a review five years ago.
2.  There is no indication in the wallet, or in either of the app stores, that Coin Wallet charges for sending transactions.  The home page touts **Free Bitcoin Wallet** and **Get a Free Account** with no hints there are any charges.  The charges are not trivial, with the current minimum charge being over 10 times the amount of the network fee I regularly pay.
3. There is no indication in the wallet of what the charge is being paid to CoinSpace for any given transaction.
4. The Bitcoin network fees used by the wallet seem excessive, often being 10 times or more the estimates of other wallets.
5.  The website FAQ and the tooltip in the wallet points to a 1000+ word [All About Bitcoin Fees](https://www.coin.space/all-about-bitcoin-fees/) which answers questions such as **Why are fees so damn high?** implying that the exceptional fees and charges imposed by the Coin Wallet are simply a defined characteristic of the Bitcoin network.  

I can no longer recommend Coin Wallet for listing.

@wbnns Note that this PR is untested and removes the last wallet from the Web wallets category which may necessitate other updates before merging.


Cobra-Bitcoin:
For me the charging for service isn't an issue, as wallets obviously have expenses like server infrastructure and developers. But such a charge should be communicated properly to users, along with the risk of privacy leaks that come with it, then users can decide whether they want to continue using the wallet. It's distasteful to cover up the charge in this way and hide it behind normal network fees, it gives users the wrong impression about fees on the Bitcoin network. It's OK to want to make some money from a wallet to fund continued development, but **_not_** in this sneaky way.

I was particularly concerned when you discovered they were generating HD Segwit addresses in a manner different to what's specified in BIP 49 and BIP 84. To me that's a huge red flag, worse than anything else IMO. The reason these standards exist is to enable cross compatibility between wallets so users can get their funds back easily should a wallet disappear. I'm glad they fixed it but that was really alarming.

Fee estimation being badly broken is unacceptable and grounds for removal just on its own too. It's really sad that we sent thousands of users to this wallet and let them pay these fees without them even knowing where their money is going. ACK on the removal.

crwatkins:
> For me the charging for service isn't an issue

I agree.   Sorry that I might have implied otherwise by reporting the status of the other listed wallets.

wbnns:
@crwatkins 

I've opened #3369 which includes your commit in this PR (cd214a3) along with the additional required changes to omit the Web wallet category, since as discussed, there now are none.

It also removes references to Coin from the various translation files.

Dolorozo:
Hi @Cobra-Bitcoin!
We’d like to reopen the discussion about adding Coin Wallet back to the list on bitcoin.org. It’s been almost 5 years since the removal, and we’ve addressed the key concerns raised at the time.

Summary of resolved issues:
1. HD SegWit derivation (BIP49/BIP84 compatibility)
This was fixed immediately after it was reported.
https://github.com/CoinSpace/CoinSpace/commit/b83cb76ed39c3d7eddba94bcc1b29564b6ae41e3
https://github.com/CoinSpace/cs-wallet/commit/dea1d02f9595ae6c1cf6873d9faf5a3e99a53cc3

2. Use of the word "free” on website and apps
Phrases like “Free wallet” and “Get a free account” have been completely removed. There are no mentions of “free” left, even when searched across the app or site.

3. Service fee transparency
We’d like to focus on this point. As noted in the original discussion, charging a fee isn’t the issue - the key is making it clearly understandable to users.

Currently, in the final step of sending BTC, the user sees:
- The total amount being sent
- The total fee (in BTC) — tapping this shows the equivalent in the wallet’s selected fiat currency (e.g. USD)

To clarify:
Are we correct in understanding that the main concern is the lack of separation between the Bitcoin network fee and our service fee?
If so, we’re open to adjusting the UI to display a clear breakdown.

fee
fee2

crwatkins:
The major concern of this PR was

> I believe that Coin Wallet violates the intent of the bitcoin.org address reuse criteria by [attaching a charge](https://github.com/bitcoin-dot-org/bitcoin.org/pull/3335#issuecomment-637437354) to every transaction which the wallet creates and sending that charge to a common address (or addresses) controlled by CoinSpace.

which still has not been addressed, even after five years.

Ashworth78:
👍

Dolorozo:
Hi @crwatkins, @Cobra-Bitcoin,
We discussed with the dev team the issue of using a single address when sending transactions. We can fix this and implement a solution, but the task is not the easiest one, so we would like to **clarify** with you once again: what is considered a bigger red flag?

We remember this.
>To me that's a huge red flag, worse than anything else IMO.

But it turned out that this is not actually a red flag.

Btw, you have the “privacy” and “fees” labels with “caution.” Maybe they could be used to mark the wallet. We don’t mind if they are applied, along with a note about the high fees. I understand that the “fees” label usually refers to the network fee amount, but why not make it clear and write the full explanation?

We would greatly appreciate honest feedback so that we don’t have any false expectations!

crwatkins:
As I [stated ](https://github.com/bitcoin-dot-org/Bitcoin.org/pull/3335#issuecomment-640146559) over five years ago, I'm personally not interested in reviewing this wallet for at least two years after the issues that have been enumerated have been addressed.  As an aside, I had expected that these issues would have been immediately addressed as serious and obviously the right thing to do, rather than simply a checkbox to qualify for listing five years later.

In any case, as a purely administrative recommendation, this PR has been closed for years and if you wish to find someone to review a proposal, I would recommend making a proposal in a new PR to gain the proper attention.

Ashworth78:
My thoughts exactly,👍

PR #3404, opened 2020-07-01
Change Neutral wallet score label to Not applicable

### LINKED ISSUE #3136: Eliminating confusion surrounding presence of scores associated with hardware wallets
There is some [confusion](https://twitter.com/pavlenex/status/1179753704490229761) regarding why hardware wallet scores are presented the way they are. For example, despite the presence of certain characteristics (like in the aforementioned link) why criteria like Privacy or Validation would be rated Neutral.

Rather than have a blanket neutral classification that applies to all hardware wallets for different criteria, we should consider the default software that accompanies each hardware wallet when it's used. In other words, if the default software that accompanies a hardware wallet allows a user to have control over fees when making a transaction with the hardware wallet, the criteria would be rated accordingly. Same with privacy, etc. From a site visitor's perspective, this would be easier to understand than us explaining concepts like hardware signing, etc., and why a privacy rating doesn't apply to the hardware. It would be better to have a "Ok, cool, so I'll be able to have good fee control and decent privacy with this hardware wallet, too" UX instead of having thoughts like "Wait, why is this neutral" or "Why is there no privacy rating like the Android wallets" when reading the page.

In the past it has been discussed whether or not hardware wallets should have their own unique criteria altogether compared to software wallets / if certain criteria should be omitted, however, it's enough of a challenge as it is for users to understand the ratings when they're new to bitcoin. Also, software is required in order to use a hardware wallet anyway. So from the user's perspective, if they purchase a hardware wallet, and the default accompanying software used to make transactions provides them with good privacy and fee control, this should be indicated on the wallet page for that hardware wallet.

Summing up - this issue is proposing to provide a rating for all criteria as described above, rather than a lengthy explanation, the removal/omission of criteria or a neutral score.

Cc: @crwatkins @nvk

crwatkins:
Resolves #3136 

This change is to replace the confusing term "Neutral" with a more accurate "Not applicable".  This only affects hardware wallets.

@nvk

### DIFF
--- a/_translations/en.yml
+++ b/_translations/en.yml
@@ -269,7 +269,7 @@ en:
     wallet-scoring-good: "Good"
     wallet-scoring-pass: "Acceptable"
     wallet-scoring-fail: "Caution"
-    wallet-scoring-neutral: "Neutral"
+    wallet-scoring-neutral: "Not applicable"
     walletcatmobile: "Mobile"
     walletcatmobile-description: "Wallets are available for Android and iOS based operating systems."
     walletcatdesktop: "Desktop"


Cobra-Bitcoin:
LGTM.

PR #3439, opened 2020-08-08
Add lightning support to Electrum

brianddk:
Reflecting the lightning support added since Electrum 4.0.0

CC: @ecdsa as the maintainer of `spesmilo/electrum`

### DIFF
--- a/_wallets/electrum.md
+++ b/_wallets/electrum.md
@@ -16,7 +16,7 @@ platform:
       link: "https://electrum.org"
       source: "https://github.com/spesmilo/electrum"
       screenshot: "electrum.png"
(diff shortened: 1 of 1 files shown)


Cobra-Bitcoin:
LGTM.

crwatkins:
LGTM.

@ecdsa has [noted](https://github.com/spesmilo/electrum/blob/3b5b020941a25495cb9e8625ba4c8299928c3714/RELEASE-NOTES) that Lightning Network support is experimental.   Normally we don't acknowledge beta or experimental support of features, but I believe that LN is a special case since the entire network is still considered experimental.    I'll note that, as can be expected, the current UX relies on a fairly high level of user sophistication, particularly when handling errors and exceptions.

PR #3444, opened 2020-08-14
Update _config.yaml

k3tan172:
Just updating the bitcoin chain size. Might be worthwhile changing other variables as well.


harding:
ACK bc6a17b163c463ab340d72e2506cffccd3722460

PR #3420, opened 2020-07-14
Add OKCoin to list of exchanges

dyljm1:
https://www.okcoin.com

PR #3445, opened 2020-08-15
Revert "Add OKCoin to list of exchanges"

Lev-8:
Reverts bitcoin-dot-org/bitcoin.org#3420


Cobra-Bitcoin:
Can you specify a reason for why we should revert the change?

PR #3448, opened 2020-08-19
fix broken line break tags

jlopp:
I noticed some broken HTML at https://bitcoin.org/en/bitcoin-core/features/requirements which led me to discover several pages that appear to have had the closing ">" of line break tags moved to a separate line from the "<br"


Cobra-Bitcoin:
Nice catch!

PR #3477, opened 2020-09-11
Add Bitcoin Global to list of P2P platforms

vladyslavstartsev:


vladyslavstartsev:
If there is a need in some description, I'm happy to add it. But there is no description what should be added.

vsus:
As far as I know, the website was created on July and it almost has no visits.

vladyslavstartsev:
@vsus what is the minimum age of the platform that you will accept?

vsus:
I'm not a maintainer of Bitcoin.org. I just wanted to mention some facts I consider relevant.

Cobra-Bitcoin:
We usually add only established platforms, so you need to build out more and get more usage, because it isn't beneficial for us to point our users to services that aren't really being used yet.

Sorry, hope you understand.

PR #3481, opened 2020-09-22
Add WhiteBIT link in exchanges.html

eugenesaenko:


Cobra-Bitcoin:
Exchange is not big enough. We can re-visit this in a year.

PR #3485, opened 2020-09-27
Add 'BTCPay Server' to free software projects

l33d4n:
BTCPay Server is a free and open-source cryptocurrency payment processor which allows you to receive payments in Bitcoin directly, with no fees, transaction cost or a middleman.

BTCPay Server is a non-custodial invoicing system which eliminates the involvement of a third-party. Payments with BTCPay Server go directly to your wallet, which increases the privacy and security. Your private keys are never uploaded to the server. There is no address re-use, since each invoice generates a new address deriving from your xpubkey.


crwatkins:
LGTM.

PR #3514, opened 2020-10-07
Add Trustee Wallet

rhrusha:
Hi guys, 

There is our second try to list Trustee Wallet on bitcoin.org.
We reviewed [listing criteria](https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/docs/managing-wallets.md) and believe meet all of them. 

Previous PR #3190


crwatkins:
Thanks for your submission.

Note that I don't believe that the current Travis CI build failure is related to your commit.    I believe the wallet syntax checks passed OK.

crwatkins:
A criteria for listing is 

* Source code is public and kept up to date under version control system

Your repository seems to have tagged release (some with thousands of lines of code changed) and no update comments.  This makes reviewing development very difficult.  All other listed wallets have active development on GitHub with granular commits.   Do you perform your development in some other manner?

rhrusha:
yes you are right, for a bunch of reasons we are developing in our internal repository, and publish releases only. 

In order to protect ourselves from copying our ideas, we cannot publish intermediate stages of development. And we can’t share features releasing plans, which can be easily obtained if we share internal repository. 

Despite the fact that we are publishing the full source code to Github. We are not ready to publish anything that is still in development or on which we are experimenting. 

And finally we want to protect the results of our work from the risk of being blocked or removed from the GitHub for any legal reason. Like the result of a potential attack using courts or lawsuits against us or third parties that may indirectly affect us.

rhrusha:
Sorry for late response, we tried to reach verifiable build first according to [this](https://github.com/trustee-wallet/trusteeWallet/issues/1) request. But still have some de-compilation differences which seems will take more time that we expected to investigate why this happening.

Giszmo:
While I find @crwatkins' objection weak, I don't follow @rhrusha's excuse for the squashed commits neither.

If having granular commits and up-to-date repositories is required, it's merely requested for some rite of passage and then never checked on again. At WalletScrutiny I follow up with projects, checking on the latter and my checks fail all the time for missing tags or even missing commits and here reproducibility is also listed as some feature of "the wallet" while it is at best a feature of the wallet at certain tags and most relevant for certain binaries (from Google Play for example).

On the other hand, if a project doesn't want to develop under public scrutiny, it can still provide granular commits and that is not only important for public but also for internal scrutiny. Ideally each developer would sign his commits so that rogue changes can be attributed when found. At Mycelium we have a private repository and only with new binary releases we push our changes to the public repository but feature branches we still don't want to publish we keep on the private repo, only. Many features live a boring live since years in their abandoned branches after the feature was canceled :D

Vadymhr:
Hi Guys! Good comments!
I have fews important points to add about Trustee Wallet.

Trustee is not just another multi-coin wallet with Bitcoin. First of all it is a Bitcoin wallet. I say this because as far as I know it is the world's first multi-coin wallet that focuses on Bitcoin first that much. And at the same time far ahead many bitcoin only wallets. 
Trustee has a flexible approach, any beginner has to learn almost nothing to get started using Bitcoin. And learn more as you go. Trustee one of the first implemented CPFP, SegWit, RBF,... In the next few weeks Trustee released the ability for selecting inputs and sending several recipients with a single transaction. At the same time keeps working not only modern phone OS versions but also Android 4.4
Users loved this solution and   This is not everything that Trustee enables for users and much more is coming.

crwatkins:
The last four versions on the iOS app store have not be reflected in the referenced source repository.  As I've mentioned above a requirement for listing is:

* Source code is public and kept up to date under version control system

rhrusha:
Hi @crwatkins 

For the last couple of month we made our development much more open for the community. All new commits is going now to "dev" branch.  So the wallet repository contains all actual commits. 

We did not commit any release to Github for the last month because starting from Dec 15 and till Jan 1 our app stuck with Playmarket "review" state. We was not able to publish anything to Playmarket even for internal tests. Then our app passed google review for short period  of time and again stuck with "review" state. 

According to our development process we making release on Github after releasing app in all app markets.
Now we are changing our processes to make Github release independent from app markets release and new release will be available till the end of this week.

crwatkins:
@rhrusha Thanks.

Giszmo:
> According to our development process we making release on Github after releasing app in all app markets.

So you release a binary update to all 100k users and **then** share the source code? That's the wrong order if you want your users to profit from public scrutiny. A good wallet should release the source well before the binary. As Google is often not quick to approve updates, I find it fair to push the source to GitHub before pushing the binary to Google with subsequently a slow rollout (2 days to review, 4 days with beta testers). This should give interested parties time to check what changed but the ideal process is to not do code drops at all but to develop in the open.

Did you [look into reproducibility](https://walletscrutiny.com/android/com.trusteewallet/)?

Vadymhr:
@Giszmo Thanks for your comment! As always focused and detailed!

I have good news, these days we develop Trustee in public repo. 
Meaning there is no chance any changes might appear in app market before Github. It is in branch dev. @crwatkins FYI 
This actually what @rhrusha tried to say above.

https://github.com/trustee-wallet/trusteeWallet/tree/dev

Vadymhr:
> Did you [look into reproducibility](https://walletscrutiny.com/android/com.trusteewallet/)?
 
@Giszmo Yes, we did look in to it and spend a lot of effort to fit your certification requirements. But we did dive in to problems to get what you requested which we don't have a solution for now. We already did talk with you about it on another thread. 
If you can suggest or propose the solution we will be happy to implement it.

Giszmo:
> If you can suggest or propose the solution we will be happy to implement it.

If you are referring to [this issue](https://github.com/trustee-wallet/trusteeWallet/issues/1#issuecomment-577612039), I don't think I can help much. Absent a concrete question I would have to get very familiar with your individual project and I'm monitoring 300 of those.

crwatkins:
Do you expect the Testnet wallet option to work?  I get errors such as "Sorry... undefined is not an object (evaluating 'T.selectedTransferAll Balance')".   I also can't seem to set a fee.  No matter what I type in "Fee for byte" (very confusing) the [Apply] doesn't work.   Then when I click Send I get "Sorry... Please, select the available free from the list" and I see no list.

crwatkins:
In an hour or so, I have observed a number of serious issues with this wallet that either directly relate to listing criteria or the simple overall functioning of the wallet.

* The iOS app crashed a number of times
* The wallet double-spent an input, making a sent transaction invalid
* The wallet seems to, by default, re-use a single received address
* The wallet displays "Received from" addresses
* Many English messages and dialogs are not well worded and are very confusing.  For example, one dialog asks Yes/No? and there is no question asked to answer
* I did not understand a number of the English error messages
 *  In attempting over a dozen transactions, the wallet only ever presented a selection of fees once, making for many failed send attempts and a completely confusing user experience for all other attempts
* After a manual fee was entered, it showed 0 BTC for the fee on the send page

I have not done a complete review, but I cannot recommend this wallet for listing at this time.  A serious amount of work needs to happen before this can be reviewed.

Vadymhr:
@crwatkins first of all, thank you very much for your efforts! And sorry that we didn’t meet your expectations for the first time(. But be sure we will address all the issues and will make it work in a best way for further bitcoin development and movement.

> * The iOS app crashed a number of times
Probably the actual review happened not in the best time. As I mentioned before we planned a major release that actually happened in January and there is indeed was a few unwanted issues. Fixed for these days.

> * The wallet double-spent an input, making a sent transaction invalid
  This is very weird since thousands of people use the product on a daily basis and we didn’t see this issue. It would be great if you can share more details about it or perfectly if you give exact steps to reproduce.

> * The wallet seems to, by default, re-use a single received address
  As we understood you are talking about this section
  https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/docs/managing-wallets.md#score section Privacy 
  “To get a passing score, the wallet must avoid address reuse by using a new change address for each transaction.”
  
  We don’t know what is your experience with Bitcoin new users but from our experience they have a very hard time with understanding address rotation mechanics. 
  So, what if do the next mechanics by default - accepting address is reusable but change address is new for each transaction. It looks that this meets the accepting policy mentioned above.
  
  What do you think?

> * The wallet displays "Received from" addresses
  Didn't get what wrong with this? We show incoming address in case there is only one input address in the tx and don’t show if there is more than one.

> * Many English messages and dialogs are not well worded and are very confusing.  For example, one dialog asks Yes/No? and there is no question asked to answer
Could you please point us to some examples? Here is our lang file https://github.com/trustee-wallet/trusteeWallet/blob/master/locales/en.json

Regarding few other issues, wallet in use by many people so it works. Probably you did very trick scenarios which basic users don't do. But we will be happy to fix just need to understand the reproduce scenario. Will put an effort to do it on the info you provided. In case you have any information to add please add in any comfortable way for you.

crwatkins:
> This is very weird since thousands of people use the product on a daily basis and we didn’t see this issue. It would be great if you can share more details about it or perfectly if you give exact steps to reproduce.

I sent a transaction from the wallet to an address.   Later that day I sent another transaction to the same address (while the first transaction was still unconfirmed).   The second transaction re-used a txo from the first transaction, cancelling the first transaction (RBF).  The recipient only received one payment, not the required two payments.

> Could you please point us to some examples? 

The two example I can remember are the Yes/No question without any question (I could certainly guess what it meant, but I shouldn't have to) and on the page that wasn't displaying any fee options, I tried to enter a manual fee and it was telling me I need to enter an address first.   There were no addresses on that page.    You could search your modal dialogs for a Yes/No response with no question marks in the dialog.

> As we understood you are talking about this section
> “To get a passing score, the wallet must avoid address reuse by using a new change address for each transaction.”

That, plus just as important, immediately before that criterion:

* Avoid address reuse by displaying a new receiving address for each transaction in the wallet UI

Please carefully review [Bitcoin wiki address reuse](https://en.bitcoin.it/wiki/Address_reuse)

Your current practices not only damage the security and the privacy of your users, but they damage they
privacy of the entire community.  It is very concerning that any wallet would be doing this in 2021.

If the current stability, accuracy, privacy, and basic functionality issues are resolved, I would be willing to reconsider in about a year.

Vadymhr:
@crwatkins, thanks for the quick reply and additional details!

Checking these issues.

In the meanwhile regarding reusable addresses, we did our own scientific cryptographic research a few years ago. As an summary there are no real risks for security although it downgrades some cryptographic strength. So, it is mainly a privacy question which beginners do not really look for and in case needed it is very easy to switch.

crwatkins:
I'll close for now and look forward to a full review after issues have been addressed in about a year.

PR #3521, opened 2020-10-22
Add BlockSettle exchange

BlockSettle:
Our exchange is non custodial on the BTC side. We hold and manage the fiat, buyers have to deposit fiat with us. We do not hold any BTC. We do not have any control over user coins: we do not proxy/escrow coins on user's behalf. We exist as a third party that handshakes buyer and seller, to generate a set of BTC delivery transactions, based on our protocol. The coins go from seller to buyer, they do not transit through us.
Our central matchmaking server is closed source. The clients (the open source wallet) have the full fledged code to verify the handshake and validity of the delivery transactions. Users run the protocol checks on their own as well (with their own full node if they wish to, this is the Armory stack after all). None of the fiat side code is available to users, that part is fully proprietary and centralized.
Once a BTC transaction is setup, we monitor it and release the cash to the seller when the buyer takes delivery of the BTC (6 confs wait period).


Cobra-Bitcoin:
Your exchange isn't yet big enough, but we can review again in a year.

PR #3527, opened 2020-11-03
Update he.yml

ShimonGurman:


Cobra-Bitcoin:
Better to fix the issue directly in Transifex so next time we pull the translations from there, this change won't get overridden.

PR #3536, opened 2020-11-25
Fix content overflow for button

yslpn:
Hey. My native language is russian. “Сделать пожертвование” and “Пожертвовать” have the same meaning. But in the first option, there is a content overflow in the button.


Cobra-Bitcoin:

PR #3537, opened 2020-11-26
Fix typo

yslpn:
Corrected a typo. It was "electronic almost" now "electronic mail".


Cobra-Bitcoin:

PR #3546, opened 2020-12-16
Fix word in ru.yml

delef:


Cobra-Bitcoin:

PR #3555, opened 2021-01-04
Add Bitclude.com into exchanges.html 

bitclude:
Second biggest exchange in Poland, operate in Bitcoin mainly, regulated in Poland.


Cobra-Bitcoin:
LGTM. I have verified that this is a legitimate and licensed exchange in Poland.

PR #3560, opened 2021-01-14
Bitcoin Core 0.21.0

laanwj:
Can be merged if it passes Travis


laanwj:
I've replaced the release notes with a link to bitcoincore.org's release notes page. This is both less hassle (getting it to pass Travis) and serves to direct people to bitcoincore.org for the download.

PR #3563, opened 2021-01-18
Add Norwegian Block Exchange to exchange list

gcaven-nbx:
Adds Norwegian Block Exchange to list of exchanges, in a new "Norway" sub-section
See https://nbx.com


Cobra-Bitcoin:
LGTM. Seems like a legitimate exchange.

PR #3565, opened 2021-01-19
add guides panel to resources

RandyMcMillan:
Screen Shot 2021-01-19 at 2 19 25 PM

### DIFF
--- a/_templates/resources.html
+++ b/_templates/resources.html
@@ -127,5 +127,20 @@ id: resources
         </p>
       </div>
 
+      <div class="card resources-card">
+        <img src="/img/icons/ico_invoice.svg?{{site.time | date: '%s'}}" alt="Icon">
+        <!-- <h2 id="guides">{% translate guides %}</h2> -->
+        <h2 id="guides">Guides</h2>
+        <p>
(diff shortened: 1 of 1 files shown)


Cobra-Bitcoin:
LGTM.

Let's see what other people think.

RandyMcMillan:
I will be happy to remove or add other guides that the community suggests.

crwatkins:
I'm OK with the concept of guides with the caveat that I could imagine that this list could get overrun with random guides without enough description for a user to select something appropriate.   For example, the first guide isn't really a Coldcard[*] guide.   It's more a Specter guide, or even more accurately a Specter/Coldcard/Bitcoin Core guide.  In any case, folks looking for the simplest way to use their Coldcard could be quite disappointed.  

[*] Coinkite is fairly inconsistent with the spelling of Coldcard vs ColdCard, however they do use the "Coldcard" spelling on their legal page as we do on bitcoin.org.

RandyMcMillan:
after some more input/suggestions - i will add/remove any links that are agreed upon...

Cobra-Bitcoin:
OK, I'll merge this, but let's be conservative in the future with how many guides we add to this list. We'll keep it to only the best ones. The resources page always needed more love anyway, so thank you for adding this.

PR #3567, opened 2021-01-19
Add new Apple download domain

crwatkins:
Apple is [now](https://developer.apple.com/news/?id=06142019a) using apps.apple.com for downloads.
Also, cleaned up some trailing whitespace.


Cobra-Bitcoin:
Thanks, glad Apple changed this, these new links look much cleaner.

PR #3581, opened 2021-02-02
Donation: Ensure address link has full URI (incl amount)

monokh:
The link on the donation link that has the URI `bitcoin:<address>` does not update when the QR code is updated. As a result clicking the link will not relay the selected amount to the application that handles it. This PR fixes this.

image


Cobra-Bitcoin:
Good catch, thank you!

PR #3586, opened 2021-02-07
Adding Coinmama to Intl, UK and UAE exchanges

dazaraf:
Coinmama is a reputable broker-dealer servicing customers across 188 for 8 years now. They make it super easy and secure to buy Bitcoin and should therefore be included in the Bitcoin.org exchanges section.

### DIFF
--- a/_templates/exchanges.html
+++ b/_templates/exchanges.html
@@ -19,7 +19,6 @@ id: exchanges
     <br>
     <a class="marketplace-link" href="https://gemini.com/">Gemini</a>
     <br>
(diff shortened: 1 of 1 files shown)


Cobra-Bitcoin:
Agreed with adding it in the UK/UAE sections, but not in international, as that part of the exchanges page is for really big and very reputable exchanges.

dazaraf:
Hi Cobra thanks for the response.

What about adding it to the US section as well such that it will be on the
UK, UAE and US section?

Let me know and I’ll make a new pull request with the changes.

Cobra-Bitcoin:
@dazaraf Yes, that's OK.

PR #3622, opened 2021-03-19
Replace blockcypher with mempool.space

jsarenik:
git grep -i blockcypher \
      | cut -d: -f1 \
      | while read file \
    do sed -i 's|live.blockcypher.com/btc|mempool.space|g' $file
    done

Mempool.space is an actively maintained GNU AGPL-licensed
project available at https://github.com/mempool/mempool
It focuses on Bitcoin network only.

BlockCypher seems to track multiple blockchains but for example
does not contain the genesis block for Ethereum which contains
all the pre-mined 72 million ETHs.

What more, [mempool.space](https://mempool.space) does not provide merely Mainnet
and Testnet, but **also Signet**.


jsarenik:
FYI: @wbnns (because of 98fd672d)

crwatkins:
* I believe mempool.space would be an excellent addition to the resources list
* The `block170` reference seems to be dangling and I believe it can be deleted
* I would be fine with changing the `advantagesli5` to reference the excellent mempool.space instead
* However, I don't see any compelling reason to remove BlockCypher from the resources list

jsarenik:
@crwatkins Thank you for feed-back! Excuse me, just now I found that https://github.com/blockcypher/explorer is also open. Yes, there does not seem to be any compelling reason to replace BlockCypher with Mempool.Space right now.

crwatkins:
@jsarenik Could you modify this PR to add mempool.space in addition or submit a separate PR?  Thanks!

jsarenik:
> @jsarenik Could you modify this PR to add mempool.space in addition or submit a separate PR? Thanks!

Will do, sure.

PR #3624, opened 2021-03-21
Make minimum requirements accurate to improve decentralization

CrazyPython:
The easier it is to run a full node, the more full nodes Bitcoin will have. The more full nodes Bitcoin has, the more people use full nodes, and the fewer people use custodial solutions, making bitcoin more decentralized. By noting how only 5GB of disk space is required, we can improve the decentralization of bitcoin.

This page is the page most people land on. The wording on this page talks about "minimums". While there is a separate page on the actual minimum requirements, most people will not see it, they will see this 350GB requirement and think "better use a custodial solution."

### DIFF
--- a/en/full-node.md
+++ b/en/full-node.md
@@ -190,7 +190,7 @@ have an easy-to-use node.
 * Desktop or laptop hardware running recent versions of Windows, Mac OS
   X, or Linux.
 
-* {{site.text.bitcoin_datadir_gb}} gigabytes of free disk space,
+* {{site.text.bitcoin_datadir_gb_pruned}} gigabytes of free disk space,
   accessible at a minimum read/write speed of 100 MB/s.
 
 * 2 gigabytes of memory (RAM)
--- a/_config.yml
+++ b/_config.yml
@@ -616,7 +616,7 @@ text:
   chain_gb: 340
   ## The main difference between the datadir size and the chain size, if
   ## you don't have any extra indices
-  bitcoin_datadir_gb_pruned: 5
+  bitcoin_datadir_gb_pruned: 7
   ## the tx= field in: grep UpdateTip .bitcoin/debug.log | tail -n1
   total_tx_count_in_millions: 230
   typical_ibd_time_in_hours: 4


en/full-node.md
```diff
@@ -190,7 +190,7 @@ have an easy-to-use node.
 * Desktop or laptop hardware running recent versions of Windows, Mac OS
   X, or Linux.
 
-* {{site.text.bitcoin_datadir_gb}} gigabytes of free disk space,
+* {{site.text.bitcoin_datadir_gb_pruned}} gigabytes of free disk space,
   accessible at a minimum read/write speed of 100 MB/s.
```
luke-jr, line 194: Maybe drop this minimum speed bit too. Most are fast enough, and there isn't strictly a minimum...

en/full-node.md
```diff
@@ -190,7 +190,7 @@ have an easy-to-use node.
 * Desktop or laptop hardware running recent versions of Windows, Mac OS
   X, or Linux.
 
-* {{site.text.bitcoin_datadir_gb}} gigabytes of free disk space,
+* {{site.text.bitcoin_datadir_gb_pruned}} gigabytes of free disk space,
```
luke-jr, line 193: Note: Bitcoin Core GUI currently only allows pruning down to 2 GB (+5 GB overhead = 7 GB). So the 5 GB here only holds if the user chooses Knots, or knows how to configure and use bitcoind.

luke-jr (commented):
Concept ACK

en/full-node.md
CrazyPython: I see. Let me know what speed to drop it too.

en/full-node.md
```diff
@@ -190,7 +190,7 @@ have an easy-to-use node.
...
   accessible at a minimum read/write speed of 100 MB/s.
 
 * 2 gigabytes of memory (RAM)
```
CrazyPython, line 196: Most consumer laptops only have 8GB of RAM. Using 2GB of RAM for bitcoin, what is the minimum memory, if you keep UTXO on disk, and still relay blocks? I read elsewhere the minimum of 1GB, but I want confirmation from someone who runs a full node with minimal memory. (@luke-jr)

en/full-node.md
CrazyPython: Most people have no idea how to check their disk read/write speed, so I imagine they would ignore this requirement.

en/full-node.md
CrazyPython: Altered to 7GB in f509116.

Cobra-Bitcoin:
LGTM.

PR #3625, opened 2021-03-23
Adding Prokey Optimum

northmz:


crwatkins:
Thanks for your submission!

While very quickly skimming prokey.io, since this device seems to be mostly a Trezor clone (correct me if I'm wrong), I was looking for differentiating features.  I see references to "Protected against supplychain attacks" and also see references to some "mutual authentication" method, however neither is explained in sufficient detail for me to understand what is really being offered.   The "innovative method for generating secondary keys" in the [blog post ](https://prokey.io/blog/what-is-mutual-authentication) does not seem to provide what I generally consider to be mutual authentication, so it leaves me somewhat confused as to the point.  Can you explain how the supply chain is protected?

Note that there are no requirements for new listings to have any specific level of differentiating features.  I just happened to be looking at those.

northmz:
> Thanks for your submission!
> 
> While very quickly skimming prokey.io, since this device seems to be mostly a Trezor clone (correct me if I'm wrong), I was looking for differentiating features. I see references to "Protected against supplychain attacks" and also see references to some "mutual authentication" method, however neither is explained in sufficient detail for me to understand what is really being offered. The "innovative method for generating secondary keys" in the [blog post ](https://prokey.io/blog/what-is-mutual-authentication) does not seem to provide what I generally consider to be mutual authentication, so it leaves me somewhat confused as to the point. Can you explain how the supply chain is protected?
> 
> Note that there are no requirements for new listings to have any specific level of differentiating features. I just happened to be looking at those.

Thanks for your attention, Our device is a derivation of Trezor with some differences: 

Bootloader has been developed from scratch, it only allows firmware installation after mutual authentication with our servers. so no one cannot manipulate the bootloader and firmware. Mutual authentication is used to check if the device is genuine or not, [this article ](https://prokey.io/blog/what-is-mutual-authentication) is just explained in detail how the website sends a challenge  (a randomly generated number) to the device before installing firmware, and the device signs the challenge if authentication gets done, so It can prevent supply-chain attacks.

PIN on device, the user enters PIN on device instead of entering PIN via computer.

Wallet Recovery has been implemented on firmware, so users don't need to connect to a computer for restoring their wallet, and it can be done 100% offline.

Removing Bridge, Prokey doesn’t need any bridge or running an application, it works with WebUSB on chromium based browsers

Briefly, the motivation behind Prokey Optimum came from the lack of an user-friendly hardware wallet available to users - which don't require installing browser extensions, applications or bridges. In order to solve this it was designed to not require any additional software.

crwatkins:
@northmz 

> Bootloader has been developed from scratch, it only allows firmware installation after mutual authentication with our servers. so no one cannot manipulate the bootloader and firmware. 

Could you explain the advantages of this over the method used by other wallets (such as Trezor) of signing firmware updates?  Are your firmware updates signed?  If so, what additional value does this have?

> Mutual authentication is used to check if the device is genuine or not, this article is just explained in detail how the website sends a challenge (a randomly generated number) to the device before installing firmware, and the device signs the challenge if authentication gets done, so It can prevent supply-chain attacks.

If the device signs the challenge, does each device have a unique private key?  (I don't see that in your article).

I'm not sure I completely follow the "example"  in the article.   Do you have a more complete description somewhere?  Can you refer me to the code?   For example, you mention a "Seed" and you also mention a "SEED".  Are these the same thing?  How are these generated?  How are "KEYS" generated?

The article refers to your "servers".  Are those servers used only for device firmware updates or do they provide other services also?

Could you explain how this "can prevent supply-chain attacks"?  What secret information does the device possess that can not be replicated by an attacker?

northmz:
@crwatkins 
> Could you explain the advantages of this over the method used by other wallets (such as Trezor) of signing firmware updates? Are your firmware updates signed? If so, what additional value does this have?

In Trezor, the firmware checks if correctly signed by SatoshiLabs, this is a client side process and firmware file transfer between client and server but in Prokey, Firmware gets encrypted by a session key from mutual authentication and bootloader authenticates with the server and gets encrypted firmware, so there is no need to hologram seal or vendor reputation because the device will recognize the firmware is original or not. 

> If the device signs the challenge, does each device have a unique private key? (I don't see that in your article).
> I'm not sure I completely follow the "example" in the article. Do you have a more complete description somewhere? Can you refer me to the code? For example, you mention a "Seed" and you also mention a "SEED". Are these the same thing? How are these generated? How are "KEYS" generated?
> The article refers to your "servers". Are those servers used only for device firmware updates or do they provide other services also? Could you explain how this "can prevent supply-chain attacks"? What secret information does the device possess that can not be replicated by an attacker?

Each device does not have a unique key, They all have a 32 bytes seed from which private key is derivative. The keys are secret information that prevent supply chain attacks, For security reasons this part of code has different key in official release,  Please check https://github.com/prokey-io/prokey-optimum-firmware/blob/cdd591bd4fde0c2d2fcb3649ffaf161c5abc5b78/auth.c#L156.

Since our conversation isn't related to this merge request, please open an issue here https://github.com/prokey-io/prokey-optimum-firmware to continue our conversation.

crwatkins:
Criteria for listing include:

* Protects the seed against unsigned firmware upgrades
* Provides source code created for all open components and provides detailed specification for blackbox testing of any closed-source secure elements

I'm trying to understand how this device meets those criteria.  It seems this device contains secret values and and secret code not in the repository and that does not meet the open source requirements.   Correct?

It's hard to understand how the device can protect firmware upgrades based on symmetric algorithms using secret material that is run on a common STM32 processor with no secure protection for the secret material.

northmz:
> Criteria for listing include:
> 
> * Protects the seed against unsigned firmware upgrades
> * Provides source code created for all open components and provides detailed specification for blackbox testing of any closed-source secure elements
> 
> I'm trying to understand how this device meets those criteria. It seems this device contains secret values and and secret code not in the repository and that does not meet the open source requirements. Correct?
> 
> It's hard to understand how the device can protect firmware upgrades based on symmetric algorithms using secret material that is run on a common STM32 processor with no secure protection for the secret material.

The method is clear that how it protects firmware by mutual authentication through private key. 

Keys are confidential in all hardware wallets.
Why Trezor does not publish the private key that signed the firmware? Or why sourcecode of Ledger doesn’t contain root of trust private key?


Source code should show transparency of the method, not publishing keys!

crwatkins:
* I believe Ledger performs device attestation using a well known public key algorithm and a private key stored in a secure element
* Trezor performs firmware authentication using a well known public key algorithm and a public key stored in deterministically built firmware

Both follow well established and vetted best practices, although that is not at issue here.  I am concerned:

1. Prokey does not meet the open source requirement listed above
2. It is not possible to determine whether Prokey meets the firmware requirement because of the use of unpublished code

A specific concern that cannot be verified without this code is whether an attacker can extract secret information from a single Prokey device (as has been shown to be quite possible with similar hardware in a Trezor) and then impersonate/counterfeit any other Prokey or, since symmetric algorithms with secret keys seem to be used, impersonate the Prokey firmware server.  Can you explain why this is or is not possible?

crwatkins:
There has been no activity here for over two months.  

@northmz Are there any updates or should this PR be closed?

northmz:
@crwatkins We decided to make some changes to meet criteria for listing, feel free to close this merge request, I'll make a new pull request once development and test process get done.

PR #3654, opened 2021-04-23
Update exchanges.html

digiorgiu:
Added cryptofacile.io, web app for comparisons based on prices, fees and features.


devdavidejesus:
Thanks, but I'm closing this. The linked site (cryptofacile) is no longer an exchange comparison tool, the domain now resolves to an online casino/gambling affiliate site. Not appropriate for the exchanges page.

PR #3658, opened 2021-04-26
bitcoin-paper: add braille version

RandyMcMillan:


RandyMcMillan:
REF: https://github.com/neatnik/braille-bitcoin-whitepaper

Screen Shot 2021-04-26 at 1 03 12 PM

Cobra-Bitcoin:
Wow, this is really great. Thanks for finding this!

This also has me thinking, is there an audio version of the whitepaper somewhere we can use? We'll need one that's well narrated. That way we have both a braille version, and an audio version too.

RandyMcMillan:
i can do that. Any limit on file size?

RandyMcMillan:
Another option is to add a bash command to generate an audio file from a basic text file - this will allow you to create it on the server and not add a lot of weight to the repo.

RandyMcMillan:
Or we can embed a youtube video?

cryptohakka:
I'm very interested in this attempt. I'll try to make Japanese braille version.
Bitcoin is for everyone!

PR #3677, opened 2021-05-18
Fix download page to use pruned size

CrazyPython:
This is a follow up to #3624. (In summary, a pruned node is a fully verifying node, capable of strengthening Bitcoin in the hands of economic actors and away from custodians that operate full nodes. It uses significantly less disk space, which encourages more people to run full nodes.)

> You should make sure that you have enough bandwidth and storage for the full block chain size (over 350GB).

This pull request changes it to the following:

> You should make sure that you have enough bandwidth and storage for the block chain size (7GB).

- Changes "full block chain size" to "block chain size"
- Changes "over 350GB" to "7GB"

This only changes the English version.

There is an open draft pull request to my version to add translations: crazypython/Bitcoin.org#1

Translators are welcome to translate this change. I would do it myself (removing words that I guess mean "over" and "full"), though there is a chance I will make a mistake while doing so.


Cobra-Bitcoin:
LGTM.

PR #3692, opened 2021-06-19
update IRC link from freenode to libera chat

jlopp:
Hopefully this is the correct way to propose a change to this transfinex template. Noticed that this page was still linking to freenode even though folks have migrated to libera.


ndalliard:
there are a lot more references to freenode througout the site, not only in the english translation

bitcoin-public-domain:
Freenode should stay for various reasons including **historical**, originally it was created by Satoshi Nakamoto. 
At the moment @luke-jr very nicely moderating the `#bitcoin` channel on freenode. 

Libera sucks. :vomiting_face:

luke-jr:
@bitcoin-public-domain For better or worse, the `#bitcoin-core-dev` channel which this is about only exists on Libera.

PR #3701, opened 2021-07-01
Add js

Cobra-Bitcoin:


rebroad:
Can we have some reference to why this is blocked?

rebroad:
Removing the download to bitcoin is overkill.

https://twitter.com/CobraBitcoin/status/1410889281904709634?ref_src=twsrc%5Etfw

Based on the above rationale this could just as easily be interpreted to mean that all UK ISPs should close down (given that they provide a way to use VPNs).

PR #3731, opened 2021-08-24
Remove BLW

btcontract:
BLW has been deprecated for reasons explained here: https://lightning-wallet.com/posts/blw-is-deprecated/

If that's fine in about 3 months I also plan to submit a PR to revive Simple Bitcoin Wallet here since it's alive again and is backwards-compatible for existing users.


crwatkins:
@btcontract thanks for the update.   Could you also remove the blw.png icon file?

crwatkins:
LGTM

PR #3746, opened 2021-10-04
Update en.yml - grammar fix

git-sgmoore:
Grammar fix at line 300


crwatkins:
This change looks like a good idea.  Perhaps you could close #3745 and combine it with this PR.  Also, I think that entire description could use some better word smithing if you are up to it.

git-sgmoore:
Thank you for the response. I'll do that, and I'm happy to take a look at the rest of the section.

PR #3793, opened 2021-11-27
Add SBW

akumaigorodski:
This adds version 2 of Simple Bitcoin Wallet which is under active development again.


crwatkins:
@btcontract Thanks for your submission!   It looks like you are going back to your roots.   I'm fully expecting the next version to have a coin selection GUI that involves flicking disks around on the screen!  But seriously, it looks like another solid submission from our most prolific wallet submitter.

It might be a bit early for this, but I have some suggestions for your wallet description text.   Since you've decided to submit under the name SBW, I would recommend "SBW (Simple Bitcoin Wallet)" instead of what you have so that you lead with the proper name of the wallet.    I would also suggest removing "On top of that". I'm not really sure that adds anything and less words moves focus to the important ones.    One last suggestion is somehow changing the word "store".  We try not to lead users to believe that their coins are actually stored on their phone.

crwatkins:
Thanks @btcontract.  I apologize I wasn't very clear in what I wrote above.   I was suggesting replacing "Simple Bitcoin Wallet (aka SBW)" in the wallet description with "SBW (Simple Bitcoin Wallet)" since SBW is what you've chosen as the proper name.  You might want to start the description with the proper name.

crwatkins:
@btcontract Thanks!

PR #3810, opened 2022-01-10
Add Prokey Optimum to wallets

northmz:
This pull request adds Prokey Optimum to hardware wallets list.


crwatkins:
Thanks for the resubmission!

I had a number of unanswered questions in your [previous submission]( https://github.com/bitcoin-dot-org/Bitcoin.org/pull/3625) mostly related to your firmware authentication methods.  Have those changed significantly?   If so, could you point us to documentation on the new system or describe them?  If not, could you follow up on the questions in your previous submission?

northmz:
> Thanks for the resubmission!
> 
> I had a number of unanswered questions in your [previous submission](https://github.com/bitcoin-dot-org/Bitcoin.org/pull/3625) mostly related to your firmware authentication methods. Have those changed significantly? If so, could you point us to documentation on the new system or describe them? If not, could you follow up on the questions in your previous submission?

Thanks for taking your time to review it. Yes, We have changed it based on listing criteria. Now each device has its own dedicated key which will be written in OTP memory section in production lane, so there is no hard coded authentication key in the source code that prevented us making it 100% opensource.

Here is the code if you would like to have a look: https://github.com/prokey-io/prokey-optimum-firmware/blob/4f9aba2b4036832911b6d357bf31f95c6010af02/auth.c#L412

crwatkins:
1. Does each device get its own unique secret key? (just confirming I understand what you said above)
2. Can you describe, preferably in a few lines of pseudocode, how downloaded firmware is verified?
3. Could you describe what you believe are the pros and cons of your approach compared to traditional public key digital signatures for code signing?

northmz:
> 1. Does each device get its own unique secret key? (just confirming I understand what you said above)
Yes

> 2. Could you describe, preferably in a few lines of pseudocode, how downloaded firmware is verified?

Firstly, the device and server will perform mutual authentication using a unique secret key, which we call AuthKey, after that a session key will be generated and the firmware will be AES encrypted by the server using this session key and then will be sent to the client-side. Secondly, the device will decrypt the received data using the same session key and will install the new firmware. 

> 3. Could you describe what you believe are the pros and cons of your approach compared to traditional public key digital signatures for code signing?

by this method, we can always be sure that the device is genuine because we perform mutual authentication not only when we want to install new firmware but also each time the Prokey device wants to be connected to our wallet interface and a fake device can not pass the mutual-authentication process. This can prevent the risk of evil maid attacks and any other person or company cannot create fake devices.

crwatkins:
> Firstly, the device and server will perform mutual authentication using a unique secret key, which we call AuthKey, after that a session key will be generated...

Could you explain the algorithm, preferably in pseudocode,  that is used to generate this session key?

> ... the firmware will be AES encrypted by the server using this session key and then will be sent to the client-side. Secondly, the device will decrypt the received data using the same session key and will install the new firmware.

Could you explain how you ensure data integrity since encryption does not provide integrity?

> by this method, we can always be sure that the device is genuine because we perform mutual authentication not only when we want to install new firmware but also each time the Prokey device wants to be connected to our wallet interface and a fake device can not pass the mutual-authentication process.

What prevents a customer from extracting an AuthKey from OTP memory and using it in other devices, since it seems to be a goal to prevent that?

northmz:
> Could you explain the algorithm, preferably in pseudocode, that is used to generate this session key?

1.	The server requests a Random from the device (RndD) and the device will send its SN + RndD.
2.	Server encrypts RndD using the corresponding unique secret key and returns Enc(RndD) + a new Random number (RndS).
3.	First, the device checks if Enc(RndD) is correct then it will send SN + Enc(RndS).
4.	Server checks if Enc(RndS) is correct then the session key will be generated, SessionKey = Hash(Hash(Hash(Key + RndD + RndS))).
5.	And in the last step, Server will send Hash(SessionKey) to the device to make sure the process completed successfully.

```C++
byte[32] DeviceKey;
byte[16] DeviceRandom = Randomize();
var serverRespond = SendToServer(SerialNumber + DeviceRandom) // Server will return AES256(DeviceRandom, DeviceKey) and ServerRandom
if (serverRespond.EncodedRnd != AES256(DeviceRandom, DeviceKey))
return false;

var SessionHash = SendToServer(AES256(serverRespond.ServerRandom, DeviceKey))

byte[32] SessionKey = Hash(Hash(Hash(DeviceKey + DeviceRandom + ServerRandom)))

if (Hash(SessionKey) == SessionHash)
return true;
```



> Could you explain how you ensure data integrity since encryption does not provide integrity?

First, please note that the bootloader will not start firmware installation before completion of mutual authentication. 
Secondly, if attacker uses a man-in-the-middle attack, they won’t able to install a fake firmware on the device again because they don’t know the session key to encrypt the firmware. However, if a firmware is encrypted with a wrong key, nothing bad will happen due to wrong decrypted data that cannot be executed on ARM processor and First boot and Bootloader will not touch during firmware install. Worth mentioning that Bootloader also checks the 4 first bytes if it is match to Initial Stack Pointer address (the first 4 bytes of flash is always Initial Stack Pointer) before finishing the installation. We have plan to improve this checking in the near future by adding Enc(Hash(firmware)) to the end of firmware content.

> What prevents a customer from extracting an AuthKey from OTP memory and using it in other devices, since it seems to be a goal to prevent that?

Please note that:
1.	The bootloader and firmware will not send OTP content out for any reason. It’s only used as a key for encryption.
2.	Because of device protection, accessing through JTAG, IAP or any interface is denied. 
3.	To install a fake firmware, mutual authentication is needed, so without having OTP, it is not possible.
4.	Finally, If we assume that OTP leaks for any reason, the attacker can produce another fake device with fake bootloader that should send exact same Serial Number to server(serial number of original device) , However, Once the official firmware get installed from the our website on the new fake device, a true device serial number will be used(our firmware reads the UUID of STM32 chip) that is not available in our server and device can not pass mutual authentication and generate any wallet.

crwatkins:
I noticed that you made some [changes](https://github.com/prokey-io/prokey-optimum-firmware/pull/25) to the algorithm you described above yesterday and more changes a couple weeks ago.  Is the algorithm you described above the algorithm currently in use in production or the one you just updated in your repository?

crwatkins:
> We have plan to improve this checking in the near future by adding Enc(Hash(firmware)) to the end of firmware content.

This is a significant omission.  Encryption alone does not provide integrity.  On this issue alone, I cannot recommend Prokey Optimum for listing based on the criteria.  However, what is more troubling is that Prokey seems to have forked a well established and vetted signature system and replaced it with a home brew system containing multiple novice cryptographic mistakes.  I would highly recommend getting an expert cryptographic review of the system and resubmitting not sooner than 12 months after improvements have been in production.

northmz:
>This is a significant omission. Encryption alone does not provide integrity.

It is obvious that your sentence is true in general. But as I mentioned in my previous message, the wrong encrypted data neither gets executed by the bootloader nor is able to be run on the ARM processor. 

However, signed firmware also does not itself provide integrity. For example, I can sign a firmware with a wrong key and program a Trezor HW (during the writing process to the flash memory, the hash of data only be checked, not the signature), right? But prior to firmware boot, their Bootloader will check the signature of the programmed firmware and show an error/warning message. So, this integrity issue is there too.

> ...replaced it with a home brew system containing multiple novice cryptographic mistakes.

The challenge-response process and mutual authentication and encrypt the firmware with a session key is not homebrew, I refer you to the below documents an standards. and I am wondering to know what is cryptographic mistakes? Please point out the mistakes directly. 
- ISO/IEC 7816-4 section 7.5
- DESFire EV1/EV2 Authenticate section
- https://en.wikipedia.org/wiki/Challenge%E2%80%93response_authentication

> …resubmitting not sooner than 12 months after improvements have been in production.
We always try to improve our system in both new features and security. However, please let us know if this 12 months is part of the criteria or your personal concern.
> …Is the algorithm you described above the algorithm currently in use in production or the one you just updated in your repository?
The algorithm we described above is not in production right now and will be published with next firmware update.

crwatkins:
@northmz 

> Please point out the mistakes directly.

I want to be clear that I did not do a comprehensive cryptographic review of the codebase; that is why I recommended one.   Relying on encryption alone without providing an integrity check is significant.  In the [previous PR](https://github.com/bitcoin-dot-org/Bitcoin.org/pull/3625) I had asked questions about algorithms that were never answered which makes it hard to evaluate.

> However, please let us know if this 12 months is part of the criteria or your personal concern.

As I've mentioned before in regards to other wallets, I don't believe that any of the criteria were intended to permanently disqualify listing even though no specific time periods were mentioned.  Given the fact that the improvements described above are not yet implemented in production, this is going to take some time.   I've spent time on this review twice within the last year and I have very limited resources and I need to be fair to other wallets.  We're all just contributors here and reviews can be performed by any respected contributor whenever they have the time.

Issue #3852, opened 2022-03-23
Max number of outgoing connections is now 10, instead of 8


nodet:
Running `Bitcoin Core version v22.0.0 (release build)`, I can see that the max number of outgoing connections is 10.

```
$ cat ~/.bitcoin/bitcoin.conf
datadir=/mnt/disk/blockchain
blocksonly=1
$ bitcoin-cli -getinfo
{
  "version": 220000,
  "blocks": 728655,
  "headers": 728655,
  "verificationprogress": 0.9999964627698886,
  "timeoffset": 0,
  "connections": {
    "in": 5,
    "out": 10,
    "total": 15
  },
  "proxy": "",
  "difficulty": 27452707696466.39,
  "chain": "main",
  "relayfee": 0.00001000,
  "warnings": ""
}
```

The full node page at https://bitcoin.org/en/full-node has a number of references to this number being 8. I'll submit a PR shortly.

tadeubas:
Hi @nodet , have you realized that this project is about the bitcoin dot org site and not about the bitcoin core software?

Saedx01x01iiiio:
"{+''

amimahd:
Very good

PR #3853, opened 2022-03-23
Max outbound connections is now 10

nodet:
As of 22.0, the max number of outgoing connections is 10.
C.f. https://github.com/bitcoin-dot-org/Bitcoin.org/issues/3852 for details.


Cobra-Bitcoin:
LGTM. Thanks for noticing this.

PR #3856, opened 2022-03-26
Add Armenian and Serbian whitepapers

Cobra-Bitcoin:
Add Armenian and Serbian translations of whitepaper


Cobra-Bitcoin:
Thanks it's fixed now.

PR #3886, opened 2022-05-22
Add tauros mexican exchange

cpleonardo:
Changes to be committed:
	modified:   _templates/exchanges.html


Cobra-Bitcoin:
Yeah, but why? It would be nice if PR's like this included some justification for why to add a particular exchange, such as volume, customers, social media activity, etc? In general we want to know that exchanges have some substance behind them before we recommend them to users.

PR #3889, opened 2022-06-08
add a new font family for rtl languages

ashkanjalaliQ:
Vazirmatn is a user-friendly font for Arabic and Persian languages and has been published as open source.
I added this font for rtl languages.

Vazirmatn source: [https://github.com/rastikerdar/vazirmatn](https://github.com/rastikerdar/vazirmatn)


Cobra-Bitcoin:
Why is it better?

ashkanjalaliQ:
because Vazir font is more legible than the current font for Persian and Arabic languages.
so Vazir font is a better choice

PR #3902, opened 2022-07-06
Replacing dead domain with webarchive version

hugonun:
iamsatoshi.com has been down since February 2022, unsure if it will come back.


Cobra-Bitcoin:
LGTM.

Issue #3022, opened 2019-06-20
Recommending wallets ideal for new users in new Wallet selector


wbnns:
As part of the wallet selector in the Wallet pages that are currently being redesigned, a User must self-identify as "New" or "Experienced". After selecting the Operating System, choosing "Experienced" will present all available wallets. Choosing  "New" will present wallets for the selected operating system that are ideal for new users:

![image](https://user-images.githubusercontent.com/1130872/59835462-fbecb300-9349-11e9-8782-42c135798c16.png)

This issue is an opportunity to share feedback/comments on what wallets should be recommended to new users. 

@crwatkins [mentioned](https://github.com/bitcoin-dot-org/bitcoin.org/issues/2861#issuecomment-472258443) that a potential shortlist to consider recommending could be:

+ Bitcoin Wallet
+ BRD
+ BitPay
+ Electrum
+ GreenAddress

Feedback/comments welcome.

Cc: @Cobra-Bitcoin, @crwatkins, @maxwellmons, @alexcherman 
Related: #2861, #3017

wbnns:
We should try to qualify what makes wallets ideal candidates for new users so that as wallets are added to the site, there is an objective methodology for determining if they can be categorized a such.

crwatkins:
To come up with the above list, I created a list of items with a yes/no rating for each wallet.

1. Simple features
2. Advanced features
3. Account Required
4. Confusing
5. Hardware

To make the list, a wallet had to have 1 and could not have 3-5.  The problem with this system is that my ratings here are highly subjective.  My goals were easy and fast onboarding of new users.

**Simple features** mean that the wallet has very straightforward ways to perform simple send, receive, and balance functions.  The wallet had to have obvious discoverable options using basic terminology with no jargon.  For example, wallets that may have a button labeled **Sign** or **Broadcast** instead of something simpler such as **Send** did not qualify.

**Advanced features** meant having features beyond simple sending and receiving, or perhaps having those **Sign** and **Broadcast** buttons.   Having advanced features did not disqualify a wallet, as long as the wallet also had simple send and receive functions.  One might say this item is not necessary, but it helped to categorize wallets into simple wallets with advanced features or just plain advanced wallets.

[...]

PR #3946, opened 2022-10-11
Add Wasabi Wallet for New Users

thibistaken:
Wasabi is currently only recommended for advanced users. Providing Wasabi Wallet as a recommendation for new users makes sense now since the release of Wasabi 2.0 earlier in June 2022. The on-boarding user experience has been simplified to abstract away complexity from the end user, without compromising on privacy. A total redesign of the client was released allowing less technical users to use a bitcoin wallet that has privacy by default. More can be done to make it even simpler, of course, but it is a step in the right direction for mainstream bitcoin privacy.

### DIFF
--- a/_wallets/wasabi.md
+++ b/_wallets/wasabi.md
@@ -6,6 +6,7 @@ id: wasabi
 title: "Wasabi Wallet"
 titleshort: "Wasabi"
 compat: "desktop windows mac linux"
+user: beginner
 level: 3
 platform:
   - desktop:


crwatkins:
I definitely agree that the Wasabi 2.0 on-boarding user experience has been greatly improved.   The UX on the new version is impressive.  However, my biggest concern is having Coinjoin turned on for new users by default.   Coinjoin can improve fungibility in the long term, but depending on the exact situation, there have been issues with some services accepting outputs that have recently been in a Coinjoin.  I personally believe that understanding the landscape of the fungibility of Coinjoin outputs currently requires expert analysis and is not for beginners.

Likewise, complete understanding of how the coordinator fee works is not obvious (I don't actually know myself), and even though the fee is disclosed, some new users may be surprised by the default charge even before they send funds.

The actual criteria for the beginner designation has not been well codified, but the criteria that is being used for the current listings is described [here](https://github.com/bitcoin-dot-org/Bitcoin.org/issues/3022).   One disqualification for the beginner designation is a long legal agreement that is required to be accepted to use the wallet.  I can certainly understand the need for such an agreement in some situations, particularly when an organization is providing Coinjoin services, but I don't believe that being a legal expert or having to consult with a lawyer is consistent with the goal of easy and fast onboarding of new users.  I have no idea whether that agreement would be required if Coinjoin were not enabled by default, but that is something to consider.

Because of the above issues, I can't personally recommend the beginner designation at this time.

nopara73:
Thanks, this is very insightful! Let me try to parse point by point:

**On long legal agreement.** Just to make sure you're talking about the [2.0 agreement](https://github.com/zkSNACKs/WalletWasabi/blob/master/WalletWasabi/Legal/Assets/LegalDocumentsWw2.txt) and not the [1.0 agreement](https://github.com/zkSNACKs/WalletWasabi/blob/master/WalletWasabi/Legal/Assets/LegalDocumentsWw1.txt)? The 2.0 has been simplified greatly. That being said, if this seems long I'll examine if we can do something about it.

**On refusing coinjoin outputs.** Although it's not as common as Twitter may make you think, it is indeed a problem and can potentially get worse. Generally of course we'd love to improve this, but we're short on ideas, so any suggestion is welcome!

**On coinjoins by default.** There's a philosophical constraint here. The design decision here comes from the notion that, by default the usage of money should not involve the knowledge of third parties. Privacy by default. The point is taken though, unfortunately this is not something we can compromise on.

**On fee transparency.** Currently every time the user creates a wallet we have a few lines explaining the fees are 0.3% coordination fees + network fees. I'm not sure what's the problem here. Can you elaborate? Nevertheless we're currently working on improving the UX here though. For reference here're the details: every time someone receives money to his wallet, he'll be charged a 0.3% constant fee with 3 exceptions:
- no fee is taken if the UTXO is under the pleb limit (I think it's 1 million satoshi)
- no fee is taken if the money you received is coming from a coinjoin
- no fee is taken if money you received is coming one hop from a coinjoin (this also means if you receive money from a friend who uses Wasabi, then no fee is taken)

### Summary

| Point                     | Status                                                     |
|---------------------------|------------------------------------------------------------|
| long legal agreement      | may be a misunderstanding, maybe it can be fixed |
| refusing coinjoin outputs | would love to fix, not sure how                            |
| coinjoins by default      | wont fix                                                   |
| fee transparency          | currently we're in the process of fixing it, more feedback is welcome                        |

crwatkins:
> On long legal agreement. Just to make sure you're talking about the [2.0 agreement](https://github.com/zkSNACKs/WalletWasabi/blob/master/WalletWasabi/Legal/Assets/LegalDocumentsWw2.txt) and not the [1.0 agreement](https://github.com/zkSNACKs/WalletWasabi/blob/master/WalletWasabi/Legal/Assets/LegalDocumentsWw1.txt)? The 2.0 has been simplified greatly. That being said, if this seems long I'll examine if we can do something about it.

The 2.0 agreement; the one in the current version of the code.  The file is over 250 lines long and contains some fairly complex legal language.  For example, I spent quite a while trying to decide if a user can purchase legal fireworks with the wallet or whether it only applied to illegal fireworks (and I still don't know, I'm not a lawyer).  I got stuck for quite some time during onboarding.   In addition, the agreement uses terms such as UTXO and Coinjoin which in my opinion are not defined anywhere near sufficiently for a new user to understand what they are agreeing to.

I do want to be clear that I'm not expressing an opinion on whether there should or should not be such an agreement.  I'm just stating an opinion that this is not stuff that we can expect beginners to be able to understand.

As aside, which makes this particular issue even more difficult during onboarding, the app does not make the agreement available for searching, downloading, or even text selection for copying.  The links in the agreement are not "live" and cannot be followed nor copied.

> On fee transparency. Currently every time the user creates a wallet we have a few lines explaining the fees are 0.3% coordination fees + network fees. I'm not sure what's the problem here. Can you elaborate? 

New Bitcoin users probably have never heard of a Coinjoin before, let alone have any idea what a coordinator is or what it does or how often a coordinator charges them (every month?) or which funds it applies to.

> For reference here're the details: every time someone receives money to his wallet, he'll be charged a 0.3% constant fee with 3 exceptions...

I think that explanation is much easier to understand than what I read during onboarding.

Just one more note along these lines.  Beginners are going to have questions and need some help.  The current support path seems to be through GitHub which requires a GitHub account.   Signing up for GitHub and understanding the issue  tracking system is not exactly for beginners.

mayreed11 (approved):

crwatkins:
As of the last discussion, it doesn't seem like Wasabi Wallet qualifies for the beginner flag and there has been no further discussion for over a year.   If that situation changes, this PR can be re-opened or resubmitted.

nopara73:
Unfortunately Wasabi doesn't default coinjoin anymore which IIRC addresses the issues raised here. Considering it's by far the most user friendly desktop wallet out there, this should be reopened?

crwatkins:
@nopara73 Thanks for the update.   Let's reopen for comments.

thibistaken:
Do you need anything else to get this merged @crwatkins? Lmk

PR #3959, opened 2022-11-01
update features and transparency for electrum-android

accumulator:
Electrum-android is built reproducibly and supports 2fa and multisig


Cobra-Bitcoin:
@crwatkins Should be merged, or no?

PR #3999, opened 2023-02-22
Just adding a note about the config file location and running as admin on windows.

bn-l:


bn-l:
Without running as administrator I get the error "error opening bitcoin-qt.exe". I am running as a non-admin user on windows (a somewhat strange thing) and it is likely it was not tested in this situation.

bn-l:
Also, I chose a custom data location and the conf file was created there, not in appdata--this tripped me up.

PR #4010, opened 2023-03-17
Bitcoin Core Download: redirect to BitcoinCore.org download page

harding:
All Bitcoin Core download links on the website currently use version 22.0, which was released in September 2021.  Newer releases are being published on BitcoinCore.org, so this PR updates the download box to redirect to that website.

I've previewed this update locally; the updated download box looks like this:

latest

(The main difference from the existing box being that it doesn't list a specific version.)

When running tests, I encountered pre-existing errors unrelated to this change so I was not able to fully test.


Cobra-Bitcoin:
Addressed by https://github.com/bitcoin-dot-org/Bitcoin.org/pull/4078.

PR #4036, opened 2023-05-03
Add Green wallet

alessandro-saglimbeni:
This PR adds the Blockstream Green wallet for listing on bitcoin.org.

Most scores are tentative to the best of my interpretation, and deducted by comparing to other wallets currently listed. 

The scores prone to ambiguity or controversy are currently followed by a `# comment` for the purpose of facilitating the review.


crwatkins:
Thanks for your submission!

A quick glance over the comments in your submission seems like your scoring is consistent with other wallets.  Two notes:

* The Android wallet is scored `checkgoodtransparencydeterministic` however the folks at [Wallet Scrutiny](https://walletscrutiny.com/) have been [unable](https://walletscrutiny.com/android/com.greenaddress.greenbits_android_wallet/) to verify version 4.0.4.   Do you know why that is?

* The beginner tag (`user: beginner`) has been included.   The actual criteria for the beginner designation has not been well codified, but the criteria that is being used for the current listings is described [here](https://github.com/bitcoin-dot-org/Bitcoin.org/issues/3022). One disqualification for the beginner designation is a long legal agreement that is required to be accepted to use the wallet. I noted that new users must agree to the Terms of Service and Privacy Policy of over 4,000 words. I can certainly understand the need for an agreement in some situations, however I don't believe that being a legal expert or having to consult with a lawyer is consistent with the goal of easy and fast onboarding of new users.

alessandro-saglimbeni:
Hey Craig, thanks for the quick feedback!

I just pushed a couple updates:
- I substituted the score for the deterministic build on Android with the open source grade. There is a [bug](https://github.com/Blockstream/green_android/issues/169) that requires Google to update AGP, when they'll release that fix we should be able to pick it up and hopefully get back to have deterministic builds. Also iOS needs some fixes in the build pipeline before we can get back to have deterministic builds. I'll provide an update later on when we release a fix for this
- I removed the `user: beginner` tag for the moment, we'll explore how to improve this UX. We definitely can't forego the ToS acceptance, but we could improve how we communicate it to the users

crwatkins:
@alessandro-saglimbeni The submission states:
```
validation: "checkpassvalidationspvservers" # txs can be SPV validated (a la electrum) using a list of default electrum servers, and if a custom electrum server is set then it gets used for SPV validation
```

Is that currently the case?   It appears that by default my Blockstream Green always connects to the Electrum server at blockstream.info.  If I'm not mistaken, then the validation score should instead be *checkfailvalidationcentralized*.

(There are certainly reasons these days that one might want to avoid using random Electrum servers, but that's a different issue.)

crwatkins:
In the description text, would you want to use "multisig" instead of "multisigs"?

crwatkins:
> what is the meaning of these other values for validation?

You can find the "source" [here](https://github.com/bitcoin-dot-org/Bitcoin.org/blob/master/_translations/en.yml).

> I'm not sure I understand, I haven't included a description, I don't understand what you are referring to

I was referring to the change you made to [_translations/en.yml](https://github.com/bitcoin-dot-org/Bitcoin.org/pull/4036/files#diff-4fffe94a93ec54d2a87539cb7e41ed85c5d0c159380e7564346df9e119af2783)

alessandro-saglimbeni:
> `validation: "checkpassvalidationspvservers"`
> Is that currently the case? It appears that by default my Blockstream Green always connects to the Electrum server at blockstream.info. If I'm not mistaken, then the validation score should instead be checkfailvalidationcentralized.

Ok then why wouldn't it be `    checkneutralvalidationvariable: "Variable validation"` rather than `    checkfailvalidationcentralized: "Centralized validation"`?

Advanced users can set their own preferences around SPV, as well as their Electrum server(s) of choice, it is varaible rather than fixed and centralized.


I've fixed the typo, sorry didn't understand what you were referring to

crwatkins:
> Ok then why wouldn't it be     checkneutralvalidationvariable: "Variable validation" rather than     checkfailvalidationcentralized: "Centralized validation"?

These tags control the description that is displayed for the wallet.  Variable validation is used for hardware devices such as Jade that rely on an external app for validation.   The text displayed is:
```
Payment validation features are provided by the software wallet you use with this device.
```
Tags are chosen to reflect the default behavior of the wallet.  There have been discussions in the past to expand the list of features available, but multiple validation options are not currently listed.

alessandro-saglimbeni:
Ok understood, I've changed the validation to `checkfailvalidationcentralized` in a2e9766d3115cf4ba1747ff12d6d3a2ed065962e

LMK if there's anything else pending, thank you!

crwatkins:
The comment in the file reads
```
# txs can be SPV validated (a la electrum) using a list of default electrum servers, and if a custom electrum server is set then it gets used for SPV validation
```
This comment is of course not shown, so it is not critical, but for my own information, is it correct?  How would one enable Blockstream Green to choose from the "list of default electrum servers"?     Is that list curated in any way?

alessandro-saglimbeni:
Fixed the comment around SPV functioning in 0943fbe593999e3fe8e05ed4030cb5718a400a9a

Cobra-Bitcoin:
LGTM. Merged.

It's nice to see a wallet developer really engage with the review process and communicate effectively.

PR #4071, opened 2023-07-10
add btcmap for merchant discovery

secondl1ght:
Hello, I have submitted this PR to replace `coinmap.org` with `btcmap.org` on the https://bitcoin.org/en/spend-bitcoin page and also add BTC Map to the https://bitcoin.org/en/resources page.

BTC Map better aligns with bitcoin and bitcoin.org because it is open source and bitcoin-only. In addition to this the BTC Map dataset is also higher quality due to the fact that we use OpenStreetMap to leverage the global mapping community. In contrast the coinmap dataset is largely considered stale and abandoned. Users are also prompted to purchase altcoins when navigating the coinmap website. 

BTC Map has become the main bitcoin map solution since it's inception in the past year with major bitcoin wallets such as Wallet of Satoshi integrating with us.

There are many other reasons as to why listing BTC Map over coinmap would be an improvement but I have outlined a few key points here. Please let me know if you require any further information.

Thanks! :)


Cobra-Bitcoin:
LGTM. I'm re-running the tests and will merge once they pass. Thanks.

PR #4080, opened 2023-08-10
Update buy.html

martcpp:
Added strong and red colour the thirds party system moonpay for easy seeing by users to avoid misunderstanding

### DIFF
--- a/_templates/buy.html
+++ b/_templates/buy.html
@@ -33,7 +33,7 @@ end_of_page: |
   	</div>
 </div>
 	  <div style="margin-top: 50px;text-align: center;">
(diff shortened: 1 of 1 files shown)


martcpp:
This will make it noticable by user so they understand that using moonpay is third-party

martcpp:
This has been major issue to buyer after they had done thier transaction believing it from bitcoin with knowing it from a third party

martcpp:
These will reduce fraud on third party

martcpp:
This will save customer the stress to know it s third party service

AugustoResende (approved):

PR #4111, opened 2023-09-27
Remove the words "open source" from the description of Coldcard

ai-oleynikov:
The "Commons Clause" License, [used by Coldcard's firmware](https://github.com/Coldcard/firmware/blob/master/COPYING-CC), doesn't meet the criteria of an open-source license. Even the authors of the license themselves don't claim it to be open source.

https://commonsclause.com/#faq

> Is this “Open Source”?
> No.

So we shouldn't call Coldcard open source.


AugustoResende (approved):

nvk:
If it's just a question of words. "Verifiable Source Code" is acceptable. As long as the indicating icon stays screen.

This is not a reference to the person crating this PR, but I find the motivation for this change often from bad faith as ppl try to paint the device as if the source is not available. Let a lot the endless wars over this term being unsettled.

crwatkins:
@ai-oleynikov How about this using the words suggest by @nvk?

Coldcard is an easy to use, ultra-secure and affordable hardware wallet with verifiable source code that is easy to back up via an encrypted microSD card.

crwatkins:
Would you like to update the PR?   You don’t have to make changes to all the language files as the Transfix team can do that.

crwatkins:
Hey @ai-oleynikov Would you like to update your PR?

crwatkins:
I've created a new PR #4235 to implement the change discussed above.

Thanks fo @ai-oleynikov for proposing this.

PR #4118, opened 2023-10-11
Remove the claim that deterministically built wallets are open-source

### LINKED ISSUE #4110: The transparency score misuses the term "open source"
https://github.com/bitcoin-dot-org/Bitcoin.org/blob/fbc8db1a32b63042f9a723c9a74bd8f54606ef11/_translations/en.yml#L340

The source code being readable and the build process being deterministic do not necessarily mean that the software is open source as per the [official OSI definition](https://opensource.org/osd/). For example, both Coldcard's and Passport's firmware can be built reproducibly, but they aren't open source. I suggest we correct this description to say that the code is just publicly available, e.g.:

```diff
- This wallet is open-source and built deterministically.
+ The source code for this wallet is public, and the build process is deterministic.
```

ai-oleynikov:
Related to #4110

This PR provides a correction to RU, EN, and ES translations. Further work is needed for other languages, as I only speak these three.


crwatkins:
The English en.yml LGTM.  The Transfix team can deal with the translations.

angel0702 (approved):

PR #4121, opened 2023-10-20
Add Bitcoin Core 25.1 release

fanquake:
Similar to #4078.
See also https://github.com/bitcoin-core/bitcoincore.org/pull/991.

PR #4146, opened 2024-01-04
Add Bitcoin Core 26.0 release

lateminer:
Similar to #4121.


fanquake:
Note that v26.1 is also now available.

fanquake:
I've closed #4121 in favour of this one. Note that v27.0 is also now available.

PR #4179, opened 2024-04-08
Add SeedSigner as HWW

omerskywalker:
This PR adds SeedSigner as a HWW to Bitcoin.org Wallets section

PR #4194, opened 2024-04-30
Add environment score for DIY hardware wallets

crwatkins:
This PR is to add a new option for the hardware wallet environment score.  This score is meant for hardware wallets that allow the user to provide their own hardware.  It is important to note:
* This PR alone makes no visible change to the site.  This update is for use on future wallet listings.
* Historically the community has been hesitant to add new scoring that might only apply to only one or two wallets.
* Currently there are two open PRs for wallets ([SeedSigner](https://github.com/bitcoin-dot-org/Bitcoin.org/pull/4179) and [AirGap Vault](https://github.com/bitcoin-dot-org/Bitcoin.org/pull/4019)) that this could apply to, and other wallets exist that could qualify (e.g. Specter-DIY, Krux).
* The icon that will appear to users in the comparison charts for this new score will be labelled *Acceptable*.
* The term "hardware wallet" may be a bit of a misnomer, but that's an issue for another time.

### DIFF
--- a/_translations/en.yml
+++ b/_translations/en.yml
@@ -352,6 +352,8 @@ en:
     checkfailtransparencynewtxt: "This wallet has not been tested and publicly reviewed by a significant number of people. This means this app might be more at risk of hiding dangerous code or doing something you wouldn't agree to."
     checkgoodenvironmenthardware: "Very secure environment"
     checkgoodenvironmenthardwaretxt: "This wallet is loaded from a secure specialized environment provided by the device. This provides very strong protection against computer vulnerabilities and malware since no software can be installed on this environment."
+    checkpassenvironmentbyod: "User sourced environment"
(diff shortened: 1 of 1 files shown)


nvk:
These should ideally be on it's own section, "DIY & Other Hardware" with appropriate warnings about trade offs, similar to how OPENDIME was not qualified to be added to main section.

Cobra-Bitcoin:
I'm supportive of this, it's good for now, though I think in the near future we need to reform completely how we present wallets. I was thinking we find someone with strong UX knowledge to have a look at that part of the site and come up with suggestions. I don't mind doing the web stuff and restructuring if needed, but I think there is a way to present information more intuitively on those pages that we're missing.

It feels like there are lots of different types of wallets coming out now and it's getting increasingly hard to put them into neat categories.

crwatkins:
I would be in favor of that.

PR #4195, opened 2024-04-30
Prevent firmware downgrades

crwatkins:
There has been an optional criterion for many years on hardware wallet listings to avoid attacks by preventing the downgrading of firmware to a potentially vulnerable version.   Some currently listed hardware wallets simply prevent downgrades while others allow the capability of the current firmware to prohibit the installation of specified versions.

I propose making this a required criterion.   The original criterion was slightly reworded to take into account both implementations described above.


Cobra-Bitcoin:
LGTM. Will merge soon unless anyone has objections.

PR #4233, opened 2024-06-06
Create Tangem Wallet

sidorovpavel:
Add Tangem Wallet to wallet list


crwatkins:
Thanks for your submission.   Could you update the "source" entry to point to a specific repository that contains the hardware source instead of pointing to an entire organizaiton?

Have you carefully verified that Tangem Wallet meets all the listing criteria specified [here](https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/docs/managing-wallets.md)? Do you have any questions on any of the criteria?

crwatkins:
@sidorovpavel It has been [reported ](https://cryptoguide.tips/hardware-wallet-comparisons/) that Tangem Wallet 
* Reuses the same address for each Bitcoin transaction in the default mobile app
* Does not support any apps other than the default mobile app

Is this currently accurate?

sidorovpavel:
We reuse the same address for each bitcoin transaction for simplicity.
Yes we does not support any other apps.

crwatkins:
The [listing criteria](https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/docs/managing-wallets.md) include 
* Avoid address reuse by displaying a new receiving address for each transaction in the wallet UI
* Avoid address reuse by using a new change address for each transaction

While technically, hardware wallets don't control the selection of receiving addresses (the companion app does) and historically these criteria have not been applied to hardware wallets (since hardware wallets are reviewed independent of the companion app), all of the hardware wallets that are currently listed support multiple companion apps which do not reuse addresses.   Since Tangem only has one companion app which reuses addresses, recommending Tangem hardware would be much the same as recommending address reuse.

You may want to review [Bitcoin wiki address reuse](https://en.bitcoin.it/wiki/Address_reuse).  Tangem address reuse not only damages the security and the privacy of its users, but it also damages they privacy of the entire community. It is very concerning that any wallet would be doing this in 2024 as there have been no wallets listed that reuse address for about ten years.  Respected Bitcoin developers have been more [blunt](https://twitter.com/TheBlueMatt/status/1787192996854710772) on this subject.

For these reasons, I cannot personally recommend listing of Tangem wallet and I personally will not be doing a full review.   Any other  members of the community are of course always welcome to provide a review and a recommendation.

octobocto:
In addition, Tangem lets you export the xpub for your wallet, but does not search for the additional addresses and adds the corresponding balance to the wallet.

So in addition to damaging the security and privacy of their users, users might even end up losing funds.

Tangem is a dangerous wallet to use.

crwatkins:
It's been about a year, so let's close this for now.   Please resubmit when the situation changes.

PR #4344, opened 2025-01-21
Removing BitQuick from exchanges pages

timbalabuch:
https://bitquick.co/ is no longer active, I am removing the link from the exchanges page.


devdavidejesus:
Resolved by #4754 . Thanks!

PR #4410, opened 2025-05-17
Remove favicon white background

encualva:
Remove favicon.ico (16x16) white background to render more neat in dark mode browsers


Cobra-Bitcoin:
Which browser is showing a white background?

PR #4437, opened 2025-07-03
Fix Transifex links

jorgesumle:
Links were broken (404 error).


devdavidejesus:
The site is being updated. Thank you!

PR #4541, opened 2025-10-09
Add Bitcoin Knots as a node software

Retropex:


harding:
> I'm curious what [...] former contributor @harding who created a lot of bitcoin.org's promotional material for Bitcoin Core.

I feel the download and marketing pages for Bitcoin Core on Bitcoin.org are a historical artifact, due to the GitHub repo for this site being under Bitcoin Core control over a decade ago, rather than an actively maintained site feature.  New releases and new project information for Bitcoin Core has been posted to BitcoinCore.org since 2016 (initially in addition to posting to Bitcoin.org, but later most Bitcoin Core project material material was exclusively posted to BitcoinCore.org).

I think it makes the most sense for every project to have its own independent site.  Bitcoin.org can then link to those projects as its maintainers see fit, e.g. a page for full nodes can link to multiple implementations based on various criteria.  This would give btcd, libbitcoin, librerelay, and other projects an equal shot at being listed alongside Core and Knots.

If it's perceived as excessively unfair that Bitcoin.org hosts Bitcoin Core material, then I would advocate removing that material rather than encouraging every node implementation to have separate pages on the site.

crwatkins:
I agree with @harding [above](https://github.com/bitcoin-dot-org/Bitcoin.org/pull/4541#issuecomment-3403654073).  The wallet listing is a curated list that meets fairly well established criteria with very simple descriptions (a maximum of 320 characters).  It seems reasonable that the same model could apply to full nodes.

Retropex:
I agree with @harding too.

PR #4637, opened 2026-03-11
CreateQDN

Qusatotada:
Hello everyone…

If I did not say it, no one would know. What has been happening has been eroding society for the past 17 years.

What is happening today was foretold more than 2,000 years ago, in all four of the world’s great religions. Everything is part of a plan of the Creator — a process that spans many generations, unfolding over hundreds of years.

Today I stand here just like everyone else. There is nothing outwardly different about me. Yet I am the person who has received the largest inheritance of wealth in the history of humanity.

I am a logical person; I do not follow any religion. Because of that, my acceptance of this inheritance came late.

The assets I have inherited are immense. In 2008, my wealth was comparable to that of an entire nation. Unfortunately, I did not realize it at the time.

Bitcoin was created by five other individuals, but they attributed the creation of BTC to me. Because of that, it effectively became anchored to my assets. At the same time, the profits generated from my custodial investment assets have been used — in part or in full — to invest in crypto.

I state with complete certainty: my assets have been funding the entire crypto ecosystem that exists today.

Yet this is also the most condemnable part of it — because that entire system has been acting against me.

Today, this must come to an end so that things can move forward in a better way. Otherwise, I have the capacity to bring the crypto system to an end simply by no longer investing in it.

Mr. Satoshi Nakamoto.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | :white_check_mark: |
+| 5.0.x   | :x:                |
+| 4.0.x   | :white_check_mark: |
+| < 4.0   | :x:                |
+
+## Reporting a Vulnerability
+
+Use this section to tell people how to report a vulnerability.
+
+Tell them where to go, how often they can expect to get an update on a
+reported vulnerability, what to expect if the vulnerability is accepted or
+declined, etc.


Qusatotada:
Hello everyone…

Up to this moment, all the information and tens of thousands of articles about crypto, Bitcoin… all of that is the legacy of ERSTENG, or of someone similar to him. Lies have become the lifestyle and the objective of crypto.

First, I do not have 1 million BTC. If I did, it would be deposited as capital for the entire crypto system. All connections are structured to appear completely real. For example, I and five others have a structure for an online meeting of six people. The most important thing in the system I received. But to open it following the six-step instructions, my computer downloaded the most advanced types of viruses—decades ahead of their time.

I also do not know whether the origin of Bitcoin was created with good or bad intentions. But up to now, crypto has become a disaster for humanity and for me.

The idea that Satoshi Nakamoto must remain anonymous was requested from the SEC (sec.gov) so that Bitcoin would not collapse. That is a demand from a thief, with no other purpose. Everything that is public is only a stunt double. The truth is 100% different from what the media presents.

Publicly, BTC would increase tenfold from its current level if it were anchored to my assets. That is being talked about as a conspiracy theory. It is not a conspiracy theory. It is the truth.

BTC locks 1 BTC and creates hundreds of times that amount, yet the price is still maintained. So where does all that money come from?

Crypto is a pathway for transferring money from one person’s name to another—trillions of USD every year. All of this must end, because it is more immoral than any form of life on Earth.

Its existence relies on my assets. But everyone has turned against me. Because by now it is already too much and has gone too far.

Mr. Satoshi Nakamoto.

PR #4671, opened 2026-04-18
Add bearby wallet

hicaru:
Hi, i want to add Bearby wallet to the list.

Bearby is a non-custodial bitcoin wallet written in Rust. It has post-quantum key storage which i think is pretty unique compared to other wallets here. Available on Android, iOS, macOS and Linux.

What i added:
- wallet entry file with mobile and desktop platforms
- icon (144x144)
- 4 screenshots for each platform (250x350)
- translations for all locales

Some details about the wallet:
- site: https://bearby.io
- source code: https://github.com/bearbywallet/Bearby
- license: GPL-3.0
- supports legacy, segwit and bech32 addresses
- ledger hardware wallet support
- connects through rotating pool of independent electrum servers
- fully open source, user controls keys

let me know if something is missing or needs to be changed


crwatkins:
Thanks for your submission.

I installed the wallet on two iOS devices and spent just a few minutes with the app.  When trying to restore a wallet I got an error `Unsupported operation:  Cannot clear a fixed-length list`. Trying again I got `background Error: BIP39 error: Invalid word count:0`.   When I tried to restore a SeedQR code, the app read the QR code and silently did nothing.  I finally got it to take a 12 word seed, yet it didn't display any of the funds in the wallet even after leaving it for many minutes and performing many refreshes.

I later asked the wallet for an address, sent funds to that address, and asked the wallet for another address and it did not give me a new address (as required for listing).

On the Send screen, the app asked me to "Edit gas" and gave me a screen including Max Priority Fee, Gas Limit and Nonce which are not Bitcoin terminology (at least in this context). 

I will be candid here:  The errors and inability to restore funds I encountered while trying to restore a wallet, and the general UI issues and cryptic error messages I encountered while using the app for only a few minutes along with disregard for Bitcoin terminology/concepts do not reflect a wallet which is at the level required for a bitcoin.org review.   While other members of the community may be willing to proceed with such a review, I feel I should let you know that I personally will not be available for that review.

hicaru:
Thank you for the detailed review. The issues you identified are real and
not acceptable for a bitcoin.org listing. I'm closing this PR to address
them properly rather than patching in place. Specifically I will fix:

- Restore wallet flow (the Cannot clear a fixed-length list and
Invalid word count:0 errors indicate broken state handling in the
restore path)
 - SeedQR support with proper user feedback on failure
Balance fetch and address subscription for restored wallets
 - Address rotation on the receive screen — a new address per request,
never reuse
 - Separate the Bitcoin Send flow from the EVM one: no gas / nonce /
priority fee on Bitcoin, replace with fee rate in sat/vB, RBF toggle
if supported, etc.
 - Audit error surfaces so users see actionable messages instead of
raw Dart/Rust errors

I'll resubmit once these are fixed and I have a public release with at
least 3 months of user feedback confirming the fixes work.

crwatkins:
@hicaru Sounds good!   Thanks!

Be sure to carefully review the listing requirements.  Some are fairly terse, so please feel free to reach out for any help or clarification.   You can open a new issue, or actually just ask right here.

Since you seem receptive to input, I'll provide some extra comments here that are very subjective and not directly related to the listing criteria and thus I would not normally mention.  The general UX seemed very confusing compared to most other wallets.   The three icons across the bottom meant absolutely nothing to me as well as the scattered UI elements on the page other than the Send and Receive buttons (and only because they were labeled in English) and the gear icon.  Every other UI element/icon on the page was a total "whack-a-mole" and on more than half, when selected and a page popped up, I wasn't sure what I was supposed to do on it or what it was telling me.  There is sufficient obfuscation in the UI such that the few times I went looking for something I had already found once (such as changing the address type) I had absolutely no idea where to look again.  There is definitely some room for usability improvements.

hicaru:
> @hicaru Sounds good! Thanks!
> 
> Be sure to carefully review the listing requirements. Some are fairly terse, so please feel free to reach out for any help or clarification. You can open a new issue, or actually just ask right here.
> 
> Since you seem receptive to input, I'll provide some extra comments here that are very subjective and not directly related to the listing criteria and thus I would not normally mention. The general UX seemed very confusing compared to most other wallets. The three icons across the bottom meant absolutely nothing to me as well as the scattered UI elements on the page other than the Send and Receive buttons (and only because they were labeled in English) and the gear icon. Every other UI element/icon on the page was a total "whack-a-mole" and on more than half, when selected and a page popped up, I wasn't sure what I was supposed to do on it or what it was telling me. There is sufficient obfuscation in the UI such that the few times I went looking for something I had already found once (such as changing the address type) I had absolutely no idea where to look again. There is definitely some room for usability improvements.

hello, Which format or standard did you use for the QR code when trying to restore the SeedQR? For example, was it BIP39 mnemonic encoded as plain text, UR (Uniform Resource), or a custom wallet-specific format?

crwatkins:
> hello, Which format or standard did you use for the QR code when trying to restore the SeedQR? For example, was it BIP39 mnemonic encoded as plain text, UR (Uniform Resource), or a custom wallet-specific format?

I would like to be clear that no particular format is required by the listing criteria, however I believe [SeedQR format](https://satochip.io/seedqr/) as supported by Sparrow wallet was used.

hicaru:
> > hello, Which format or standard did you use for the QR code when trying to restore the SeedQR? For example, was it BIP39 mnemonic encoded as plain text, UR (Uniform Resource), or a custom wallet-specific format?
> 
> I would like to be clear that no particular format is required by the listing criteria, however I believe [SeedQR format](https://satochip.io/seedqr/) as supported by Sparrow wallet was used.



> > hello, Which format or standard did you use for the QR code when trying to restore the SeedQR? For example, was it BIP39 mnemonic encoded as plain text, UR (Uniform Resource), or a custom wallet-specific format?
> 
> I would like to be clear that no particular format is required by the listing criteria, however I believe [SeedQR format](https://satochip.io/seedqr/) as supported by Sparrow wallet was used.

Hello gain! I here studyed and did some work on Address Roation, becouse my wallet use quantum-resistive encryption NTRU Prime, so in total it's full Post Quntum since I don't reviel public keys, every time addresses change with default work of TapRoot. So I wonder will they let me pass if I directly force users to use TapRoot addresses? Of couse we pot support the others and you can do transactions from them, but as soon as you do that then the whole tree of old addresses (by all standarts) we send on change to TapRoot address.

crwatkins:
@hicaru I don't understand any of what you wrote.  How does NTRU Prime apply to Bitcoin?   I don't see any way in which that would affect address rotation.

hicaru:
> @hicaru I don't understand any of what you wrote. How does NTRU Prime apply to Bitcoin? I don't see any way in which that would affect address rotation.

I mean I encrype (private keys, secret words) with quantum resistive encryption NTRU Prime, and with Address rotation integraton it becomes full quantum secured, while into bitcoin itself you dont need to integrate quantum resistive signitures.

crwatkins:
> I mean I encrype (private keys, secret words) with quantum resistive encryption NTRU Prime, 

This implies that the seed material on the device is encrypted with NTRU Prime.  That's interesting, but post-quantum  crypto doesn't buy you much over using AES-256 since the quadratic speedup on symmetric algorithms such as AES can be mitigated against Grover by simply doubling the key size.

In the process of trying to figure out what the point is here, I looked at the bearby.io site which states:

> Unlike wallets relying on ECDSA — which a quantum computer running Shor's algorithm could break — Bearby's architecture is designed to protect assets against threats that don't yet exist.

I don't know of any wallets that use ECDSA for storing seed material.  I would like to know of some examples.

Back to your message above...

> and with Address rotation integraton it becomes full quantum secured, 

I have no idea what this has to do with address rotation.

> while into bitcoin itself you dont need to integrate quantum resistive signitures.

if you are saying that you don't need to change any Bitcoin signature algorithms to make the transactions quantum resistant, I think you are going to be famous.

In addition, on bearby.io:

> Native Bitcoin Support to store and trade your bitcoin securely. Designed to protect assets from Quantum risks.

This implies transactions are quantum secure.   You may not want to imply that.

hicaru:
> > I mean I encrype (private keys, secret words) with quantum resistive encryption NTRU Prime,
> 
> This implies that the seed material on the device is encrypted with NTRU Prime. That's interesting, but post-quantum crypto doesn't buy you much over using AES-256 since the quadratic speedup on symmetric algorithms such as AES can be mitigated against Grover by simply doubling the key size.
> 
> In the process of trying to figure out what the point is here, I looked at the bearby.io site which states:
> 
> > Unlike wallets relying on ECDSA — which a quantum computer running Shor's algorithm could break — Bearby's architecture is designed to protect assets against threats that don't yet exist.
> 
> I don't know of any wallets that use ECDSA for storing seed material. I would like to know of some examples.
> 
> Back to your message above...
> 
> > and with Address rotation integraton it becomes full quantum secured,
> 
> I have no idea what this has to do with address rotation.
> 
> > while into bitcoin itself you dont need to integrate quantum resistive signitures.
> 
> if you are saying that you don't need to change any Bitcoin signature algorithms to make the transactions quantum resistant, I think you are going to be famous.
> 
> In addition, on bearby.io:
> 
> > Native Bitcoin Support to store and trade your bitcoin securely. Designed to protect assets from Quantum risks.
> 
> This implies transactions are quantum secure. You may not want to imply that.
Hello Thank you for helping me!

What I’m trying to explain is that if a blockchain uses elliptic curves, wallets sign transactions with a public key. Once a transaction is broadcast to the blockchain, that public key is revealed  and that’s exactly where the quantum vulnerability lies.

However, if I use the Taproot address standard (BIP86), which does not by itself expose the public key for extraction, then with address rotation as I implement it — for example:

Sender balance(bip44, 49, 84, 86): 1 BTC → sending 0.1 BTC → recipient → change (only bip86): 0.9 BTC

* in this case, the change address is new, and we have not signed anything with it. Therefore, those funds remain safe, even though we are signing the transaction itself using elliptic curves.

* The key point is that I forcibly enforce the Taproot standard (while still providing full support for all other address types, including legacy). After the first transaction, the change simply goes to a Taproot change address.

My question: does this violate any rule or best practice? I haven’t found anything to suggest it does except that legacy addresses must be supported (which they are, though the user never sees them).But the wallet itself will still find the balance and keep it safe.

crwatkins:
If you are asking "Are there any listing criteria which specify which signature types to be used for certain transactions?" the answer to that is "no."

However, unless I'm still misunderstanding you, I believe you are confused as taproot is considered less quantum secure than segwit, not more.    Perhaps you need some expert advice in these areas and this not the place for such advice.   I would suggest using more appropriate forums such as a Bitcoin Stack Exchange or employing a professional expert.

hicaru:
> If you are asking "Are there any listing criteria which specify which signature types to be used for certain transactions?" the answer to that is "no."
> 
> However, unless I'm still misunderstanding you, I believe you are confused as taproot is considered less quantum secure than segwit, not more. Perhaps you need some expert advice in these areas and this not the place for such advice. I would suggest using more appropriate forums such as a Bitcoin Stack Exchange or employing a professional expert.

Thank you for your reply that’s exactly what I’m getting at. Since I’m positioning the wallet as quantum-resistant in terms of key storage, I need to understand whether I can somehow avoid forcing users to directly adopt the Taproot standard, and instead use SegWit by default.

If no such specifications exist, then thank you for your answer I’ll prepare new PR and describe everything we’ve discussed.

Please excuse my imperfect English in advance. 🙏

PR #4724, opened 2026-05-12
Move optional wallet criteria to required

crwatkins:
There have been a number of optional criteria that have been optional for about a decade now and have been generally well received with developers updating any non-compliant wallets.  I recommend making them required.


devdavidejesus:
LGTM

PR #4734, opened 2026-05-20
Add Liana wallet

kloaec:
Adding Liana wallet here.

3 additions (logo, screenshot, liana.md) and one modification (adding the description in translations/en).

Liana has a one click Core pruned node download+setup. It can also connect to Core locally, to a remote Electrum, or to a trusted node (Liana Connect). **By default, the user runs their own full node.**

Liana is a desktop wallet. (Windows, Mac, Linux)

Liana interacts with Signing Devices aka hardware wallets, as long as they support Miniscript.

Liana does NOT support Tor directly.

Everything else is a pass, including bootstrappable (and therefore reproducible) builds and HSTS.

Let me know if I need to follow a specific formatting here for this PR.


crwatkins:
Thanks for your submission!  Just wanted to let you know reviews are a bit backlogged at the moment.

PR #4747, opened 2026-05-26
Wallets: Drop orphan mixing_shuffling feature from Ginger

devdavidejesus:
The `mixing_shuffling` feature was fully removed from the site in #3280 (Feb 2020), including the `wizard-feature-mixing` and `wizard-feature-mixing-description` translation keys.

The Ginger wallet file still carries the tag in its `features:` list. With no translation string to render, this produces an empty checkmark with a blank tooltip on the wallet page, between Hardware Wallet and SegWit.

This removes the orphan tag so the feature list renders cleanly:

`features: "2fa bech32 hardware_wallet mixing_shuffling segwit"` → `features: "2fa bech32 hardware_wallet segwit"`

### DIFF
--- a/_wallets/ginger.md
+++ b/_wallets/ginger.md
@@ -15,7 +15,7 @@ platform:
       link: "https://gingerwallet.io/"
       source: "https://github.com/GingerPrivacy/GingerWallet"
       screenshot: "ginger.png"
-      features: "2fa bech32 hardware_wallet mixing_shuffling segwit"
+      features: "2fa bech32 hardware_wallet segwit"
       check:
         control: "checkgoodcontrolfull"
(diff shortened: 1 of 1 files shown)


crwatkins:
LGTM.   That was my fault as when I reviewed it and saw that tag (which was accurate) I assumed it would be simply be ignored since it had been removed from the UI.   I was wrong; good catch.

devdavidejesus:
> LGTM. That was my fault as when I reviewed it and saw that tag (which was accurate) I assumed it would be simply be ignored since it had been removed from the UI. I was wrong; good catch.

No worries at all...I only caught it because I happened to notice it on the site.

Cobra-Bitcoin:
LGTM.

PR #4836, opened 2026-06-29
Add bearby wallet

hicaru:
### Description
Bearby is a non-custodial Bitcoin wallet written in Rust (core logic) and flutter (frontend), available on Android, iOS, macOS, linux (source build).

### Technical Features
* **Hybrid Storage Encryption:** Local software keys and secrets are encrypted using a hybrid scheme combining AES, Kuznyechik, and lattice-based NTRU Prime algorithms.
* **Hardware Wallet Support:** Native Ledger integration. Private key storage is handled entirely by the Ledger device. Bearby only receives the public key required for operation and does not persist it locally, preventing public key exposure.
* **Zero-Exposure Public Key Policy:** To minimize the cryptographic attack surface, public keys—whether derived from a software seed or queried from a hardware wallet (e.g., Ledger) are never exposed or persistently stored in an unencrypted state locally.
* **SegWit Default:** Fully supports Legacy, SegWit, and Bech32 addresses, but enforces SegWit formats as the immutable default for internal address generation.
* **Address Rotation:** Automatic address rotation is enabled by default for all transactions.
* **Network Redundancy:** Connects via a randomized pool of independent Electrum servers with automatic failover if a node drops.

### What I added:
* Wallet entry file with mobile and desktop platforms
* Icon (144x144)
* 4 screenshots for each platform (250x350)
* Translations for all locales

### Metadata
* **Website:** https://bearby.io
* **Source Code:** https://github.com/bearbywallet/Bearby
* **License:** GPL-3.0


crwatkins:
I regret that my very brief experience with the latest iOS version is very similar to my original experience detailed in #4671.  The UI/UX confusion remains high to the point of making it very difficult to actually use this wallet.  As someone that has used about one hundred different wallets, I can't imagine how hard it would be for a new user to be successful with this wallet.   In just minutes with the new version:

* There were layout problems on tablets often hiding back navigation arrows making it impossible to navigate
* There are "No transactions yet" on a wallet with dozens of transactions (even after an hour of waiting).  This is a serious bug.
* When sending there is a dialog which says "Select Address" with only the index 0 address of the wallet listed which makes no sense
* On the sending screen, there are no clues or instructions that one is on a sending screen.   The button says "Submit"
* On the sending screen there is a button to send 0% which leads to pause and confusion (I have no idea of the purpose)

This was observed after about three minutes of usage.   None of the previous serious bugs in #4671 were re-tested.

I continue to be concerned about the claims on the website.   I believe they are potentially misleading implying that the wallet can somehow protect Bitcoin transactions on the blockchain from "Quantum risks".    I'm no cryptographer, but as I [previously explained](https://github.com/bitcoin-dot-org/Bitcoin.org/pull/4671#issuecomment-4579677539) I believe the wallet's advantage over other wallets is quite exaggerated.

There is a serious amount of UI/UX work and bugs that need to be fixed before I personally would be ready to formally review this wallet.  Others may be willing to review it sooner, but I won't be revisiting it for at least twelve months.

devdavidejesus:
@hicaru

I recommend reading Craig’s feedback carefully and addressing the issues he pointed out.

Craig has reviewed many Bitcoin wallets over the years and I think his comments deserve serious consideration.

At this point, I can’t recommend listing this wallet on Bitcoin.org. I share Craig’s assessment and believe there are still significant UI/UX issues and bugs that should be resolved before the wallet is ready for inclusion.

I hope you’ll continue improving the project.

hicaru:
> I regret that my very brief experience with the latest iOS version is very similar to my original experience detailed in #4671. The UI/UX confusion remains high to the point of making it very difficult to actually use this wallet. As someone that has used about one hundred different wallets, I can't imagine how hard it would be for a new user to be successful with this wallet. In just minutes with the new version:
> 
>     * There were layout problems on tablets often hiding back navigation arrows making it impossible to navigate
> 
>     * There are "No transactions yet" on a wallet with dozens of transactions (even after an hour of waiting).  This is a serious bug.
> 
>     * When sending there is a dialog which says "Select Address" with only the index 0 address of the wallet listed which makes no sense
> 
>     * On the sending screen, there are no clues or instructions that one is on a sending screen.   The button says "Submit"
> 
>     * On the sending screen there is a button to send 0% which leads to pause and confusion (I have no idea of the purpose)
> 
> 
> This was observed after about three minutes of usage. None of the previous serious bugs in #4671 were re-tested.
> 
> I continue to be concerned about the claims on the website. I believe they are potentially misleading implying that the wallet can somehow protect Bitcoin transactions on the blockchain from "Quantum risks". I'm no cryptographer, but as I [previously explained](https://github.com/bitcoin-dot-org/Bitcoin.org/pull/4671#issuecomment-4579677539) I believe the wallet's advantage over other wallets is quite exaggerated.
> 
> There is a serious amount of UI/UX work and bugs that need to be fixed before I personally would be ready to formally review this wallet. Others may be willing to review it sooner, but I won't be revisiting it for at least twelve months.

Thank you for taking the time to review the wallet again, I appreciate the detailed feedback.
You're right about the UI/UX issues, and I apologize that they persist. The "No transactions yet" bug on a wallet with transaction history is serious and will be my top priority, along with the tablet layout problems, the confusing "Select Address" dialog, and the unclear send screen. My focus has been on the cryptography and privacy layer, so the UX clearly suffered as a result. That's on me to fix.
Regarding the quantum claims, your concern is fair and I want to be precise about what we do and don't provide. The post-quantum protection applies to local key storage only: private keys at rest are encrypted with Streamlined NTRU Prime (sntrup761, a lattice-based scheme) combined with a symmetric cipher cascade. On-chain, transactions still use standard ECDSA like every other Bitcoin wallet. We cannot and do not protect signatures on the blockchain from a future quantum adversary. Additionally, the wallet avoids address reuse and never exposes public keys unnecessarily, so unspent funds stay behind hash160, but that's a hardening measure, not quantum immunity. I will revise the wording on the website to make this distinction explicit and remove anything that implies on-chain quantum protection.
I'll work through the reported issues and won't ask for a re-review until the wallet is genuinely ready. Thanks again for your time.

PR #4746, opened 2026-05-25
Add missing Bitcoin Core release notes (28.2 through 31.0)

devdavidejesus:
This PR adds the 11 Bitcoin Core release notes that are currently missing from `_releases/`. The most recent release present in the directory is 28.1; this brings the directory up to date through 31.0.

Relates to #4704.

### Releases added

| File | Version | Published |
|------|---------|-----------|
| `_releases/v28.2.md` | 28.2 | 2025-06-26 |
| `_releases/v28.3.md` | 28.3 | 2025-10-17 |
| `_releases/v28.4.md` | 28.4 | 2026-03-18 |
| `_releases/v29.0.md` | 29.0 | 2025-04-14 |
| `_releases/v29.1.md` | 29.1 | 2025-09-04 |
| `_releases/v29.2.md` | 29.2 | 2025-10-14 |
| `_releases/v29.3.md` | 29.3 | 2026-02-10 |
| `_releases/v30.0.md` | 30.0 | 2025-10-10 |
| `_releases/v30.1.md` | 30.1 | 2026-01-02 |
| `_releases/v30.2.md` | 30.2 | 2026-01-10 |
| `_releases/v31.0.md` | 31.0 | 2026-04-19 |

### Method

- Each file follows the existing format of `_releases/v28.1.md` (YAML header, `{% githubify %}` wrapper, `XX Release Notes` heading).
- Release-note text is copied verbatim from the official notes in `bitcoin/bitcoin` (`doc/release-notes/release-notes-XX.md`), with only the Markdown formatting adjustments described in `docs/adding-release-notes-and-alerts.md` (blank line before lists).
- The `date:` field of each file is the official announcement date, taken from the "was published on" line of the corresponding page at `bitcoincore.org/en/releases/XX/`.
- `optional_magnetlink` is left empty in each file, consistent with it being an optional field.

### Two things worth flagging 

1. **30.0 and 30.1 binaries.** The Bitcoin Core project pulled the 30.0 and 30.1 binaries from distribution due to the wallet-migration advisory (https://bitcoincore.org/en/2026/01/05/wallet-migration-bug/). If CI checks the download links in those two release notes, they may fail. I've still included both files for completeness of the historical record.

[...]

Issue #4704, opened 2026-05-02
Bitcoin Core page is 23 releases behind, including unreflected wallet-loss security advisory


devdavidejesus:
## Summary

As of 2026-05-01, the `_releases/` directory of this repository contains only **7 modern Bitcoin Core releases** (22.0+), while the Bitcoin Core project has tagged **30 releases** in that range on https://github.com/bitcoin/bitcoin. **23 releases are absent**.

Of those 23 absent releases:

- **4 are currently maintained** under the project's three-major-version policy (29.0, 29.1, 29.2, 29.3) and have been published between 79 and 381 days ago.
- **3 are pulled or critical** — 30.0 and 30.1 were pulled from official distribution due to a critical wallet-loss bug; 30.2 contains the fix.
- **1 is the current latest release** (31.0, published 11 days ago).
- **15 are EOL** (the entire 22.1, 23.x, 24.x, 25.1+, 26.x, and recently 28.x series) — most never added at any point in this repository's history.

The list of absent releases also includes versions covered by a **still-unreflected security advisory** (30.0 and 30.1, advisory published 2026-01-05, 119 days ago).

The same gap is visible at https://bitcoin.org/en/bitcoin-core/ (which lists 28.1 as the latest in its News section — note: 28.x became EOL on 2026-04-20 with the 31.0 release) and https://bitcoin.org/en/version-history (which jumps from 22.0 → 25.0 → 27.0).

This is not a recent slip — issue #3926 (September 2022) reported the same pattern when the page was stuck on 22.0 vs 23.0; that issue was closed without resolution.

**Part A** documents the gap, with all dates verified against signed Git tags in `bitcoin/bitcoin`.
**Part B** proposes structural options to prevent reoccurrence.

---

### Method

Verified on 2026-05-01 by:
1. Enumerating the file listing of `_releases/` in this repository via the GitHub web interface.
2. Cross-referencing each absent version against its signed tag at `https://github.com/bitcoin/bitcoin/releases/tag/v<version>`.

All release dates below are pulled from the GitHub release page for each tag (UTC). A small Python script using only the standard library is included in the appendix for reproducibility.

### Releases present in `_releases/` (modern era, 22.0+)

`v22.0.md`, `v25.0.md`, `v27.0.md`, `v27.1.md`, `v27.2.md`, `v28.0.md`, `v28.1.md` — total: **7**.

### Releases absent from `_releases/` (modern era, 22.0+)

[...]

PR #4863, opened 2026-07-10
Add Bitcoin Core 30.3 and 31.1, and fix the release date field

devdavidejesus:
This PR adds the two Bitcoin Core releases published on 2026-07-08, and fixes the date field used by the 19 most recent release notes.

### Part 1 — Add 30.3 and 31.1

| File | Version | Published |
|------|---------|-----------|
| `_releases/v30.3.md` | 30.3 | 2026-07-08 |
| `_releases/v31.1.md` | 31.1 | 2026-07-08 |

The release-note text is copied from `bitcoin/bitcoin` (`doc/release-notes/release-notes-XX.md`). The only change to that text is the insertion of blank lines before Markdown lists, per step 1 of the reformatting instructions in `docs/adding-release-notes-and-alerts.md`. Step 2 (URLs in angle brackets) required no change — the upstream files already follow it.

Both upstream files already begin with a `vXX.X Release Notes` heading, so none was added.

Dates are taken from the announcement URLs at `bitcoincore.org/en/2026/07/08/release-30.3/` and `bitcoincore.org/en/2026/07/08/release-31.1/`.

### Part 2 — Fix the date field

Every template that renders release information reads `optional_date`:

- `en/version-history.html` — sorts by it, and displays it
- `en/rss/releases.rss` — sorts by it, and uses it for `<pubDate>`
- `_layouts/release.html` — displays it under the page title

Since `v0.21.1` (May 2021), new release notes have used `date` instead. Nothing in the site reads that field, so those 19 files are invisible to the sort and to every date display.

Observable effects:

- **`/en/version-history` is out of order.** Files without `optional_date` are pushed to the end of the sort by the `'last'` argument, then to the top by `reversed`. Among themselves they appear in filesystem read order — the live page currently shows 31.0, 28.4, 29.3, 30.2, 30.1, 28.3, 29.2, 30.0, …
- **No date is displayed** for those 19 entries, since the template guards on `{% if p.optional_date %}`.
- **The RSS feed carries the wrong `pubDate`.** The `{% else %}` branch falls back to `site.time`, so all 19 entries currently show the last build date (`Thu, 25 Jun 2026`) instead of their release date. The template also emits `{% warn "This release doesn't have optional_date set" %}` for each of them on every build.

This PR renames `date` to `optional_date` in those 19 files. The values are unchanged.

### After this PR

`/en/version-history` becomes a single continuous chronological list, from 31.1 (2026-07-08) back to 0.3.21 (2010), with a date on every entry. The RSS feed carries real publication dates. The 19 build warnings stop.

Ordering remains chronological rather than by version number — 28.4 (March 2026) appears above 30.0 (October 2025), since Bitcoin Core maintains several branches in parallel. This is what the template intends; the assigned variable is named `date_sorted_releases`.

### Note — the 31.1 binaries need to go up before deploy

Once merged, `_plugins/releases.rb` sets `site.DOWNLOAD_VERSION` to 31.1 (the highest `required_version` across `_releases/`). `_templates/download.html` builds every download link from that value:

```liquid
{% capture PATH_PREFIX %}/bin/bitcoin-core-{{ site.DOWNLOAD_VERSION }}{% endcapture %}
```

`https://bitcoin.org/bin/bitcoin-core-31.1/` currently returns 404, so all ten `class="dl"` links on the download page would break once this is deployed.

CI does not catch this: `check-for-broken-bitcoin-core-download-links` exists in the Makefile but is not reached from the `travis` target — neither `pre-build-tests-fast` nor `post-build-tests-fast` invokes it. That is why #4746 built green while ten of its eleven releases had no binaries on the server (28.2 through 30.2 all still return 404 today).

Same situation as #4746, where the 31.0 binaries went up before the merge. I don't have server access, so this needs the binaries in place first.

### Verification

- All 84 files in `_releases/` now use `optional_date`; none use `date`.
- The 19 renamed files show exactly one changed line each: 19 deletions of `date: …` and 19 insertions of `optional_date: …`, with identical values. No other line was touched.
- The two new files were compared line by line against their upstream source; the content is identical apart from the blank lines inserted before lists.
- Neither new file contains the `][` sequence that `check-for-broken-markdown-reference-links` flags, and both carry the copyright header required by `check-for-missing-copyright-licenses`.

Relates to #4704.

cc @Cobra-Bitcoin

### DIFF
--- a/_releases/v0.21.1.md
+++ b/_releases/v0.21.1.md
@@ -11,7 +11,7 @@ id: en-release-0.21.1
 name: release-0.21.1
 permalink: /en/releases/0.21.1/
 excerpt: Bitcoin Core version 0.21.1 is now available
-date: 2021-05-01
+optional_date: 2021-05-01
 
 ## Use a YAML array for the version number to allow other parts of the
 ## site to correctly sort in "natural sort of version numbers"
--- a/_releases/v22.0.md
+++ b/_releases/v22.0.md
@@ -10,7 +10,7 @@ id: en-release-22.0
 name: release-22.0
 permalink: /en/releases/22.0/
 excerpt: Bitcoin Core version 22.0 is now available
-date: 2021-09-13
+optional_date: 2021-09-13
 
 ## Use a YAML array for the version number to allow other parts of the
 ## site to correctly sort in "natural sort of version numbers".
--- a/_releases/v25.0.md
+++ b/_releases/v25.0.md
@@ -10,7 +10,7 @@ id: en-release-25.0
 name: release-25.0
 permalink: /en/releases/25.0/
 excerpt: Bitcoin Core version 25.0 is now available
-date: 2023-05-26
+optional_date: 2023-05-26
 
 ## Use a YAML array for the version number to allow other parts of the
 ## site to correctly sort in "natural sort of version numbers".
--- a/_releases/v27.0.md
+++ b/_releases/v27.0.md
@@ -10,7 +10,7 @@ id: en-release-27.0
 name: release-27.0
 permalink: /en/releases/27.0/
 excerpt: Bitcoin Core version 27.0 is now available
-date: 2024-04-02
+optional_date: 2024-04-02
 
 ## Use a YAML array for the version number to allow other parts of the
 ## site to correctly sort in "natural sort of version numbers".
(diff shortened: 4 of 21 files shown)


devdavidejesus:
Heads-up on the RSS impact: after this is deployed, the 19 affected entries will change their `<pubDate>` from the last build date to their actual release date. Feed readers may surface them as updated. This is the intended behaviour, the current dates are wrong, but flagging it since it will be visible on the next deploy.

PR #4877, opened 2026-07-16
Update Bull Bitcoin regions

BullishNode:
Add Bull Bitcoin availability for  Europe, Mexico, Argentina, Costa Rica and Colombia. Also add a section for Costa Rica. In the future, bitcoin.org maintainers may want a dedicated section for Central America, but for now just Costa Rica is fine.


devdavidejesus:
The Central America & Caribbean section is now on master (#4884). Could you rebase and move the Costa Rica block into it? The other four look good where they are.

Issue #4673, opened 2026-04-19
Update Passport listing: website URL outdated after Foundation Devices rebrand and product renaming


devdavidejesus:
## Summary

The current `_wallets/passport.md` entry lists the Passport hardware wallet 
with a website URL pointing to `foundationdevices.com/passport/`, which now 
redirects twice:

1. Domain rebrand: `foundationdevices.com` → `foundation.xyz`
2. Product path renaming: `/passport/` → `/passport-core/`

A code-based verification confirms that Foundation Devices has rebranded to 
"Foundation" and renamed the product listed on bitcoin.org from "Passport" 
to "Passport Core". The company also released a separate new product, 
"Passport Prime", which is a distinct device.

The existing link still resolves via HTTP 301 chain, so this is not a 
broken-link issue, but the listing no longer points directly to canonical 
locations and the product name shown to users may no longer match how the 
vendor refers to it.

This Issue is opened to bring visibility so the update can be addressed 
collaboratively.

## Evidence

All evidence below is verifiable via public sources:

| # | Finding | Source |
|---|---------|--------|
| 1 | `foundationdevices.com` returns HTTP 301 → `foundation.xyz` | Direct HTTP request (`curl -sIL`) |
| 2 | `foundationdevices.com/passport/` returns HTTP 301 → `foundation.xyz/passport-core/` (double redirect) | Direct HTTP request (`curl -sIL`) |
| 3 | Site canonical URL: `<link rel="canonical" href="https://foundation.xyz/" />` | https://foundation.xyz/ HTML source |
| 4 | Open Graph URL: `<meta property="og:url" content="https://foundation.xyz/" />` | https://foundation.xyz/ HTML source |
| 5 | schema.org Organization name: `"name": "Foundation"` | https://foundation.xyz/ JSON-LD |
| 6 | schema.org description: `"Building Bitcoin-centric freedom tech"` | https://foundation.xyz/ JSON-LD |
| 7 | GitHub organization blog field: `https://foundation.xyz` | https://api.github.com/orgs/Foundation-Devices |
| 8 | GitHub organization email: `hello@foundation.xyz` | https://api.github.com/orgs/Foundation-Devices |
| 9 | Product page title tag: `<title>Passport Core - Foundation</title>` | https://foundation.xyz/passport-core/ HTML source |
| 10 | Passport Prime (separate new product) has its own dedicated page at `foundation.xyz` (homepage) | https://foundation.xyz/ HTML source |
[...]

PR #4891, opened 2026-07-20
Update Passport Core listing after rebrand and product rename

BitcoinQnA:
Closes part of #4673.

I'm on the Foundation team. Following up on the discussion in #4673 with the vendor-side context.

### What changed since the issue was opened

The issue proposed pointing `link` at `https://foundation.xyz/passport-core/`. That URL is no longer valid. As of now:

```
foundationdevices.com/passport/       301 -> foundation.xyz/products/passport-prime
foundation.xyz/passport-core/         301 -> foundation.xyz/products/passport-prime
foundation.xyz/passport               301 -> foundation.xyz/products/passport-prime
foundation.xyz/products/passport-core 404
```

There is no Passport Core page on foundation.xyz. The only product page is Passport Prime, confirmed by our sitemap. This means the current listing sends users to a **different device** than the one described: the listing says "air-gapped," and Passport Prime is not air-gapped.

### This PR

- `title` / `titleshort`: `Passport` to `Passport Core`, matching our current naming and disambiguating from Passport Prime
- `link`: now `https://docs.foundation.xyz/passport/home/`, the Passport Core documentation. Since we have no Core product page, this is the canonical Core-specific destination we can offer, and it will not redirect to a different device
- Description: name updated, "is manufactured" to "was manufactured", and a note that Core is no longer in production but still receives firmware updates and support
- `id` left as `passport` to avoid breaking existing anchors and inbound links
- `source` unchanged, `Foundation-Devices/passport2` is still correct and actively maintained

### On Passport Prime

Prime is a separate device and would need its own listing assessed against the hardware wallet requirements. I'll raise that separately rather than bundle it here.

Thanks @devdavidejesus for the audit and for flagging this.


devdavidejesus:
cc @crwatkins

devdavidejesus:
LGTM

crwatkins:
LGTM.   Thanks for the update.

PR #4896, opened 2026-07-23
Modernize build toolchain: Ruby 2.5.8 → 3.3.10, clear Dependabot alerts

devdavidejesus:
> **Draft — do not merge until the build server has Ruby 3.3.10 (`rvm install 3.3.10`), per the Gemfile's own notes.**

## Why

The 40 open Dependabot alerts all live in the Gemfile.lock, which has been frozen on the Ruby 2.5.8 (EOL) toolchain; the bulk of them (the nokogiri chain, pinned by html-proofer 2.1.0) can't reach secure versions without a newer Ruby. These are build/test dependencies on a static Jekyll site, nothing is served to visitors, so real-world risk is low, but the alerts are valid. 

## What

- **Ruby 2.5.8 → 3.3.10** in Gemfile, `.ruby-version`, `.travis.yml`. Travis also drops the 2.5.8-era RubyGems/bundler pins, which would break on Ruby 3.3. (3.3.10 is the patch Travis images ship prebuilt — no rvm binaries exist for 3.3.12; the pin can move to any 3.3.x the server gets.)
- **html-proofer 2.1.0 → ~> 5.0**, pulls fully patched nokogiri 1.19.4; rexml, concurrent-ruby, addressable and RedCloth all resolve to secure versions, clearing the alerts.
- **therubyracer/libv8 (2014) → mini_racer**, needed as the ExecJS runtime for `check-jshint` (verified: `ExecJS.runtime.name` → "mini_racer (V8)" on Linux). **`less` removed**, nothing in the build invokes it (no plugin, no Makefile target; CSS comes from the SCSS pipeline, `ie8.less` is copied as-is).
- **Plugin fixes for APIs removed in Ruby 3.x** (~14 lines across 7 files): Psych 4 rejects YAML aliases (8× `YAML.unsafe_load_file`); `File.exists?` removed in 3.2 (5× `File.exist?`); `Kernel#open` with URLs removed in 3.0 (2× `URI.open`, the old call failed silently behind a bare rescue, caching an empty contributors list).
- **`_contrib/bco-htmlproof` ported to the html-proofer 5.x API**, keeping the exact 2.x checking scope (internal links only, no https enforcement, no hash checking).
- **One config line:** `sass: silence_deprecations: ["import"]`, Dart Sass deprecation warnings were tripping `check-for-build-errors`; migrating the SCSS to `@use` remains a future cleanup (sass-migrator automates it).
- Lock includes `x86_64-linux` and `arm64-darwin` platforms; `docs/setting-up-your-environment.md` updated to 3.3.10.

## Proof

Full `make build test` passes end to end in a Linux container (`ruby:3.3` Docker image, same GNU toolchain as the build server): 3,823 pages built, every check green, link checker included (5,234 internal links, zero failures). The full build also passed on macOS arm64 during the sandbox PoC; the test suite on macOS only trips on pre-existing BSD-vs-GNU tool assumptions in the Makefile, unrelated to this change.

## Notes for review

- The 3.3.10 pin follows the current exact-pin convention and matches what Travis ships.
- Backlog findings, deliberately not in this PR: the footer's MIT license link is `http://` on every page; two dead anchors in old posts (`#sponsorship`/`#github` on about-us); `_alerts` destination-conflict warnings in the build log; the Makefile assumes GNU tools.


Cobra-Bitcoin:
Thank you! This is great work and long overdue. Everything looks good. I think with a change like this the most important thing is knowing there isn't something silently breaking somewhere, and I don't think there is. I will merge this myself once the build server is prepared and the conflict has been resolved (good that we no longer need the events plugin). Will report back if it's a success.

devdavidejesus:
Thanks, @Cobra-Bitcoin !

PR #4912, opened 2026-08-04
Spend Bitcoin: name an online-store directory on the Shop online card

GalaxyMind21:
## Rewritten after #4942

I opened this on 4 August proposing a second card for online-store directories. #4942 rebuilt the page two days ago around durable ways to spend rather than a list of external directories, and that answered the structural question I was asking. The old patch no longer applies and I have replaced it rather than leave a conflicting diff sitting here.

What is left is one sentence.

Every card on the rebuilt page names one concrete example: BTC Map and Steak 'n Shake for local businesses, Bitrefill for gift cards, the Hal Finney fund, OpenSats and HRF for donations. "Shop online" is the only card that names none, so it tells a reader what to look for at checkout but not where to start looking. This adds one example to that card, in the same shape as the others. No new card, no new icon, no change to the page structure.

If you would rather that card name a different directory, or name none at all, close this and I will not push back. The gap is what I wanted to flag; I am not attached to being the answer to it.

## Disclosure, including a correction

I run Galaxy Mind, so treat this as a self-submission and weigh it accordingly.

**Correcting my own PR description.** The original said "no affiliate links." That was true when I wrote it on 4 August. On 8 August I added a MiningRigRentals referral link to a hashrate rental page on the site, which made that sentence false while this PR was open, and I did not come back and fix it. I removed the referral on 13 August, so the claim is true again, but it was wrong for five days and you should know that from me rather than find it.

The site also runs a small ASIC rental service of its own. It is a real service, it is disclosed, and it is not part of the directory. I have taken its promotion off the directory pages so the two are not mixed.

**On the questions from your DM.** Listings are re-verified on a rolling 30-day cycle, and every card shows the UTC date a human last confirmed it. I told you 30 days in the DM while the site's own policy still said 90, which was me describing intent rather than the system. I have since walked the whole roster by hand and reset the dates, and tightened the site's threshold to 30 so the two agree. You can check any claim here against `/api/vendors.json`, which exposes `verifiedAt` per listing.

Nothing is auto-added. A person confirms on-chain or Lightning payment for a physical good before anything lists, and new listings need an X or Nostr identity, because a Lightning address on its own proves a wallet resolves and not that an accountable merchant stands behind it. No casinos, no gambling, no altcoins, and no paid placement of any kind. Methodology is public at https://galaxymind.space/methodology.


devdavidejesus:
Hi @GalaxyMind21 ,

I reviewed it and liked what I saw. 

As long as the vendors genuinely accept Bitcoin, don’t promote illicit activities or gambling, and the directory remains curated and up to date, I’ll be happy to include Galaxy Mind on bitcoin.org in the coming days.

Initiatives like this help strengthen the ecosystem. Good job!

GalaxyMind21:
Thank you @devdavidejesus

_translations/en.yml
```diff
@@ -1097,7 +1097,7 @@ en:
...
     spend-online: "Shop online"
-    spend-online-text: "A growing number of online stores accept Bitcoin at checkout, often through a payment processor. Look for a Bitcoin or Lightning payment option when paying: you scan a QR code or open a payment link with your wallet, and the store gets paid."
+    spend-online-text: "A growing number of online stores accept Bitcoin at checkout, often through a payment processor. Look for a Bitcoin or Lightning payment option when paying: you scan a QR code or open a payment link with your wallet, and the store gets paid. Independent stores that take Bitcoin directly are harder to find than large retailers; <a href=\"https://galaxymind.space/\">Galaxy Mind</a> keeps a human-verified directory of them, showing the date each listing was last confirmed."
```
devdavidejesus: I think we can make this wording a little cleaner and more neutral. Could you please update this line to:

“A growing number of online stores accept Bitcoin at checkout, often through a payment processor. Look for a Bitcoin or Lightning payment option when paying: you scan a QR code or open a payment link with your wallet, and the store gets paid. For independent stores that accept Bitcoin directly, [Galaxy Mind](https://galaxymind.space/) maintains a directory of manually verified stores and shows when each listing was last confirmed.”

I think this reads more naturally and fits better with the tone of the page.

GalaxyMind21: Done, thank you - applied exactly as written in e6e3d11.

One small thing I changed and want to flag rather than slip past you: your suggestion used a markdown link, and en.yml uses HTML anchors throughout (there are no markdown links anywhere in the file), so I wrote it as an <a href>. The wording is otherwise word for word yours. Happy to adjust if you'd rather it read differently.

On the three conditions you set: no gambling or illicit vendors, and the directory is manually vetted with an X or Nostr identity required before anything lists. There are no affiliate links on the website. Continuous monitoring is built too, so a listing that goes dark or stops showing a Bitcoin payment option gets flagged for review rather than waiting on the next manual sweep.

Issue #4801, opened 2026-06-19
Proposal: retire the Events and Alerts infrastructure, modernize the Release Notes guide for 2026


devdavidejesus:
@Cobra-Bitcoin —> opening this as a discussion before any PR, since the Events and Alerts items below are policy calls that are yours to make.

### Summary

I'd like to propose retiring the **Events** and **Alerts** infrastructure (both long inactive) and **modernizing** the **Release Notes** section, which is active and was updated recently (#4746). The guide documenting all three, `docs/adding-release-notes-and-alerts.md`, last touched in 2018, would be trimmed to a focused, current Release Notes guide and renamed accordingly.

My recommendation is action on each item, but they're independent, each can be accepted or rejected on its own. Every factual claim here was verified directly against `upstream/master` (`git show` / `git log` / `git grep`), not the rendered site; a verification table is appended.

### Verified current state

- **The guide** was last modified **2018-04-06**. It still references Freenode (defunct 2021) and the `0.10.0.md` naming scheme replaced in 2021.
- **Events** (`_events.yml`): last content commit **2022-09-28**; now holds a single long-past event (Satsconf, 2022-11-08). The live `/en/events` page shows nothing upcoming.
- **Alerts** (`_alerts/`): most recent alert `2018-09-21-required-upgrade.md`; no commits to the directory since (the last, 2019, was a test fix). The live page reads "There is no ongoing event on the Bitcoin network."
- **Release Notes** (`_releases/`): active, current through `v31.0.md`, recently backfilled in #4746. Maintained in catch-up batches by several contributors (recent commits from laanwj in 2025-03, lateminer, Cøbra, and myself). Bodies are copies of the upstream Bitcoin Core notes, as the files' own headers state.

### Decisions

**1. Events — recommend removal.**

A single 2022 event, no curation in nearly four years, behind a "Submit an event" flow nobody uses. No Makefile/CI target depends on `_events.yml`, so removal does not affect the build (verified). If you approve, removal would cover:

  - `_events.yml` (data)
  - `_plugins/events.rb` (the plugin that loads it)
  - `_templates/events.html`
  - `en/rss/events.rss` (feed)
  - the `/events` links in `_includes/layout/base/head-menu.html` and
    `footer-menu.html`

  Historical mentions of past events in old `_posts/` would be left untouched, they're accurate records of their time.

**2. Alerts — recommend removing the publishing workflow + site machinery; the historical archive is a separate call.**

[...]

PR #4921, opened 2026-08-06
Remove the Events page and its infrastructure

devdavidejesus:
Refs #4801

Removes the Events page and its supporting infrastructure. This takes effect across all 31 languages at once, since the page and menu entries are generated from shared templates/includes.

Removed:
- `_events.yml` (single long-past event, Satsconf 2022-11-08)
- `_plugins/events.rb` (also drops a build-time Google geodata dependency)
- `_templates/events.html`
- `en/rss/events.rss`

Edited (events link removed):
- `_includes/layout/base/head-menu.html`
- `_includes/layout/base/footer-menu.html`
- `_templates/community.html` (meetups card keeps its other links)

Deliberately untouched:
- _translations/*.yml — left untouched on purpose: one key in the events block is still consumed elsewhere (rsslink, used by version-history.html), and editing 31 YAML files by hand is an avoidable deployment risk
- Historical mentions of past events in `_posts/`
- Everything related to Alerts, that part of #4801 remains pending

Verified with a full `make deployment` build (Ruby 2.5.8) before pushing: site generates completely, no events pages emitted in any language, menus render without the events link, all post-build checks pass.


devdavidejesus:
Travis caught two historical posts linking the removed page; unlinked those, plus one absolute-URL link the proofer doesn't check, text preserved, only the anchors removed.

Issue #4911, opened 2026-08-04
Site deployment build is broken


Cobra-Bitcoin:
This commit https://github.com/bitcoin-dot-org/Bitcoin.org/commit/e771fd2c21383a757daaa87edd57a50a9575c9fa seems to have broken the site somehow and subsequent commits don't go live because the site won't build.

devdavidejesus:
The anchor check is the symptom: `find: '_site/img': No such file` means the Jekyll build wrote almost nothing into `_site`, and the checker just trips over the missing files first, the real error is further up the log. e771fd2 only brought #4891's small diff into master (the rest of that merge bubble was already there and had passed CI), and I reproduced a full green `make deployment` today from a fresh clone of current master (d4286bfb) on Ruby 2.5.8, `check-for-missing-anchors` included, so the tree builds, and the difference is on the server. This one-shot either completes or names the cause in the last lines:

    df -h . ; free -m ; ruby -v
    dmesg 2>/dev/null | grep -iE 'killed|oom' | tail -5
    git fetch origin && git reset --hard origin/master
    rm -rf _site .jekyll-cache
    make deployment 2>&1 | tee /tmp/build.log | tail -60

If it goes green, the publish step can be run; if not, paste the tail of `/tmp/build.log` and I'll send the fix.

devdavidejesus:
Confirmed experimentally: running the post-build checks against an empty `_site` (skipping the Jekyll build entirely) on a healthy machine reproduces your error output exactly, same five files, same ten missing anchors, same `Makefile:135` failure:

    find: '_site/img': No such file or directory
    grep: _site/en/bitcoin-core/features/privacy.html: No such file or directory
    grep: _site/en/bitcoin-core/features/privacy.html: No such file or directory
    grep: _site/en/bitcoin-core/features/user-interface.html: No such file or directory
    grep: _site/en/bitcoin-core/features/validation.html: No such file or directory
    grep: _site/en/bitcoin-core/features/validation.html: No such file or directory
    grep: _site/en/bitcoin-core/features/validation.html: No such file or directory
    grep: _site/en/bitcoin-core/help.html: No such file or directory
    grep: _site/en/bitcoin-core/help.html: No such file or directory
    grep: _site/en/bitcoin-core/help.html: No such file or directory
    grep: _site/en/development.html: No such file or directory
    ERROR:
    _site/en/bitcoin-core/features/privacy.html#decentralized-peer-discovery not found
[...]

PR #4936, opened 2026-08-11
Add pre-build YAML syntax check

devdavidejesus:
## Problem

An invalid YAML escape in `_translations/en.yml` aborted the Jekyll build, and no check named the file. The existing `check-validate-yaml` only schema-validates `_wallets/`; the plain YAML files the build loads are never syntax-checked.

## Fix

New `check-yaml-syntax` target in `pre-build-tests-fast` (runs before Jekyll in `test`, `valid`, `travis` and `deployment`): parses every tracked `*.yml`/`*.yaml` with Psych  and fails naming each invalid file. Stdlib only, no new dependencies. Implements #4911.

## Validation

Passes on master (34 files). Re-introducing an invalid escape into `en.yml` fails with the file name and position:

    check-yaml: _translations/en.yml: found unknown escape character while parsing a quoted scalar at line 8 column 13
    
    
cc @Cobra-Bitcoin

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,6 +59,7 @@ pre-build-tests-fast: check-for-non-ascii-urls check-for-wrong-filename-assignme
     check-for-consistent-bitcoin-core-titles \
     check-for-too-many-wallets-on-one-platform \
     check-validate-yaml \
+    check-yaml-syntax \
     check-wallet-description-length \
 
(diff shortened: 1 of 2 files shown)


Cobra-Bitcoin:
LGTM! Very much needed.

PR #4938, opened 2026-08-11
Fail loudly when the site build dies (pipefail + built-site guard)

devdavidejesus:
## Problem

In the recent outage (#4911), `jekyll build` died on the server but the `| tee` pipeline masked its exit status, so `make` continued and the post-build checks failed confusingly against an empty `_site`. The second outage again froze the site with no reported cause. Nothing verifies that a site was actually produced, and the build-error grep misses environmental failures such as `Killed` or `No space left`.

## Fix

* Enable `pipefail` (with `SHELL=/bin/bash`) so Jekyll failures stop `make` immediately.
* Add `check-site-was-built` to verify `index.html`, `img/`, and at least 3000 HTML files (a healthy build currently produces ~3.8k) before post-build checks run.
* Extend build-error detection to catch `killed`, `no space left`, and `cannot allocate`.

## Validation

* Invalid `_config.yml` → `make build` fails at the Jekyll step.
* Empty `_site` → `check-site-was-built` fails with a clear error.
* `Killed` in the Jekyll log → `check-for-build-errors` fails.

Together with #4936, this ensures invalid input fails before the build and failed or incomplete builds cannot proceed silently.

cc @Cobra-Bitcoin

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@
 S=@  ## Silent: only print errors by default;
      ## run `make S='' [other args]` to print commands as they're run
 
+SHELL=/bin/bash
 SITEDIR=_site
 JEKYLL_LOG=._jekyll.log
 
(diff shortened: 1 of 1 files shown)


Cobra-Bitcoin:
Very nice, this will catch even more errors.

PR #4944, opened 2026-08-14
Serve reference client locally again

Cobra-Bitcoin:
This reverts some of https://github.com/bitcoin-dot-org/Bitcoin.org/pull/4926.

I was thinking about it and we've been serving Bitcoin reference client binaries since 2009, it's quite a legacy to give up.

It's no big deal if we're not always up to date or have the latest versions, we should let bitcoincore.org serve everything and we can pick and choose and remain a few versions behind, sometimes a new version can have issues so we can strengthen the diversity of nodes in the network by pushing users to download slightly older versions.


devdavidejesus:
I agree.

PR #4952, opened 2026-08-19
Remove Bitcoin Knots from wallet selector

arthurbouquet:
Since release v29.3.knots20260508 (May 2026), the standard Bitcoin Knots build enforces BIP-110 (RDTS), a soft fork restricting arbitrary data in transactions. Mandatory enforcement began at block 961,632 on August 8, 2026, despite miner signaling never exceeding ~2.5% of hashrate (well below the 55% threshold). Knots-enforcing nodes rejected the first non-signaling block at that height and split onto their own chain.

The two chains have not reconverged. Bitcoin's main chain, followed by the overwhelming majority of hashrate and by Bitcoin Core, has continued producing blocks at the normal rate and is now well past block 963,000. The Knots/RDTS chain, mined intermittently by a single pool (Roughnecks/ Ocean), has produced only 4 blocks in the same period and stalled at block 961,635.

Luke Dashjr, Knots' maintainer, has since announced a second, more fundamental break: a planned proof-of-work change from SHA-256d to BLAKE2b (algorithm confirmed August 11, targeted for September 1, 2026), intended to let the RDTS chain mine independently of Bitcoin's SHA-256 hashrate and difficulty. Unlike the RDTS soft fork, this is a hard fork by design, producing an entirely separate network and asset. Dashjr was subsequently removed as a Bitcoin BIP editor over his role in BIP-110.

A "pre-RDTS" Knots build without any of this enforcement does exist and stays on Bitcoin's main chain, but it is not the default download, and nothing on the current wallet selector page distinguishes between the two builds.

As it stands, a user who downloads the standard Bitcoin Knots build today is not running the same network as Bitcoin Core users and the rest of the Bitcoin ecosystem, and is on a path toward an even more divergent chain in September. This removes the Bitcoin Knots entries (Windows/Mac/Linux) from the wallet selector.

References:
- https://www.kucoin.com/news/flash/bip-110-fork-chain-mined-only-4-blocks-in-7-days-plans-to-switch-to-blake2b-in-september
- https://cryptorank.io/news/feed/844e6-luke-dashjr-blake2b-bip110-pow-algorithm
- https://cryptorank.io/news/feed/1e756-luke-dashjr-removed-as-bitcoin-bip-editor
- https://start9.com/bip110/


Cobra-Bitcoin:
This will be the second wallet we've removed in a month due to the potential for loss of funds. It's very sneaky to put people on a different chain and not make that clear. There's no replay protection. The users we're pointing to these wallets have no idea about chains and forks and many of them don't even speak English and aren't keeping up closely with events. They just see a wallet recommended by what they believe is Bitcoin's "official" site and trust us that everything will be fine.

**ACK** on removing it.

FWIW I still have a lot of respect for Luke. I ran Knots for years before it was cool, he always had his quirks but the community worked around them and his contributions made it worth it to deal with him. He's obviously some kind of autistic genius who follows ideas _too logically_ to their extreme conclusion. I think a certain crowd took advantage of his nature and made him into their champion. He is the public face but there are nyms behind 110 who are running that operation for some ulterior motive.

crwatkins:
ACK.  Bitcoin Knots currently provides a wallet which by default makes it difficult for users to keep funds safe.

devdavidejesus:
ACK on removing Knots from bitcoin.org

PR #4958, opened 2026-08-20
Add a price page 

Cobra-Bitcoin:
This adds a `/price` page with a simple chart of the Bitcoin price in different currencies.

This will capture more traffic from Google and improve SEO, driving traffic to our other content.


devdavidejesus:
I’ll take a look and test it.

devdavidejesus:
Built and tested locally in a pt-BR browser: currency auto-detection, chart ranges, converter and refresh all work, and the production CSP is compatible. LGTM.

PR #4976, opened 2026-08-25
Design proposal: guided wallet journey for choose-your-wallet

devdavidejesus:
Design proposal, opened as a draft for feedback before any implementation.

**Live preview:**
https://raw.githack.com/devdavidejesus/Bitcoin.org/proposal-wallet-journey-2026/_prototypes/wallet-journey-2026-prototype.html

(Single self-contained file — it can also be downloaded from this PR and opened in any browser.)

## Proposal

- **Guided path**: three plain-language questions become criteria filters — never endorsements. Every vetted wallet that matches is shown, shuffled (same no-favorites philosophy as `shuffleWalletRows()`). Long-term storage results in a category (hardware), not a brand.
- **Browse path**: the comparison table remains, one glyph per cell, legend stated once, detail on demand.

Static Jekyll + vanilla JS; browse doubles as the noscript fallback; the guided path filters the per-platform scores already in `_wallets/*.md`. Prototype data is illustrative.

If the direction holds, implementation comes as small reviewable PRs.

### DIFF
--- a/_prototypes/wallet-journey-2026-prototype.html
+++ b/_prototypes/wallet-journey-2026-prototype.html
@@ -1,0 +1,492 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<title>bitcoin.org — Find your wallet (2026 design prototype)</title>
+<style>
+  :root{
+    --orange:#F7931A; --orange-dk:#D97B06; --navy:#14213D; --navy2:#22304F;
+    --ink:#24292F; --gray:#69707A; --line:#E4E7EB;
+    --paper:#F7F8FA; --card:#FFFFFF; --soft:#FBF7F0;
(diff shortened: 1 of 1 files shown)


devdavidejesus:
--> DO NOT MERGE <--

Cobra-Bitcoin:
I think the guided flow we have on the site now is fine as it is. We also shouldn't introduce new text that needs to be translated. The problem on the site right now is when the user is finished and is presented with the results, the results page is too busy.

I like how the results are presented here. This is much better in that respect. It's an improvement over what we have now.

Also right now if you go to an individual wallet (https://bitcoin.org/en/wallets/mobile/ios/unstoppable/?step=5&platform=ios) you will notice it has too much going on too, maybe we can simplify that and present that same information in a better way.

devdavidejesus:
Agreed on all three points. Closing this draft and implementing as two small PRs:

1. **Results table**: one glyph per cell, single legend, detail on demand.
2. **Individual wallet pages**: same information, reorganized — scores in one compact strip, per-criterion detail collapsed.

No new translatable strings — both changes reuse existing translated text and reduce it.

PR #4978, opened 2026-08-25
Declutter the wallet results table

devdavidejesus:
Follow-up to the feedback on #4976: the results page declutter, scoped to reuse-only.

- One glyph per cell: the per-cell score words become visually hidden (kept in the DOM for screen readers).
- The score legend moves from below the table to above it, stated once.
- Everything reuses the existing `wallet-scoring-*` translated strings — no new translatable text.

### DIFF
--- a/_includes/layout/base/wallets-selector.html
+++ b/_includes/layout/base/wallets-selector.html
@@ -475,6 +475,13 @@ https://opensource.org/licenses/MIT.
             </div>
           </div>
 
+          <div class="wallet-score-legend">
+            <span class="wallet-legend-item"><span class="wallet-score wallet-good"></span>{% translate wallet-scoring-good choose-your-wallet %}</span>
+            <span class="wallet-legend-item"><span class="wallet-score wallet-pass"></span>{% translate wallet-scoring-pass choose-your-wallet %}</span>
+            <span class="wallet-legend-item"><span class="wallet-score wallet-caution"></span>{% translate wallet-scoring-fail choose-your-wallet %}</span>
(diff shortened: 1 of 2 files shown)


crwatkins:
Do you have a preview or at least a screenshot?

The last time we made changes in wallet selection flow we had a professional design team with weekly meetings and months long discussions, refinement, and reviews.   I'm definitely not proposing that we do that again, but I do need to contrast that with what is going on now which seems a bit on the reckless side.  I just wasted quite a bit of my time reviewing #4976 trying to figure out how to fix the numerous little problems with it and in general not being too happy with the overall tone of it.

devdavidejesus:
Live preview of this change (built page): https://raw.githack.com/devdavidejesus/Bitcoin.org/preview-declutter/results.html?step=5&platform=ios

Cobra-Bitcoin:
Ideally what we are trying to do is just take the old version of the wallet page and apply a new skin on top.

crwatkins:
The "for your operating system title" doesn't make any sense when you selelct a hardware wallet.   There are other little nits (such as those orange arrows and the count of the number of wallets in such a prominent position) and slight layout problems on different platforms that may not really matter, but are not an improvement.

I get we want this page to match the new look, but it should be at least as functional and well tested as the previous design.

devdavidejesus:
I'm going to step back from this one. I tried to contribute on the design side, but I realized it's not my strength. The feedback made it clear that this area deserves more attention than I can give it. I'll close the PRs. Thanks, guys.

PR #4987, opened 2026-08-30
Add a transaction fees page

devdavidejesus:
Adds a transaction fees page at /fees/ for newcomers wondering what it costs to send Bitcoin:

- Live fee estimates by confirmation target, in sat/vB plus the approximate cost of a simple (~140 vB) transaction in sats and USD
- Plain-language sections: what a fee is, why fees change, how sat/vB works, choosing a fee, stuck transactions (mempool/RBF), and Lightning for small payments — with links to /en/resources#lightning and /en/spend-bitcoin
- Live data fetched client-side from mempool.space with blockstream.info as a fallback, briefly cached in localStorage; the page degrades gracefully if the APIs are unreachable and never handles funds
- English-only and gated in the menus, following the /price/ pattern and visual identity; the FAQ menu entry moves to the end so Price and Fees sit together; one menu translation key added

Static copy contains no perishable numbers, so it stays accurate as fee conditions change.

### DIFF
--- a/_includes/layout/base/footer-menu.html
+++ b/_includes/layout/base/footer-menu.html
@@ -134,6 +134,9 @@ https://opensource.org/licenses/MIT.
           <a href="/price/">{% translate menu-price layout %}</a>
       {% endif %}
       {% if page.lang == 'en' %}
+          <a href="/fees/">{% translate menu-fees layout %}</a>
+      {% endif %}
+      {% if page.lang == 'en' %}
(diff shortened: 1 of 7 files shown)


devdavidejesus:
@Cobra-Bitcoin live preview of the page, built from this branch: https://raw.githack.com/devdavidejesus/Bitcoin.org/preview-fees/fees.html — the fee estimates load live in the browser (the APIs allow cross-origin requests), so this is exactly how the page behaves.

